PR #30, opened 2019-09-06
Add support for multiple wallets, and metamask as default

### LINKED ISSUE #3: Connect wallet

mariano-aguero:
Closes #3 

### Includes
- Add env variable for wallet selection
- Add support for metamask as a connector, and as default
- Add connectors type

### DIFF
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -1,48 +1,77 @@
 import React from 'react'
-import Web3Provider from 'web3-react'
+import Web3Provider, { useWeb3Context, Web3Consumer } from 'web3-react'
 import WalletConnectQRCodeModal from '@walletconnect/qrcode-modal'
-import { WalletConnect } from './connectors'
-import { useWeb3Context } from 'web3-react'
+import connectors from './connectors'
 
 const ConnectWallet: React.FC = () => {
   const context = useWeb3Context()
 
-  const connect = () => {
-    context.setConnector('WalletConnect', { networkId: 4 })
+  if (context.error) {
+    console.error('Error in web3 context', context.error)
   }
 
   React.useEffect(() => {
-    if (context.active && !context.account) {
+    if (context.active && !context.account && context.connectorName === 'WalletConnect') {
       const uri = context.connector.walletConnector.uri
       WalletConnectQRCodeModal.open(uri, () => {})
 
       context.connector.walletConnector.on('connect', () => {
         WalletConnectQRCodeModal.close()
       })
+    } else {
+      try {
+        WalletConnectQRCodeModal.close()
+      } catch (err) {
+        console.error(err)
+      }
     }
   }, [context, context.active])
 
-  return <button onClick={connect}>Connect</button>
+  return (
+    <>
+      {Object.keys(connectors).map(connectorName => (
+        <button
+          key={connectorName}
+          disabled={context.connectorName === connectorName}
+          onClick={() => context.setConnector(connectorName)}
+        >
+          Connect with {connectorName}
+        </button>
+      ))}
+
+      {(context.active || (context.error && context.connectorName)) && (
+        <button onClick={() => context.unsetConnector()}>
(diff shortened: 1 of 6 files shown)


app/src/App.tsx
```diff
@@ -1,48 +1,77 @@
...
   React.useEffect(() => {
-    if (context.active && !context.account) {
+    if (context.active && !context.account && context.connectorName === 'WalletConnect') {
       const uri = context.connector.walletConnector.uri
       WalletConnectQRCodeModal.open(uri, () => {})
 
       context.connector.walletConnector.on('connect', () => {
         WalletConnectQRCodeModal.close()
       })
+    } else {
```
fvictorio: I don't think this goes in the `else` branch.

app/src/connectors.ts
```diff
@@ -10,3 +17,9 @@ export const WalletConnect = new Connectors.WalletConnectConnector({
   },
   defaultNetwork: 4,
 })
+
+const connector = (process.env.REACT_APP_CONNECTOR || 'MetaMask') as ConnectorsEnum
+
+const connectorToExport = connector === ConnectorsEnum.MetaMask ? { MetaMask } : { WalletConnect }
```
fvictorio: PP: I don't love the idea of exporting something based on an environment variable. I think it would be better to export all connectors (as before) and to choose which one to use based on the env var.

app/src/types.ts
```diff
@@ -0,0 +1,4 @@
+export enum ConnectorsEnum {
```
fvictorio: Is this useful? It's not type-safer (the env var is going to be stringly typed anyway) and it's not clearer. I think a plain string comparison should be enough.

app/src/App.tsx
```diff
@@ -1,48 +1,77 @@
...
   }, [context, context.active])
 
-  return <button onClick={connect}>Connect</button>
-}
-
-const Account: React.FC = () => {
-  const context = useWeb3Context()
+  return (
+    <>
+      {Object.keys(connectors).map(connectorName => (
```
fvictorio: PP: Unnecessarily generic at this point. Plus, if we are going to support only one wallet for now, then just a `Connect Wallet` button should be enough.

app/src/App.tsx
```diff
@@ -1,48 +1,77 @@
...
+          key={connectorName}
+          disabled={context.connectorName === connectorName}
+          onClick={() => context.setConnector(connectorName)}
+        >
+          Connect with {connectorName}
+        </button>
+      ))}
 
-  return <div>{context.account || 'not connected'}</div>
+      {(context.active || (context.error && context.connectorName)) && (
```
fvictorio: I think we can do without a logout for now.

app/src/App.tsx
```diff
@@ -1,48 +1,77 @@
...
 }
 
-const CurrentNetwork: React.FC = () => {
-  const context = useWeb3Context()
-
-  return <div>{context.networkId}</div>
+const ConnectionStatus: React.FC = () => {
+  return (
+    <Web3Consumer>
+      {// eslint-disable-next-line
```
fvictorio: Is this because of the `no-explicit-any` rule? I think we should just disable it, otherwise we are going to be using these comments a lot.

app/src/App.tsx
mariano-aguero: Done!

app/src/connectors.ts
mariano-aguero: Done!

app/src/types.ts
mariano-aguero: Done!

app/src/App.tsx
mariano-aguero: Done!

app/src/App.tsx
mariano-aguero: Done!

app/src/App.tsx
mariano-aguero: Done!

PR #32, opened 2019-09-09
Add Dockerfile for ganache with realitio contracts

fvictorio:
This PR adds a Dockerfile for a a ganache with the realitio contracts deployed. The README has instructions on how to use it.

### How it works

The Dockerfile installs some stuff, runs a bash script and then starts ganache with a directory as its db. The bash script builds the initial state of that db by starting ganache in the background and deploying the contracts.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,20 @@
+# Docker
+
+This Dockerfile builds an image for a container that runs `ganache-cli` with the necessary contracts pre-deployed. To
+build it, do:
+
+```
+docker build -t gnosis-cond-ganache .
+```
+
+And then run it:
+
+```
+docker run -it --rm gnosis-cond-ganache
+```
+
+The addresses of the contracts are saved in a file inside the container. You can see them with:
+
+```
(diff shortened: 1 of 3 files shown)


docker/README.md
```diff
@@ -0,0 +1,20 @@
...
+And then run it:
+
+```
+docker run -it --rm gnosis-cond-ganache
+```
+
+The addresses of the contracts are saved in a file inside the container. You can see them with:
+
+```
+docker run -it --rm cat contracts_addresses.txt
```
mariano-aguero: Q: I think it is necessary to add the name of the container? please correct me if I am wrong, like this `docker run -it --rm gnosis-cond-ganache cat contracts_addresses.txt`

fvictorio: Oops, that's right. Fixed.

leolower:
Looks good.

In theory you could use a [multi-stage build](https://docs.docker.com/develop/develop-images/multistage-build/) to generate the `/db` stuff and only keep that on the final image but I don't think it makes sense here since the final image also needs to include the tool used to generate `/db`

PR #33, opened 2019-09-09
Create market form v1

### LINKED ISSUE #31: Create market form v1

mariano-aguero:
Closes #31 

### Includes
- This is a simple form only to start work in the market creation
- Added some libraries, styled-components, react-router-dom, and some types
- Move `ConnectWallet` and `ConnectionStatus` components to a folder
- Create some common components in the `components/common` folder like a Button, Textfield, Categories.
- Create a theme folder for styled components
- Create a simple route  to the `create market` section
- Refactor the app file so it is smaller
- Move the connectors file to the util folder
- Create a Market Create component with a hooks file


fvictorio:
Do we need support for theming? (cc @leolower)

app/src/components/common/button/index.tsx
```diff
@@ -0,0 +1,18 @@
+import React, { ReactNode, HTMLAttributes } from 'react'
+
+interface Props extends HTMLAttributes<HTMLButtonElement> {
```
fvictorio: What's the difference between this component and a plain `button`?

app/src/components/connection_status/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { Web3Consumer } from 'web3-react'
+
+const ConnectionStatus: React.FC = () => {
+  return (
+    <Web3Consumer>
+      {(context: any) => {
+        const { active, account, networkId } = context
+        return (
+          active && (
+            <React.Fragment>
```
fvictorio: `<>`

app/src/components/market/create.tsx
```diff
@@ -0,0 +1,103 @@
...
+import { useCreateMarket } from './hooks/use_create'
+
+const Div = styled.div`
+  height: 50px;
+  display: flex;
+  align-items: center;
+`
+
+export const CreateMarket: React.FC = () => {
+  const { values, handleChange, handleSubmit } = useCreateMarket({
```
fvictorio: Why is this behind a custom hook? What are the odds we'll need to reuse this logic?

mariano-aguero:
> Do we need support for theming? (cc @leolower)

To consider, I write some points that can be useful:
- It's not just support for themes, we can put all the styles in the theme object and all components can use those styles. If we need to change a property like a primary color, fonts, icon sizes, etc, we can change it only in that object and the changes will be reflected in the application.
- We can have the main properties in a single file, being able to make modifications more easily
- Sometimes it's often useful to create variants of the same component that are selected via an additional prop, example a button. See more [here](https://github.com/styled-components/styled-theming#themevariantsname-prop-themes)

app/src/components/market/create.tsx
mariano-aguero: The hook is in another file to keep the main file with fewer lines, and is generic because the form still does not do something specific. I think this will probably change in the next weeks because we will add more functionality...

app/src/components/common/button/index.tsx
mariano-aguero: Is just a component reuse strategy. When we have the design we will apply styles to this component and reuse in different places of the application.
I think we will apply something like this:
```
const ButtonStyled = styled.button`
  align-items: center;
  background-color: ${props => props.theme.colors[props.type]};
  border-radius: ${props => props.theme.buttons.borderRadius};
  border: none;
  color: ${props => props.theme.buttons.color};
  cursor: pointer;
  display: flex;
  font-size: ${props => props.theme.buttons.fontSize};
  font-weight: ${props => props.theme.buttons.fontWeight};
  height: ${props => props.theme.buttons.height};
  justify-content: center;
`
```
Is just an example, I'm thinking aloud.

app/src/components/connection_status/index.tsx
mariano-aguero: Updated to use the short syntax. Thanks for all the feedback

fvictorio:
> * It's not just support for themes, we can put all the styles in the theme object and all components can use those styles. If we need to change a property like a primary color, fonts, icon sizes, etc, we can change it only in that object and the changes will be reflected in the application.

Oh, that's right, I take it back.

app/src/components/market/create.tsx
fvictorio: I think adding indirection just to reduce the average number of lines per file is not a good practice. The custom hook is not reusable nor testable, and it doesn't separate concerns. The only advantage of this over inlining everything is that a file is split in two (which is different from separating logic in two modules).

That being said, I don't think this is enough to block this PR.

PR #43, opened 2019-09-16
Add integration of ConditionalTokenService

mariano-aguero:
Connects with #9 

### Includes
- Integration of the condition token service


fvictorio:
Tested it locally and in Rinkeby. Looks good.

PR #45, opened 2019-09-17
Add Market Maker service

mariano-aguero:
Connects with #10

### Includes
- Add FEE constant with 0.1 ether
- Add market maker service, create a contract
- Integration with the `market_wizard_creator_container`


fvictorio:
@mariano-aguero The service should return the address of the deployed contract. I think this can be done with an approach similar to the one we use for getting the `questionId` in the Realitio service.

fvictorio:
@mariano-aguero My changes are non-trivial, so please review them and then merge this if you are OK with them.

PR #60, opened 2019-09-23
Add outcome component

### LINKED ISSUE #59: Validate initial odds add up to 100%

mariano-aguero:
Closes #59 

### Includes
- Outcome component
- Validation up to 100
- Validation for numbers in step 2 and step 3


fvictorio:
I think this is going to be painful in the long run. I have several concerns (the `event` object being mutated inside a map, the way the debounce is declared, the fact itself that the debounce is needed). I'm pretty sure making `Outcomes` stateless would solve all of this. A draft would be:

```jsx
const [probabilities, setProbabilities] = useState([50, 50])
const handleChange = (index, value) {
  const newProbabilities = [...probabilities.slice(0, index), value, ...probabilities.slice(index+1)]
  // change values after index if you want, but I think we can do without that for now
  setProbabilities(newProbabilities)
}

return <Outcomes
  probabilities={probabilities}
  onChange={handleChange} />
```

where `Outcomes` doesn't have state:

```jsx
return props.probabilities.map((probability, index) => 
  <input onChange={e => props.onChange(index, e.currentTarget.value)} value={probability} />
)
```

(I didn't test anything of this, but I think the idea should work)

mariano-aguero:
> I think this is going to be painful in the long run. I have several concerns (the `event` object being mutated inside a map, the way the debounce is declared, the fact itself that the debounce is needed). I'm pretty sure making `Outcomes` stateless would solve all of this. A draft would be:
> 
> ```js-jsx
> const [probabilities, setProbabilities] = useState([50, 50])
> const handleChange = (index, value) {
>   const newProbabilities = [...probabilities.slice(0, index), value, ...probabilities.slice(index+1)]
>   // change values after index if you want, but I think we can do without that for now
>   setProbabilities(newProbabilities)
> }
> 
> return <Outcomes
>   probabilities={probabilities}
>   onChange={handleChange} />
> ```
> 
> where `Outcomes` doesn't have state:
> 
> ```js-jsx
> return props.probabilities.map((probability, index) => 
>   <input onChange={e => props.onChange(index, e.currentTarget.value)} value={probability} />
> )
> ```
> 
> (I didn't test anything of this, but I think the idea should work)

I update the `Outcome` component to an stateless version, remove the debounce with lodash, and some things that were too complex, I think it's more simple right now.

PR #62, opened 2019-09-24
Add fetch market service

### LINKED ISSUE #13: Fetch market information

mariano-aguero:
Closes #13 

### Includes
- Fetch market service


app/src/services/fetch_market.ts
```diff
@@ -0,0 +1,64 @@
...
+  Paused = 1,
+  Closed = 2,
+}
+
+class FetchMarketService {
+  address: string
+  contract: any
+
+  constructor(address: string, provider: any) {
+    this.address = address
```
fvictorio: We won'w need this if we have the contract.

app/src/services/fetch_market.ts
```diff
@@ -0,0 +1,64 @@
...
+    return await this.contract.functions.pmSystem()
+  }
+
+  async getMarket(): Promise<any> {
+    const [funding, fee, conditionId, outcomeSlots, stage, collateralToken] = await Promise.all([
+      this.getFunding(),
+      this.getFee(),
+      this.getConditionIds(),
+      this.getOutcomeSlots(),
+      this.getOutcomeSlots(),
```
fvictorio: This is duplicated.

app/src/services/fetch_market.ts
```diff
@@ -0,0 +1,151 @@
...
+    const conditionalTokensAddress = getContractAddress(networkId, 'conditionalTokens')
+
+    this.marketMakerContract = new ethers.Contract(marketMakerAddress, marketMakerAbi, provider)
+    this.conditionalTokensContract = new ethers.Contract(
+      conditionalTokensAddress,
+      conditionTokenAbi,
+      provider,
+    )
+    this.daiTokenAddress = getContractAddress(networkId, 'dai')
+    this.ownerAddress = ownerAddress
```
fvictorio: I would remove this and add it as a paremeter of `getBalanceOf`.

app/src/components/market/steps/items/resume_market_step.tsx
```diff
@@ -0,0 +1,88 @@
...
+      funding,
+      outcomeValueOne,
+      outcomeValueTwo,
+      outcomeProbabilityOne,
+      outcomeProbabilityTwo,
+    } = values
+
+    const resolutionDate = resolution && moment(resolution).format('MM-DD-YYYY')
+
+    const marketMakerURL = `${APP_URL}view/${marketMakerAddress}`
```
fvictorio: I don't think we should get this from an environment variable. There must be a way to do it either with `window.location` or with `react-router` (`window.location` might not work if the app is deployed under a sub-path).

app/src/components/market/market_view_container.tsx
```diff
@@ -0,0 +1,73 @@
...
+    actualPrice: [],
+  })
+  const [address, setAddress] = useState(props.match.params.address)
+  const [status, setStatus] = useState(Status.Ready)
+
+  useEffect(() => {
+    const fetchData = async () => {
+      setStatus(Status.Loading)
+      try {
+        if (!context.networkId || !context.library) {
```
fvictorio: The `useConnectedWeb3Context` hook is meant to avoid these checks. You need to put the component under a `<ConnectedWeb3>` component, and it won't render until web3 is available, but I think that's fine and it will simplify the code a lot.

mariano-aguero: Add ConnectedWeb3 in the Main component

app/src/components/market/steps/items/resume_market_step.tsx
mariano-aguero: React router doesn't have a solution for this, I use window location

fvictorio:
I don't know if you are going to continue committing here but, if not, I think this can be merged.

PR #67, opened 2019-09-27
Add BigNumberInput component to the application

mariano-aguero:


app/src/util/token_amount.ts
```diff
@@ -0,0 +1,42 @@
+import { BigNumber } from 'ethers/utils'
+import { TokenAmountInterface } from './types'
+
+class TokenAmount implements TokenAmountInterface {
+  public amount: BigNumber
+  public decimals: number
+
+  public static fromString(value: string, decimals: number) {
```
fvictorio: Can we replace this method and the next one for `ethers.utils.parseUnits` and `formatUnits`? (And therefore drop this class entirely)

mariano-aguero: Done

PR #76, opened 2019-10-02
Fetch Realitio question from events

fvictorio:


app/src/services/conditional_token.ts
```diff
@@ -41,6 +42,40 @@ class ConditionalTokenService {
...
+
+    const conditionalTokenContract = new ethers.Contract(
+      conditionalTokensAddress,
+      conditionTokenAbi,
+      provider,
+    )
+
+    const filter: any = conditionalTokenContract.filters.ConditionPreparation(conditionId)
+
+    filter.fromBlock = '0x1'
```
fvictorio: I thought about making this configurable because I thought it could be slow, but since it's filtering the events by a indexed topic, it's super fast, even in mainnet.

PR #90, opened 2019-10-04
Disable Redeem and Withdraw buttons when they won't do anything

### LINKED ISSUE #86: Disable Redeem and Withdraw buttons when they won't do anything
Disabler Redeem button when the user doesn't have winning outcome tokens and Withdraw button when the market maker has no collateral tokens.

mariano-aguero:
Closes #86

### DIFF
--- a/app/src/components/market/profile/redeem.tsx
+++ b/app/src/components/market/profile/redeem.tsx
@@ -116,6 +116,8 @@ export const Redeem = (props: Props) => {
     }
   }
 
+  const haveShares = balance.every((balanceItem: BalanceItems) => balanceItem.shares.isZero())
+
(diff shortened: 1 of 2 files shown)


app/src/components/market/profile/redeem.tsx
```diff
@@ -116,14 +116,18 @@ export const Redeem = (props: Props) => {
     }
   }
 
+  const haveShares = balance.every((balanceItem: BalanceItems) => balanceItem.shares.isZero())
```
fvictorio: This should only check if the user has shares for the winning token. If they have shares but only for the losing outcome, the redeem button should be disabled anyways.

PR #93, opened 2019-10-07
Add improvements

mariano-aguero:
No issue related

### Includes
- Refactoring of services
- Don't approve DAI for initial trade if the odds are 50/50
- Refactor useContracts to useMemo and to receive the context
- Add ClosedMarketDetail component and have an unique behaviour


app/src/components/market/market_view_container.tsx
```diff
@@ -111,7 +111,8 @@ const MarketViewContainer: FC<Props> = props => {
     }
 
     fetchQuestion()
-  }, [address, context, conditionalTokens, realitio])
+    // eslint-disable-next-line
+  }, [address, context])
```
fvictorio: This rule is very important, why is being disabled?

PR #138, opened 2019-10-29
Add price for market

### LINKED ISSUE #98: Update price computation/fetching

mariano-aguero:
Closes #98 

#### Includes
- Calculate price for shares 'Yes' with this formula:
`priceYes = holdings(no) / (holdings(yes) + holdings(no))`
- Calculate price for shares 'No' with this formula:
`priceNo = holdings(yes) / (holdings(yes) + holdings(no))`
- The total of priceYes and priceNo must be 1

### DIFF
--- a/app/src/services/market_maker.ts
+++ b/app/src/services/market_maker.ts
@@ -4,6 +4,7 @@ import { BigNumber } from 'ethers/utils'
 import { ConditionalTokenService } from './conditional_token'
 import { getLogger } from '../util/logger'
 import { OutcomeSlot } from '../util/types'
+import { divBN } from '../util/tools'
 
 const logger = getLogger('Services::MarketMaker')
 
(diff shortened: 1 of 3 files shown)


app/src/services/market_maker.ts
```diff
@@ -61,30 +62,17 @@ class MarketMakerService {
...
-    //     .div(twoPower64)
-    //     .toNumber() / 10000
-    // actualPriceForNo =
-    //   actualPriceForNo
-    //     .mul(new BigNumber(10000))
-    //     .div(twoPower64)
-    //     .toNumber() / 10000
-    //
+    const { address } = this.contract
+    const { balanceOfForYes, balanceOfForNo } = await this.getBalanceInformation(address)
```
fvictorio: Not sure about this. We are fetching the balance information twice for no reason, and it worries me a little that we are already doing a lot of concurrent requests to the contract here. I think we should use the result of the balances in the `useMarketMakerData` hook to compute the prices.

mariano-aguero: Ok, I use the result of the balances in the useMarketMakerData, and pass the value as a parameter to compute the prices.

app/src/services/market_maker.ts
```diff
@@ -62,9 +62,11 @@ class MarketMakerService {
     return this.contract.removeFunding(amount)
   }
 
-  getActualPrice = async (): Promise<{ actualPriceForYes: number; actualPriceForNo: number }> => {
-    const { address } = this.contract
-    const { balanceOfForYes, balanceOfForNo } = await this.getBalanceInformation(address)
+  getActualPrice = async (balanceInformation: {
+    balanceOfForYes: BigNumber
+    balanceOfForNo: BigNumber
+  }): Promise<{ actualPriceForYes: number; actualPriceForNo: number }> => {
```
fvictorio: I think this doesn't need to be async anymore? Also, it doesn't use the underlying contract, so this could either be a static method or a separate util (I think a static method is fine).

mariano-aguero: > I think this doesn't need to be async anymore? Also, it doesn't use the underlying contract, so this could either be a static method or a separate util (I think a static method is fine).

Good catch, Done!

PR #143, opened 2019-10-30
Add page not found

mariano-aguero:
No issue related, added page not found section when a section doesn't exist.
The problem occurs when  any type of string is placed instead of a value corresponding to the address  of the market maker, example `localhost/#/home`.

I have some doubts regarding the text,  and we should improve it in some way, maybe we can suggest to the user that should actually put a valid address?


app/src/util/tools.ts
```diff
@@ -121,6 +124,7 @@ export const computePriceAfterTrade = (
 
 export const isAddress = (address: string): boolean => {
   try {
+    logger.log(`Address '${address}' doesn't exist`)
```
fvictorio: Shouldn't this be in the catch block?

mariano-aguero: Ohh yes, sorry about that, I will move it, thanks

PR #150, opened 2019-11-01
Fix values in funding page

fvictorio:
This PR also changes the way formatted big numbers are shown, so that less decimal values are displayed.


app/src/services/market_maker.ts
```diff
@@ -21,6 +22,7 @@ const marketMakerAbi = [
   'function calcBuyAmount(uint investmentAmount, uint outcomeIndex) public view returns (uint)',
   'function sell(uint returnAmount, uint outcomeIndex, uint maxOutcomeTokensToSell) external',
   'function calcSellAmount(uint returnAmount, uint outcomeIndex) public view returns (uint)',
+  ...erc20Abi,
```
mariano-aguero: Looks Good! I think this spreed variable will create some problems, because already exist some methods definitions like `balanceOf` and this will generate an error with multiple definitions of the same method, maybe we can remove

fvictorio: You're right, fixed.

PR #166, opened 2019-11-04
Add list market logic and filter

### LINKED ISSUE #153: List markets Logic
Only question and resolution date

### LINKED ISSUE #154: Filter: only markets I created

mariano-aguero:
Closes #153 
Closes #154

#### Includes
- A function in the marketMakerFactory service, this function obtain the markets 
- A useMarket hook, right here we call the `getMarkets` function, and for every market we obtain the realitio information for that market, finally returns all the market information needed to be listed
- Support connected and disconnected context
- Added a filter,we can filter markets that we own

Screenshot_20191108_175513

#### How to test it
- If we enter to the home section, we can see a list with the markets


app/src/pages/market_home_page.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+
+import { MarketHomeContainer } from '../components'
+import { ConnectedWeb3, useConnectWeb3 } from '../hooks/connectedWeb3'
+
+const MarketHomePage = () => {
+  useConnectWeb3()
```
fvictorio: The home page is the one that doesn't have to have this. Otherwise it doesn't make sense to have the "Connect" buttons.

app/src/components/market/market_home_container.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { useMarkets } from '../../hooks/useMarkets'
+
+interface Props {
+  marketMakerAddress: string
+}
+
+const MarketHomeContainer = () => {
+  const context = useConnectedWeb3Context()
+
+  const { markets, status } = useMarkets(context)
```
fvictorio: Is this hook going to be used elsewhere? If not, what's the point of having a hook for this?

app/src/services/market_maker_factory.ts
```diff
@@ -40,6 +43,38 @@ class MarketMakerFactoryService {
 
     return marketMakerAddress
   }
+
+  getMarkets = async (provider: any): Promise<{ markets: Market[] }> => {
```
fvictorio: I think returning `Promise<Market[]>` should be enough.

app/src/services/market_maker_factory.ts
```diff
@@ -40,6 +43,38 @@ class MarketMakerFactoryService {
...
+    const filter: any = this.contract.filters.FixedProductMarketMakerCreation()
+
+    const logs = await provider.getLogs({
+      fromBlock: 1,
+      toBlock: 'latest',
+      ...filter,
+    })
+
+    if (logs.length === 0) {
+      throw new Error(`No FixedProductMarketMakerCreation events found`)
```
fvictorio: Why throw an error? It's perfectly fine for the array to be empty, we just have to handle this in the UI ("There are no markets...").

app/src/hooks/useMarkets.tsx
```diff
@@ -0,0 +1,63 @@
...
+  useEffect(() => {
+    const fetchMarkets = async () => {
+      try {
+        setStatus(Status.Loading)
+        const provider = context.library
+
+        const getQuestionData = async (market: Market) => {
+          const { conditionId } = market
+          // Get question data
+          const questionId = await conditionalTokens.getQuestionId(conditionId, provider)
```
fvictorio: Let's refactor `getQuestionId` so that it doesn't receive the provider as an argument (the instance already has access to the provider).

app/src/hooks/useMarkets.tsx
```diff
@@ -0,0 +1,63 @@
...
+          const { question, resolution } = await realitio.getQuestion(questionId, provider)
```
fvictorio: Same comment as previous line.

app/src/services/market_maker_factory.ts
mariano-aguero: I can change it, I will return an empty array

app/src/services/market_maker_factory.ts
mariano-aguero: Done

app/src/hooks/useMarkets.tsx
mariano-aguero: Good catch, done!

app/src/hooks/useMarkets.tsx
mariano-aguero: Done!

app/src/pages/market_home_page.tsx
mariano-aguero: Mmmmm right here we will have a problem, how do we solve the filtering by owneship? Maybe we can connect the user when clicks the filter, what do you think ?

fvictorio: We can render different components when the user is connected and when she's not.

mariano-aguero: I create  two components, one connected, another disconnected, also I create another context for the case in which the user is disconnected

app/src/components/market/market_home_container.tsx
mariano-aguero: Now it is used in two components, so I think it is necessary

app/src/pages/market_home_page.tsx
```diff
@@ -0,0 +1,38 @@
...
+      <MarketHomeDisconnectedContainer {...props} />
+    </DisconnectedWeb3>
+  )
+}
+
+const MarketHomePage = (props: any) => {
+  const { active } = useWeb3Context()
+  const connector = localStorage.getItem('CONNECTOR')
+
+  return active || (connector && (connector as keyof typeof connectors)) ? (
```
fvictorio: What does `connector as keyof typeof connectors` do? That's just meaningful at compile time; at runtime this is equivalent to `connector && connector` as far as I know. Is the cast helpful in some way?

app/src/hooks/disconnectedWeb3.tsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react'
+import { ethers } from 'ethers'
+import { INFURA_PROJECT_ID } from '../common/constants'
+
+export interface DisconnectedWeb3Context {
+  account: string
```
fvictorio: If `account` is going to be the empty string, I would just remove it. It also makes more sense if you are connecting via Infura.

app/src/services/conditional_token.ts
```diff
@@ -27,10 +27,14 @@ class ConditionalTokenService {
   provider: any
 
   constructor(address: string, provider: any, signerAddress: string) {
-    const signer: Wallet = provider.getSigner()
-
-    this.contract = new ethers.Contract(address, conditionalTokensAbi, provider).connect(signer)
+    try {
```
fvictorio: I don't like this. There has to be a better way. Either checking if there's an account present in the contract, or using one of the constructor parameters to indicate if a signer is present. In fact, there's a `signerAddress` parameter, so maybe that can be used?

app/src/services/market_maker_factory.ts
```diff
@@ -1,24 +1,36 @@
...
+
 class MarketMakerFactoryService {
   contract: Contract
   constantContract: Contract
   signerAddress: string
   provider: any
 
   constructor(address: string, provider: any, signerAddress: string) {
-    const signer: Wallet = provider.getSigner()
+    try {
```
fvictorio: Same comment.

mariano-aguero: Done!

app/src/services/conditional_token.ts

app/src/hooks/disconnectedWeb3.tsx
mariano-aguero: Done, I remove it

app/src/pages/market_home_page.tsx
mariano-aguero: I added an `in` operator to validate the connector exist

gabitoesmiapodo:
Added some styling to this section.

You should get to see something like what's in the screenshots below.

Filters dropdown won't do anything, but it should be useful for https://github.com/protofire/gnosis-conditional-exchange/issues/154

Screen Shot 2019-11-08 at 16 58 02
Screen Shot 2019-11-08 at 16 58 08

app/src/services/conditional_token.ts
```diff
@@ -73,12 +75,14 @@ class ConditionalTokenService {
...
+  getQuestionId = async (conditionId: string): Promise<string> => {
     const filter: any = this.contract.filters.ConditionPreparation(conditionId)
 
-    filter.fromBlock = '0x1'
-
-    const logs = await provider.getLogs(filter)
+    const logs = await this.provider.getLogs({
+      fromBlock: 1,
+      toBlock: 'latest',
+      ...filter,
```
fvictorio: Put `fromBlock` and `toBlock` after `...filter`.

`filter` probably doesn't have those fields, but it's better to err on the safe side.

app/src/components/market/market_home.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const options = [MarketFilters.AllMarkets, MarketFilters.MyMarkets]
+  const defaultOption = options[0]
+
+  const [marketsFiltered, setMarketsFiltered] = useState<MarketAndQuestion[]>([])
+
+  useEffect(() => {
+    setMarketsFiltered(markets)
+  }, [markets])
+
+  const filterMarkets = (event: { name: string; value: string }) => {
```
fvictorio: Several comments here:

1. `name` is not used
2. Receiving an event is ugly; it's better to extract the value in a lambda function and to call the method with it.
3. I think all of this should be done in a `useEffect` whose dependencies are `markets` and something like `currentFilter`.

app/src/components/market/steps/items/ask_question_step.tsx
```diff
@@ -114,9 +113,9 @@ class AskQuestionStep extends Component<Props, State> {
           formField={
             <Arbitrators name="arbitratorId" value={arbitratorId} onChange={handleChange} />
           }
-          title={'Oracle'}
+          title={'Arbitrator'}
           tooltipText={
-            'You can choose among several available Oracles. The Oracle will resolve the market once the Resolution Date is reached.'
+            'You can choose among several available Arbitrators. The Arbitrator will resolve the market once the Resolution Date is reached.'
```
fvictorio:
```suggestion
            'You can choose among several available Arbitrators. The arbitrator will be responsible for providing a final answer to a question when there is a dispute.'
```

app/src/components/market/market_home.tsx
mariano-aguero: I update all this mechanism to use a hooks :+1:

PR #168, opened 2019-11-04
Check market address on URL and show 404

### LINKED ISSUE #162: Check market address on URL and show 404
If the address is not a contract, show a message saying what to do.

mariano-aguero:
Closes #162

### Includes
- A function that checks if an address is a contract
- A hook that uses the previous function
- A `MarketNotFound` component
- Integration in some sections

### Does not include
- Design for the `MarketNotFound` component (Create an issue for this #169 )


app/src/components/market/market_buy_container.tsx
```diff
@@ -6,23 +6,33 @@ import { FullLoading } from '../common/full_loading'
...
 interface Props {
   marketMakerAddress: string
 }
 
 const MarketBuyContainer = (props: Props) => {
   const context = useConnectedWeb3Context()
+
   const { marketMakerAddress } = props
 
+  const { contractExist } = useCheckContractExist(marketMakerAddress, context)
```
fvictorio: Rename to `contractExists` and `useCheckContractExists`.

app/src/hooks/useCheckContractExist.tsx
```diff
@@ -0,0 +1,23 @@
...
+  useEffect(() => {
+    const provider = context.library
+    const fetchIsContract = async () => {
+      setContractExist(await isContract(provider, marketMakerAddress))
+    }
+
+    fetchIsContract()
+  }, [context, marketMakerAddress])
+
+  return {
```
fvictorio: I would return the boolean instead of an object, but just PP.

fvictorio:
This has a little too much repetition, that could be avoided if we had a sub-route that checks the address. It would be something like this in the `Main` component:

```tsx
<Route path="/:address" component={MarketRoutes} />
```

(Notice that it doesn't have `exact`.) This component would check that the address exist and render the sub-routes if it does.

Not a required change, just a comment.

mariano-aguero:
> This has a little too much repetition, that could be avoided if we had a sub-route that checks the address. It would be something like this in the `Main` component:
> 
> ```tsx
> <Route path="/:address" component={MarketRoutes} />
> ```
> 
> (Notice that it doesn't have `exact`.) This component would check that the address exist and render the sub-routes if it does.
> 
> Not a required change, just a comment.

I create a market route component, this component check that the address is valid and also that it is a contract

app/src/components/market/market_fund_container.tsx
```diff
@@ -10,10 +10,10 @@ interface Props {
   marketMakerAddress: string
 }
 
-const MarketFundContainer: FC<Props> = props => {
+const MarketFundContainer = (props: Props) => {
```
fvictorio: `FC<Props>` gives better type guarantees, any reason for this change?

mariano-aguero: That change was to maintain consistency with the declaration of the other similar containers files, I will add it there and in the other files

PR #173, opened 2019-11-05
Include Category field in Realitio question and display on market details

### LINKED ISSUE #164: Include Category field in Realitio question and display on market details

mariano-aguero:
Closes #164 

#### Includes
- Add some realitio libraries , so we use the same standard for data encoding
- Integrate category in realitio service, when we create a question, and also when we get a question
- Integrate category in summary step

Note: @fvictorio mmm  should we add the category in the profile of a market (`uri:  localhost/:address`) ? 
If yes, Where should it be displayed ?


app/src/services/realitio.ts
```diff
@@ -45,11 +47,14 @@ class RealitioService {
    */
   askQuestion = async (
     question: string,
+    category: string,
     openingDateMoment: Moment,
     value = '0',
   ): Promise<string> => {
     const openingTimestamp = openingDateMoment.unix()
-    const args = [0, question, this.arbitratorAddress, '86400', openingTimestamp, 0]
+    const questionText = RealitioQuestionLib.encodeText('bool', question, [1, 0], category)
```
fvictorio: From what I see in the tests of the library (there doesn't seem to be any documentation), `[1, 0]` can (should?) be replaced with `null`.

app/src/services/realitio.ts
```diff
@@ -82,8 +87,14 @@ class RealitioService {
     const iface = new ethers.utils.Interface(realitioAbi)
     const event = iface.parseLog(logs[0])
 
+    const question: QuestionLog = RealitioQuestionLib.populatedJSONForTemplate(
```
fvictorio: Seems like you could use `defaultTemplateForType('bool')` here. Seems more readable and doesn't depend on the `'@realitio/realitio-contracts` (which, btw, it's not in the package.json dependencies, don't we have a linter rule to prevent this from happening)?

mariano-aguero: Added the suggested change, also added an eslint rule to check import of invalid libraries

app/src/services/realitio.ts
mariano-aguero: Added null for outcomes parameter

app/.eslintrc.json
```diff
@@ -17,7 +17,8 @@
     "@typescript-eslint/explicit-function-return-type": "off",
     "@typescript-eslint/no-explicit-any": "off",
     "react/prop-types": "off",
-    "no-warning-comments": "warn"
+    "no-warning-comments": "warn",
+    "import/no-extraneous-dependencies": ["error", {"devDependencies": false, "optionalDependencies": false, "peerDependencies": false}]
```
fvictorio: `devDependencies` should be true. If we are importing a dependency that is only listed in dev dependencies, we want to get a warning. Otherwise the app will work during development but not when it's built.

mariano-aguero: Done

PR #179, opened 2019-11-12
Add validation for unexisting arbitrator

mariano-aguero:
No issue related

There are some markets created with an invalid arbitrator address, 

### Includes
- Now the `getArbitratorFromAddress` function returns null
- Add Maybe in some props for Arbitrator object
- Validate display arbitrator if exist, also for category and question title ( for example, if category doesn't exist return an 'undefined' string')

### DIFF
--- a/app/src/services/realitio.ts
+++ b/app/src/services/realitio.ts
@@ -93,14 +93,18 @@ class RealitioService {
     const iface = new ethers.utils.Interface(realitioAbi)
     const event = iface.parseLog(logs[0])
 
-    const question: QuestionLog = RealitioQuestionLib.populatedJSONForTemplate(
+    const questionLog: QuestionLog = RealitioQuestionLib.populatedJSONForTemplate(
       RealitioTemplateLib.defaultTemplateForType('bool'),
       event.values.question,
     )
 
+    const category =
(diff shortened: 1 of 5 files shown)


app/src/services/realitio.ts
```diff
@@ -93,14 +93,18 @@ class RealitioService {
...
 
-    const question: QuestionLog = RealitioQuestionLib.populatedJSONForTemplate(
+    const questionLog: QuestionLog = RealitioQuestionLib.populatedJSONForTemplate(
       RealitioTemplateLib.defaultTemplateForType('bool'),
       event.values.question,
     )
 
+    const category =
+      questionLog.category && questionLog.category === 'undefined' ? '' : questionLog.category
+    const question = questionLog.title && questionLog.title === 'undefined' ? '' : questionLog.title
```
fvictorio: I think you can remove `questionLog.category &&` and `questionLog.title &&`.

If they are falsy, then the whole `field && field === 'undefined'` will be false, and the falsy value will be returned.

If the first operand of the `&&` is removed, and the value is falsy, then the `=== 'undefined'` will be false, and (again) the falsy value will be returned.

So, same result, less code and (IMO) more clear.

mariano-aguero: Done

PR #181, opened 2019-11-12
Improve 3box signing

mariano-aguero:
No issue related

Improve 3box signing, with this we only need one signing to view/update the user profile, and only another one when the user want to comment the market.

#### User signing to update/view the profile
Screenshot_20191112_161843

#### Open space to write a comment
Screenshot_20191112_161907

3box keeps the data in the localstorage, so will only be requested once
Screenshot_20191112_160417

Also I added a validation to check if the space already exist

### DIFF
--- a/app/src/components/common/three_box_comments/index.tsx
+++ b/app/src/components/common/three_box_comments/index.tsx
@@ -145,7 +145,10 @@ export const ThreeBoxComments = (props: Props) => {
       logger.log(`Open three box with account ${account}`)
 
       const newBox = await Box.openBox(account, library._web3Provider)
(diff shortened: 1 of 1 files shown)


fvictorio:
I'm curious about which part of the change avoids the second signature, since it seems like you only added a conditional (and I guess that conditional is the validation to check if the space already exists).

mariano-aguero:
> I'm curious about which part of the change avoids the second signature, since it seems like you only added a conditional (and I guess that conditional is the validation to check if the space already exists).

Yes, you're right, the openSpace method should be called only once, if  the space is not created or doesn't exist.

PR #191, opened 2019-11-15
Add value property in some transactions

### LINKED ISSUE #187: Error when a market is created
Check the image below

mariano-aguero:
Closes #187 

#### Includes
- Add value property with '0x0' in some transactions, support for some wallets mobile like trust and pillar
- Added fix for division by zero error

### DIFF
--- a/app/src/services/market_maker.ts
+++ b/app/src/services/market_maker.ts
@@ -141,12 +141,11 @@ class MarketMakerService {
   buy = async (amount: BigNumber, outcome: OutcomeSlot) => {
     const outcomeIndex = outcome === OutcomeSlot.Yes ? 0 : 1
     try {
-      const outcomeTokensToBuy = await this.contract.calcBuyAmount(amount, outcomeIndex, {
+      const outcomeTokensToBuy = await this.contract.calcBuyAmount(amount, outcomeIndex)
+      const transactionObject = await this.contract.buy(amount, outcomeIndex, outcomeTokensToBuy, {
         value: '0x0',
       })
-      await this.contract.buy(amount, outcomeIndex, outcomeTokensToBuy, {
-        value: '0x0',
-      })
+      await this.provider.waitForTransaction(transactionObject.hash)
     } catch (err) {
       logger.error(
         `There was an error buying '${amount.toString()}' for outcome '${outcome}'`,
@@ -159,9 +158,7 @@ class MarketMakerService {
   calcBuyAmount = async (amount: BigNumber, outcome: OutcomeSlot): Promise<BigNumber> => {
     const outcomeIndex = outcome === OutcomeSlot.Yes ? 0 : 1
     try {
-      return this.contract.calcBuyAmount(amount, outcomeIndex, {
-        value: '0x0',
-      })
+      return this.contract.calcBuyAmount(amount, outcomeIndex)
     } catch (err) {
       logger.error(
         `There was an error computing the buy amount for amount '${amount.toString()}' and outcome '${outcome}'`,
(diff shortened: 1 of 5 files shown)


app/src/services/market_maker.ts
```diff
@@ -85,8 +94,12 @@ class MarketMakerService {
...
 
     const totalBalance = balanceOfForYes.add(balanceOfForNo)
-    const actualPriceForYes = divBN(balanceOfForNo, totalBalance)
-    const actualPriceForNo = divBN(balanceOfForYes, totalBalance)
+    const actualPriceForYes = !totalBalance.isZero()
+      ? divBN(balanceOfForNo, totalBalance)
+      : balanceOfForNo.toNumber()
+    const actualPriceForNo = !totalBalance.isZero()
+      ? divBN(balanceOfForYes, totalBalance)
+      : balanceOfForYes.toNumber()
```
mariano-aguero: @fvictorio Please take a look at this carefully

app/src/services/market_maker.ts
```diff
@@ -66,7 +68,12 @@ class MarketMakerService {
     logger.log(`Add funding to market maker ${amount}`)
 
     try {
-      return this.contract.addFunding(amount, distributionHint)
+      const overrides = {
+        gasLimit: 750000,
```
fvictorio: Was this part of a test or do we need to hardcode the gas limit so high?

app/src/services/realitio.ts
```diff
@@ -66,7 +66,7 @@ class RealitioService {
 
     // send the transaction and wait until it's mined
     const transactionObject = await this.contract.askQuestion(...args, {
-      value: ethers.utils.bigNumberify(value),
+      value: '0x0',
```
fvictorio: This is already 0. We need to either remove the `value` parameter or to undo these changes. I'm fine with either option (actually, since this was already tested, maybe it's slightly better to just remove the parameter).

app/src/services/market_maker.ts
fvictorio: The only way `totalBalance` can be zero is if both `balanceForYes` and `balanceForNo` are zero (or one of them is negative, which shouldn't happen). I believe this only happens if the market has no funding. This situation should probably be handled in some other way. But, as a quick fix, I think this fix is OK (maybe we could just use 0 instead of `balanceOfFor{Yes,No}.toNumber()` to make it less obscure).

app/src/services/market_maker.ts
mariano-aguero: We need the gas limit high, also It throws an error

app/src/services/market_maker.ts
mariano-aguero: I will use 0 instead of `balanceOfFor...`

app/src/services/realitio.ts
mariano-aguero: I will remove the parameter

app/src/services/market_maker.ts
mariano-aguero: I remove the gas limit, and tested again with Trust Wallet and Pillar Wallet, and is working fine! So I think this was added to solve a problem with gnosis safe

app/src/services/market_maker.ts
```diff
@@ -177,7 +189,18 @@ class MarketMakerService {
     const outcomeIndex = outcome === OutcomeSlot.Yes ? 0 : 1
     try {
       const outcomeTokensToSell = await this.contract.calcSellAmount(amount, outcomeIndex)
-      await this.contract.sell(amount, outcomeIndex, outcomeTokensToSell)
+
+      const overrides = {
+        gasLimit: 750000,
```
mariano-aguero: I will check this too

PR #231, opened 2019-11-20
Update list of tokens and add CDAI

### LINKED ISSUE #200: Add WETH, OWL and cDAI as tokens to the default list
Order: DAI, cDAI, WETH, USDC, OWL

### LINKED ISSUE #208: Migrate from SAI to DAI
https://changelog.makerdao.com/

mariano-aguero:
Closes #200 
Closes #208 

Addresses for rinkeby => https://changelog.makerdao.com/releases/rinkeby/0.2.14/contracts.json
Addresses for mainnet => https://changelog.makerdao.com/releases/mainnet/1.0.0/contracts.json

### DIFF
--- a/app/src/global.d.ts
+++ b/app/src/global.d.ts
@@ -6,6 +6,6 @@ declare module '@realitio/realitio-lib/formatters/template'
 
 declare type Maybe<T> = T | null
 
-declare type KnownToken = 'sai' | 'usdc' | 'dai' | 'weth' | 'owl'
+declare type KnownToken = 'cdai' | 'usdc' | 'dai' | 'weth' | 'owl'
 
 declare type KnownArbitrator = 'kleros' | 'realitio'
--- a/app/src/global.d.ts
+++ b/app/src/global.d.ts
@@ -6,6 +6,6 @@ declare module '@realitio/realitio-lib/formatters/template'
 
 declare type Maybe<T> = T | null
 
-declare type KnownToken = 'dai' | 'usdc'
+declare type KnownToken = 'sai' | 'usdc' | 'dai' | 'weth' | 'owl'
 
 declare type KnownArbitrator = 'kleros' | 'realitio'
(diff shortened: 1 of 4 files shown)


app/src/global.d.ts
```diff
@@ -6,6 +6,6 @@ declare module '@realitio/realitio-lib/formatters/template'
 
 declare type Maybe<T> = T | null
 
-declare type KnownToken = 'dai' | 'usdc'
+declare type KnownToken = 'sai' | 'usdc' | 'dai' | 'weth' | 'owl'
```
fvictorio: I believe the `sai`/`dai` distinction is just temporary to distinguish between the old and the new token. The app should continue to call it `dai`, and there shouldn't be a `sai` token.

@leolower is this correct?

fvictorio: Besides, this is missing `cDai`.

leolower: Correct, we should replace SAI with DAI. NO need to support both.

mariano-aguero: I remove SAI, and added cDAI :+1:

fvictorio:
The order is not correct:

image

The comment in the issue says: Order: DAI, cDAI, WETH, USDC, OWL.

This *might* work just by moving the entries in the object, but keep in mind that this is [not guaranteed](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries):

> The order of the array returned by Object.entries() does not depend on how an object is defined.

mariano-aguero:
> The order is not correct:
> 
> image
> 
> The comment in the issue says: "Order: DAI, cDAI, WETH, USDC, OWL."
> 
> This _might_ work just by moving the entries in the object, but keep in mind that this is [not guaranteed](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries):
> 
> > The order of the array returned by Object.entries() does not depend on how an object is defined.

Added a sort in the Token component! @fvictorio can you check if your docker is working well with the new tokens ? Thanks

PR #232, opened 2019-11-20
Make sure probabilities always add up to 100 when creating binary markets

fvictorio:
Besides that, this PR:

- Adds `react-testing-library` for testing components.
- Makes the `outcomes` value an array of arbitrary length, but that it's handled as if it always has two elements by the rest of the code.


app/src/components/market/market_wizard_creator.tsx
```diff
@@ -152,13 +162,10 @@ export class MarketWizardCreator extends Component<Props, State> {
             back={() => this.back()}
             next={() => this.next()}
             values={{
-              outcomeProbabilityOne,
-              outcomeProbabilityTwo,
-              outcomeValueOne,
-              outcomeValueTwo,
+              outcomes: marketData.outcomes,
```
mariano-aguero: PP: Maybe you can destructure outcomes var in the line 138, and use it right here

fvictorio: Good catch, fixed.

PR #233, opened 2019-11-20
"See question in realit.io" link

### LINKED ISSUE #192: "See question in realit.io" link

mariano-aguero:
Closes #192


app/src/components/market/profile/view.tsx
```diff
@@ -116,6 +126,21 @@ const ViewWrapper = (props: Props) => {
               ]}
             />
           )}
+          {questionId && (
+            <TitleValue
+              title={'Realitio'}
+              value={[
+                <a
+                  href={`https://realitio.github.io/#!/question/${questionId}`}
+                  key={1}
```
fvictorio: What is this `key` doing here?

mariano-aguero: > What is this `key` doing here?

Mmm I think this `key` is because the value is an array. I'm going to remove the array and the key because we don't need it

PR #234, opened 2019-11-21
Show error message when opening market details without wallet connected

### LINKED ISSUE #205: Show error message when opening market details without wallet connected
- Don't allow to enter market detail screen without web3 provider (currently shows empty page)

mariano-aguero:
Closes #205 

#### Includes
- Check if the user is connected in the market details section
- Added component to show a message if something needs to be notified

Now we have a couple of message notifications ...
Screenshot_20191121_121926
Screenshot_20191121_121954
Screenshot_20191121_122039


app/src/components/market/market_routes.tsx
```diff
@@ -55,4 +58,25 @@ const MarketRoutes = (props: RouteComponentProps<RouteParams>) => {
...
+      message="Please connect to your wallet to open the market details..."
+    />
+  )
+}
+
+const MarketRoutes = (props: RouteComponentProps<RouteParams>) => {
+  const { active } = useWeb3Context()
+  const connector = localStorage.getItem('CONNECTOR')
+
+  return active || (connector && connector in connectors) ? (
```
fvictorio: Shouldn't this be `&&`?

fvictorio:
My problem with this is that these components look like toasts but they don't disappear and can't be dismissed. And it makes sense, because it's a message that should be kept in the screen. But in that case, why make them look like toasts?

Just a comment since this is a design thing, not something to block the PR.

mariano-aguero:
> My problem with this is that these components look like toasts but they don't disappear and can't be dismissed. And it makes sense, because it's a message that should be kept in the screen. But in that case, why make them look like toasts?
> 
> Just a comment since this is a design thing, not something to block the PR.

@fvictorio What about if we can click on the message and close it ? Do you agree on do that ?

fvictorio:
If you ask me, this should be a plain message that doesn't look like a toast. To me, a toast is an informative message that can disappear or be dismissed because it's temporary and/or because it hinders the rest of the content of the page.

In this case the situation is not temporary (you are not connected) and there's no content to hinder, that's why it looks so weird.

Then again: not my area of expertise, so I'm fine with whatever you choose to do.

gabitoesmiapodo:
> If you ask me, this should be a plain message that doesn't look like a toast. To me, a toast is an informative message that can disappear or be dismissed because it's temporary and/or because it hinders the rest of the content of the page.
> 
> In this case the situation is not temporary (you are not connected) and there's no content to hinder, that's why it looks so weird.
> 
> Then again: not my area of expertise, so I'm fine with whatever you choose to do.

These are just general purpose toasts, if we need something more specific for this particular use case we should think and create a new design (which we obviously don't have) and implementation.

The reasonable thing to do here (besides a more specific design and implementation) should be to make the component un-dismissable if the problem requires it, ie: annoy the user until he or she solves the problem.

If the problem can be solved with what we already have UI wise, do that.

app/src/components/market/market_routes.tsx
mariano-aguero: Done

app/src/components/market/market_routes.tsx
```diff
@@ -39,7 +42,7 @@ const MarketValidateContractAddress: React.FC<Props> = (props: Props) => {
   )
 }
 
-const MarketRoutes = (props: RouteComponentProps<RouteParams>) => {
+const MarketRoutesConnectedWrapper = (props: any) => {
```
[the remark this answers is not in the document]
mariano-aguero: Done

mariano-aguero:
Added an `onClick` props to close the message...

PR #235, opened 2019-11-21
Add "Resolve condition" button

fvictorio:
This PR adds a "Resolve condition" button in the market details page when the question is finalized in Realitio but the condition is not yet resolved in ConditionalTokens.

I also refactored `useMarketMakerData` to move some of its logic to a `usePolling` hook.


fvictorio:
@gabitoesmiapodo please check 910dfbd7770472bc71ec214b889113218d2c07ef

app/src/components/market/profile/view.tsx
```diff
@@ -140,6 +152,7 @@ const ViewWrapper = (props: Props) => {
             </ButtonAnchor>
           )}
           <ButtonAnchor href={`/#/${marketMakerAddress}/buy`}>Buy</ButtonAnchor>
+          {isQuestionFinalized && <Button onClick={onResolveCondition}>Resolve condition</Button>}
```
mariano-aguero: Q: Do we need to check here if it is not already resolved?

fvictorio: Since this is in the page that is shown when the condition is not resolved, it's not necessary to also check that.

PR #236, opened 2019-11-22
Hide markets with a different free from the one used by the app

fvictorio:
Markets whose fee is different from the one declared in `constants` will no be shown in the list of markets in the home page. If the details page of such a market is visited, a message is shown indicating that it's an invalid market and the user can't interact with that market.

Also: there were several places where arrays of known types where mapped or filtered, and the callback had type annotations. Since this is not necessary, I removed them. Ideally there should be a linter rule for that, but there isn't one yet. To find these scenarios, you can use [grep-ast](https://github.com/suchipi/grep-ast) like this (from the `app` directory):

```
grep-ast --selector 'CallExpression[callee.property.name="map"]>ArrowFunctionExpression>Identifier[typeAnnotation]' --parser '@typescript-eslint/parser'  --patterns 'src/**/*.ts src/**/*.tsx' --parserOptions '{"sourceType": "module", "ecmaFeatures": { "jsx": true, "modules": true }}'
```

There is still one result, but it's using an array of `any` so it makes sense to narrow down the type with an annotation.


mariano-aguero:
@fvictorio Looks good, can you resolve the conflicts? Feel free to merge after that ...

Issue #238, opened 2019-11-22
See my collateral balance and max link


mariano-aguero:
This task is for UI design, to improve the actual behaviour.

mariano-aguero:
See tickets #195 and #196

gabitoesmiapodo:
`Max` button is here: https://zpl.io/aRXnx8N

![02 - Conditional Exchange - Create 2](https://user-images.githubusercontent.com/4015436/70467051-43572f00-1aa3-11ea-858a-108a52536049.png)

gabitoesmiapodo:
`Max` buttons for two more sections: https://zpl.io/a8J8716 - https://zpl.io/aRXnx5v

![09 - Conditional Exchange - Buy](https://user-images.githubusercontent.com/4015436/70468832-8e267600-1aa6-11ea-95a7-4089e86d621a.png)
![10 - Conditional Exchange - Sell](https://user-images.githubusercontent.com/4015436/70468834-8ebf0c80-1aa6-11ea-8f01-9ab474498c32.png)

PR #239, opened 2019-11-22
See my collateral balance

### LINKED ISSUE #195: See my collateral balance

### LINKED ISSUE #196: Add MAX links
Buy and Fund uses DA balance. Sell uses Shares.

mariano-aguero:
Closes #195 
Closes #196 

#### Includes
- Component to display the balance of the selected token in the sections: create, buy, fund
- Component to display the balance of the selected outcome in the sell section

#### Does not include
- Design, (I create a ticket for this #238)

Some images ...
Screenshot_20191122_172703
Screenshot_20191122_172836
Screenshot_20191122_172933
Screenshot_20191122_172857


mariano-aguero:
@fvictorio Do you think we need to add a validation in the second step so that the user cannot add more collateral than he has?

app/src/components/common/balance_token/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  const calculateBalanceAmount = useMemo(
+    () => async (): Promise<[BigNumber, string, string]> => {
+      const collateralService = new ERC20Service(collateral.address)
+
+      const collateralBalance = await collateralService.getCollateral(account, library)
+      return [
+        collateralBalance,
+        formatBigNumber(collateralBalance, collateral.decimals),
+        collateral.symbol,
```
fvictorio: Why is the symbol and the formatted value returned by this function? They don't depend on the values that this function use, nor need to be computed asynchronously, they just pollute the function signature.

app/src/components/common/big_number_input/index.tsx
```diff
@@ -25,6 +25,7 @@ interface Props {
 
 interface State {
   currentValueStr: string
+  currentDecimalsStr: number
```
fvictorio: I don't get this. Before `this.props.decimals` was being used. Now there's this state variable that's only updated when `this.props.decimals` change. It seems to me like it should have the same behavior as before. What am I missing here?

app/src/components/market/market_buy.tsx
```diff
@@ -148,6 +149,30 @@ const MarketBuyWrapper: React.FC<Props> = (props: Props) => {
 
   const disabled = (status !== Status.Ready && status !== Status.Error) || cost.isZero()
 
+  const noteBalanceCollateral = () => {
+    return (
+      <React.Fragment key="1">
```
fvictorio: Unnecessary key. I'm guessing you added it because React emitted a warning when this was used in an array of components. But that `note` prop should accept any react node (right now accepts an `any`, which it should be more specific). So the solution should be simple: pass a component with these two other components instead of an array.

mariano-aguero: Grouped the two components into one

app/src/components/common/balance_token/index.tsx
mariano-aguero: I can code it this way (that is the logical , the right approach probably and what can you expect in this case) ...

```
export const BalanceToken = (props: Props) => {

  const { networkId, account, library } = context
  const { collateralId, onClickMax } = props

  const collateral = getToken(networkId, collateralId)

  const calculateBalanceAmount = useMemo(
    () => async (): Promise<BigNumber> => {
      const collateralService = new ERC20Service(collateral.address)

      return collateralService.getCollateral(account, library)
    },
    [account, library, collateral],
  )

  const collateralBalanceAmountValue = useAsyncDerivedValue(
    '',
    new BigNumber(0),
    calculateBalanceAmount,
  )

  return (
    <Wrapper>
      <BalanceTitle>
        Balance {formatBigNumber(collateralBalanceAmountValue, collateral.decimals)} {collateral.symbol}.
      </BalanceTitle>
      <ButtonLink onClick={() => onClickMax(collateral, collateralBalanceAmountValue)}>Max</ButtonLink>
    </Wrapper>
  )
}
```

But if I do in this way, this is how will work:
https://www.loom.com/share/54775208575d4bf4a2935bf95518665b
You can see how the symbol and the decimals used is not in sync with calculated value.

The approach used right now work like this...
https://www.loom.com/share/13f1ad028099436d8f37baff283ea34f

I know that they don't depend on the values that this function use, but in this case it solves an small UI problem that is annoying

app/src/components/common/big_number_input/index.tsx
mariano-aguero: If we have tokens with different decimals, the component will throws an exception

fvictorio: I'm talking about something internal to the component. Using a prop directly, or setting a state when the prop changes and then using that state, should be equivalent (but the later adds unnecessary complexity). If using the prop directly throws an error, then we should investigate why that happen, because it shouldn't.

app/src/components/common/balance_token/index.tsx
fvictorio: Oh, I see :+1:

PR #244, opened 2019-11-25
Use hasOwnProperty when is disconnected

mariano-aguero:
Use `hasOwnProperty` when is disconnected, to avoid to use `getSigner` function. See https://eslint.org/docs/rules/no-prototype-builtins for proper use


fvictorio:
I don't like this. Let's think of a better way to handle this because it's adding a lot of noise to the code. Maybe wrapping the provider?

fvictorio:
Another alternative: instead of passing the provider, we can pass the context. I would also like to have a single context, and having a `Maybe<string>` as the account. Then you can check that to see if it's a Wallet context or a Infura context.

mariano-aguero:
> Another alternative: instead of passing the provider, we can pass the context. I would also like to have a single context, and having a `Maybe<string>` as the account. Then you can check that to see if it's a Wallet context or a Infura context.

I will work on this approach, Thanks for the feedback

mariano-aguero:
I Will close this, In favor of https://github.com/protofire/gnosis-conditional-exchange/pull/250

PR #246, opened 2019-11-26
Add "Funded markets" filter in markets list

fvictorio:
Some comments on this:

- I added a `RemoteData` type. Right now it's only being used in `useMarkets`.
- I merged the `market_home_connected_container` and `market_home_disconnected_container` components, they were doing pretty much the same thing.
- For now, all the filters cause a reload. This is wasteful when moving from "All markets" to "My Markets", since that filtering could be done locally. Since this is the list of markets, I don't think it's a big deal, but it can be optimized later if needed.


fvictorio:
Ugh, thanks. I made a very ugly, very temporary fix.

PR #251, opened 2019-11-29
Infinite scroll

fvictorio:
This has several pending issues:

- The infinite scroll it's not a scroll. Instead, a "Show more" button is used.
- List behaves wrongly after the user connects his wallet.


app/src/components/market/market_home.tsx
```diff
@@ -38,11 +42,16 @@ export const MarketHome: React.FC<Props> = (props: Props) => {
         />
       )}
       <ListCard>
-        {RemoteData.is.success(markets)
-          ? markets.data.map((item, index) => {
-              return <ListItem key={index} data={item}></ListItem>
+        {RemoteData.hasData(markets)
+          ? markets.data.slice(0, props.count).map(item => {
```
mariano-aguero: I think we can take advantage of the destructuring in the line 31, and add `count` and `onShowMore` to the destructuring

app/src/services/market_maker_factory.ts
```diff
@@ -51,13 +60,14 @@ class MarketMakerFactoryService {
     return marketMakerAddress
   }
 
-  getMarkets = async (provider: any): Promise<Market[]> => {
+  getMarkets = async ({ from, to }: GetMarketsOptions): Promise<Market[]> => {
+    console.log(`fetching from ${from} to ${to}`)
```
mariano-aguero: Should be `logger.log`

app/src/hooks/useMarkets.tsx
```diff
@@ -1,113 +1,154 @@
...
-import { MarketMakerService } from '../services/market_maker'
-import { getLogger } from '../util/logger'
-import { Market, MarketWithExtraData, MarketFilters, MarketStatus } from '../util/types'
+import { useContracts, Contracts } from './useContracts'
+import { MarketWithExtraData, MarketFilters } from '../util/types'
+import { callInChunks, Range } from '../util/call_in_chunks'
 import { RemoteData } from '../util/remote_data'
 
-const logger = getLogger('Market::useMarkets')
+const EARLIEST_BLOCK_TO_CHECK = 4986777
```
mariano-aguero: This should be an env var , we can added this value `4986777` as a default value if the env var doesn't exist.

app/src/components/market/market_home.tsx
fvictorio: Just for the record, I kind of like to use `props.foo` directoy when `foo` is only used once. That being said, I changed it for consistency with what we're already doing.

app/src/hooks/useMarkets.tsx
fvictorio: Done. Also added the chunk size as an env var.

app/src/services/market_maker_factory.ts
fvictorio: Done.

fvictorio: Also added the `no-console` rule to our eslint config btw.

PR #257, opened 2019-12-04
Add custom collateral

### LINKED ISSUE #198: Allow to fund with an ERC20 token as collateral
Component to set new token + validation.

We need to add the possibility to the user to add a custom token.
- Add an address for a custom token
- Search for the name, symbol, and decimals of the custom token
- Add this custom token to the dropdown an mark it as selected
- Display the custom token on all pages

mariano-aguero:
Closes #198 

#### Includes
- Add useCollateral hook
- Add modal to select and address as collateral
- Refactoring create market process using a custom collateral
- Update BigNumberInput to hooks
- Fix signer problem in some methods
- Allow to maintain in the create market process the custom collateral that the user is adding.

#### Does not includes
- Refactoring using a context

Some token address in rinkeby:
https://rinkeby.etherscan.io/token/0x577d296678535e4903d59a4c929b718e1d575e0a
https://rinkeby.etherscan.io/token/0xddea378a6ddc8afec82c36e9b0078826bf9e68b6

Faucet for this tokens in rinkeby:
https://app.compound.finance/

Some images  with ZRX as collateral...
Screenshot_20191204_175358
Screenshot_20191204_175411
Screenshot_20191204_175443
Screenshot_20191204_175834
Screenshot_20191204_175745
Screenshot_20191204_175848
Screenshot_20191204_175903

Sorry the huge refactoring, but I needed it to make some things work


app/src/util/types.ts
```diff
@@ -125,3 +125,15 @@ export enum Wallet {
   MetaMask = 'MetaMask',
   WalletConnect = 'WalletConnect',
 }
+
+export interface Collateral {
```
fvictorio: This is almost equal to the `Token` interface. Couldn't we reuse that? `name` might be optional so that existing usages of `Token` don't break.

app/src/components/market/market_wizard_creator.tsx
```diff
@@ -38,97 +39,97 @@ interface State {
...
   }
 
-  public handleChange = (
-    event: ChangeEvent<HTMLInputElement> | ChangeEvent<HTMLSelectElement> | BigNumberInputReturn,
+  const handleChange = (
+    event:
+      | ChangeEvent<HTMLInputElement>
+      | ChangeEvent<HTMLSelectElement>
+      | BigNumberInputReturn
+      | CollateralCustomEvent,
```
fvictorio: This is getting out of hand IMO. I would just use a custom event handler like we do below with `handleOutcomesChange` and `handleDateChange`.

app/src/components/market/market_wizard_creator_container.tsx
```diff
@@ -32,7 +32,16 @@ const MarketWizardCreatorContainer: FC = () => {
       const { collateralId, arbitratorId, question, resolution, funding, outcomes, category } = data
       const openingDateMoment = moment(resolution)
 
-      const collateralToken = getToken(networkId, collateralId)
+      const erc20Service = new ERC20Service(provider, collateralId)
+      const isValidErc20 = await erc20Service.isValidErc20()
+
+      let collateralToken: Maybe<Token> = null
```
fvictorio:
```suggestion
      let collateralToken: Token
```

app/src/hooks/useCollateral.tsx
```diff
@@ -0,0 +1,49 @@
...
+    const fetchIsValidErc20 = async () => {
+      let collateralData: Maybe<Token> = null
+      const erc20Service = new ERC20Service(context.library, collateralId)
+      const isValidErc20 = await erc20Service.isValidErc20()
+
+      if (isValidErc20) {
+        const data = await erc20Service.getProfileSummary()
+        collateralData = {
+          ...data,
+        } as Token
```
fvictorio: More reason to merge the `Collateral` and `Token` interfaces.

app/src/hooks/useCollateral.tsx
```diff
@@ -0,0 +1,49 @@
...
+      } else {
+        try {
+          collateralData = getToken(context.networkId, collateralId as KnownToken)
+        } catch (err) {
+          logger.error(err.message)
```
fvictorio: Let's add some context here, something like "Collateral token couldn't be fetched and it's not known"

app/src/hooks/useCollateral.tsx
```diff
@@ -0,0 +1,49 @@
...
+  collateralId: KnownToken | string,
+  context: ConnectedWeb3Context,
+): Maybe<Token> => {
+  const [collateral, setCollateral] = useState<Maybe<Token>>(null)
+
+  useEffect(() => {
+    let isSubscribed = true
+
+    const fetchIsValidErc20 = async () => {
+      let collateralData: Maybe<Token> = null
```
fvictorio: I guess you are using this name to avoid shadowing the `collateral` variable from `useState`. We don't have a linter rule forbidding that, and in cases like this I think shadowing is fine. But this is just a PP.

app/src/services/erc20.ts
```diff
@@ -2,32 +2,38 @@ import { ethers, Wallet } from 'ethers'
...
+  'function name() external view returns (string)',
+  'function decimals() external view returns (uint8)',
 ]
 
 class ERC20Service {
   tokenAddress: string
+  provider: any
 
-  constructor(tokenAddress: string) {
+  constructor(provider: any, tokenAddress: string) {
```
fvictorio: I'd rather have the address as the first parameter, like in the other services.

app/src/components/common/tokens/index.tsx
```diff
@@ -29,8 +31,24 @@ export const Tokens = (props: Props) => {
...
+    )
+    if (!isTokenInKnowTokens) {
+      tokens.push({
+        label: collateral.symbol,
+        value: collateral.address,
+      })
+    }
+  }
+
+  const selectedValue = { value: isTokenInKnowTokens ? value.symbol.toLowerCase() : value.address }
```
fvictorio: I think doing:

```typescript
  const selectedValue = isTokenInKnowTokens ? value.symbol.toLowerCase() : value.address

  return (
    <Select {...restProps} value={selectedValue}>
```

would be more readable.

app/src/components/common/tokens/index.tsx
```diff
@@ -29,8 +31,24 @@ export const Tokens = (props: Props) => {
       value: id,
     }))
 
+  const isTokenInKnowTokens = tokens.find(token => token.label === value.symbol)
```
fvictorio: If we merge `Collateral` and `Token`, and make a `getKnownTokens(networkId)` function, then maybe we can drop all of this and just do something like this above:

```typescript
const tokens = Object.entries([...getKnownTokens(networkId), ...customValues])
```

app/src/components/common/big_number_input/index.tsx
```diff
@@ -41,92 +41,59 @@ const Input = styled.input`
...
-    } else if (
-      value &&
-      !ethers.utils.parseUnits(currentValueStr || '0', currentDecimalsStr).eq(value)
-    ) {
-      return {
-        currentValueStr: ethers.utils.formatUnits(value, decimals),
-      }
-    } else {
-      return null
+export const BigNumberInput = (props: Props) => {
```
fvictorio: This doesn't have the exact same behavior as before. For example, if the value is `0.0`, and you press backspace, nothing happens. That's why in the previous implementation there was a state variable `currentValueStr`: you need to keep track of the string in the input, and transform it in a BigNumber for these scenarios to work.

app/src/components/market/steps/items/funding_and_fee_step.tsx
```diff
@@ -84,49 +90,55 @@ const FundingAndFeeStep = (props: Props) => {
...
-            onClickMax={(collateral: Token, collateralBalance: BigNumber) => {
-              props.handleChange({ name: 'funding', value: collateralBalance })
-            }}
-          />
-        }
-      />
+      {collateral && (
+        <FormRow
+          formField={
+            <TokensAddAnotherCustom
```
fvictorio: Can we rename this to something less Yoda-ish? Maybe `AddCustomToken`?

app/src/components/common/tokens_add_another_custom/modal_collateral.tsx
```diff
@@ -0,0 +1,125 @@
...
+  context: ConnectedWeb3Context
+}
+
+const ModalCollateralWrapper = (props: Props) => {
+  const { theme, onClose, onSave, context } = props
+
+  const [collateralAddress, setCollateralAddress] = useState<string>('')
+
+  const collateralData = useCollateral(collateralAddress, context)
+  const collateralError = !collateralData
```
fvictorio: PP: I would just inline this in the disabled prop.

app/src/hooks/useCollateral.tsx
mariano-aguero: Done!

app/src/util/types.ts
mariano-aguero: Done!

PR #265, opened 2019-12-10
Add support for mutiple outcomes on market creation

mariano-aguero:
#### Includes
- Add more outcomes when creates a market
- Update realitio service to use more than one outcome
- There is an env to set the maximum amount of outcomes users can add when creating markets

#### Not includes
- Probabilities in the summary market step, we need to calculate this with the formula for the actual price


app/src/components/market/steps/items/outcomes_step.tsx
```diff
@@ -29,17 +48,54 @@ const OutcomesStep = (props: Props) => {
...
+    (emptyOutcomes: Outcome[] = [], cur: Outcome) => {
+      if (!cur.name) emptyOutcomes.push(cur)
+      return emptyOutcomes
+    },
+    [],
+  ).length
+  const messageOutcomeEmptyNamesError =
+    totalOutcomesWithEmptyNames > 0 ? 'The names of the outcomes should not be empty.' : ''
+
+  const totalOutcomeProbabilities = outcomes.reduce((prev, cur) => prev + cur.probability, 0)
```
fvictorio: Again, this should be called `acc` or `total`, not `prev`.

app/src/components/market/steps/items/outcomes_step.tsx
```diff
@@ -29,17 +48,54 @@ const OutcomesStep = (props: Props) => {
...
         Please add all the possible outcomes for the <strong>&quot;{question}&quot;</strong>{' '}
         question.
       </OutcomeInfo>
-      <Outcomes outcomes={outcomes} onChange={handleOutcomesChange} />
-      <ButtonContainer>
+      <Outcomes
+        outcomes={outcomes}
+        onChange={handleOutcomesChange}
+        messageEmptyNamesError={messageOutcomeEmptyNamesError}
+        messageProbabilitiesError={messageOutcomeProbabilitiesError}
```
fvictorio: I'm fine with this as long as there are only two props for it. If at some point we need to add a third possible error condition, we'll need to think of a different approach.

app/src/components/common/outcomes/index.tsx
```diff
@@ -71,9 +83,33 @@ const Outcomes = (props: Props) => {
...
+        <ErrorStyled>{messageProbabilitiesError}</ErrorStyled>
+        <ErrorStyled>{messageEmptyNamesError}</ErrorStyled>
+      </>
+    )
+  }
+
   const outcomesToRender = props.outcomes.map((outcome: Outcome, index: number) => (
     <TwoColumnsRowExtraMargin key={index}>
-      <Textfield name={`outcome_${index}`} type="text" value={outcome.name} readOnly />
+      <Textfield
```
fvictorio: We'll need to update the tests for this component.

app/src/services/realitio.ts
```diff
@@ -95,19 +103,27 @@ class RealitioService {
...
     const questionLog: QuestionLog = RealitioQuestionLib.populatedJSONForTemplate(
-      RealitioTemplateLib.defaultTemplateForType('bool'),
-      event.values.question,
+      RealitioTemplateLib.defaultTemplateForType(templateId),
+      question,
     )
 
-    const category = questionLog.category === 'undefined' ? '' : questionLog.category
-    const question = questionLog.title === 'undefined' ? '' : questionLog.title
+    const { category, title, outcomes = [OutcomeSlot.Yes, OutcomeSlot.No] } = questionLog
```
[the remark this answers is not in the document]
mariano-aguero: Mmm I think we can support them using the `templateId` `SINGLE_SELECT_TEMPLATE_ID` with two outcomes (like the default used right here, is simple and gives the user and example of how to use it), Right now the mechanism to parse the log is using the `templateId` so this should work

app/src/components/market/steps/items/outcomes_step.tsx
```diff
@@ -29,17 +48,54 @@ const OutcomesStep = (props: Props) => {
   const { handleOutcomesChange, values } = props
   const { question, outcomes } = values
 
+  const totalOutcomesWithEmptyNames = outcomes.reduce(
+    (emptyOutcomes: Outcome[] = [], cur: Outcome) => {
```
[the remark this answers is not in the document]
mariano-aguero: Done!

app/src/components/common/outcomes/index.tsx
mariano-aguero: I added more tests

app/src/services/realitio.ts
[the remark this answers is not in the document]
fvictorio: Yeah, it's possible but it'll probably add more work that might not be needed right now. I think just filtering them out is easier. (cc @leolower)

app/.eslintrc.json
```diff
@@ -19,6 +19,7 @@
     "react/prop-types": "off",
     "no-console": "warn",
     "no-warning-comments": "warn",
+    "camelcase": "off",
```
fvictorio: This seems like it can be removed (I removed it and it didn't trigger any error).

app/src/components/common/outcomes/index.tsx
```diff
@@ -96,8 +95,11 @@ const Outcomes = (props: Props) => {
   const messageErrorToRender = () => {
     return (
       <>
-        <ErrorStyled>{messageProbabilitiesError}</ErrorStyled>
-        <ErrorStyled>{messageEmptyNamesError}</ErrorStyled>
+        {errorMessages.map((errorMessage, index) => (
```
fvictorio: If I'm getting this right, `errorMessages` will be `['', '']` (an array with two empty strings) when everything is correct. So these two components will be rendered, albeit invisible. I'm fine with this, just making sure I'm understanding it.

mariano-aguero: Mmm yes, you getting this right, but know that you say this, I don't like it, so I change the mechanim, now is not render empty strings anymore.

PR #270, opened 2019-12-12
Add possibility to render probabilities without distribution hint

mariano-aguero:
This is how is displayed...

Screenshot_20191213_132527
Screenshot_20191213_132516

Added some TODO comments in some critical lines of code, where we need to refactoring

### DIFF
--- a/app/src/components/common/outcome_table/index.tsx
+++ b/app/src/components/common/outcome_table/index.tsx
@@ -159,15 +159,12 @@ export const OutcomeTable = (props: Props) => {
     )
   }
 
-  const yesRow = renderTableRow(balance[0], pricesAfterTrade && pricesAfterTrade[0])
-  const noRow = renderTableRow(balance[1], pricesAfterTrade && pricesAfterTrade[1])
+  const renderTable = () =>
+    balance
+      .sort((a, b) => (a.winningOutcome === b.winningOutcome ? 0 : a.winningOutcome ? -1 : 1))
+      .map((balanceItem: BalanceItem, index) =>
+        renderTableRow(balanceItem, pricesAfterTrade && pricesAfterTrade[index]),
+      )
 
-  let rows = [yesRow, noRow]
-  if (withWinningOutcome) {
-    if (balance[1].winningOutcome) {
-      rows = [noRow, yesRow]
-    }
-  }
-
-  return <TableStyled head={renderTableHeader()}>{rows}</TableStyled>
+  return <TableStyled head={renderTableHeader()}>{renderTable()}</TableStyled>
 }
--- a/app/src/components/common/outcome_table/index.tsx
+++ b/app/src/components/common/outcome_table/index.tsx
@@ -8,7 +8,7 @@ import { BalanceItem, OutcomeSlot, OutcomeTableValue, Token } from '../../../uti
 import { BarDiagram } from '../bar_diagram_probabilities'
 
 interface Props {
-  balance: BalanceItem[]
+  balances: BalanceItem[]
   collateral: Token
   pricesAfterTrade?: [number, number]
   outcomeSelected?: OutcomeSlot
@@ -44,7 +44,7 @@ const RadioInputStyled = styled(RadioInput)`
 
 export const OutcomeTable = (props: Props) => {
   const {
-    balance,
+    balances,
     collateral,
     pricesAfterTrade,
     outcomeSelected,
@@ -160,7 +160,7 @@ export const OutcomeTable = (props: Props) => {
   }
 
   const renderTable = () =>
-    balance
+    balances
       .sort((a, b) => (a.winningOutcome === b.winningOutcome ? 0 : a.winningOutcome ? -1 : 1)) // Put winning outcome first
       .map((balanceItem: BalanceItem, index) =>
         renderTableRow(balanceItem, pricesAfterTrade && pricesAfterTrade[index]),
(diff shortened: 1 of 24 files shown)


app/src/components/common/outcome_table/index.tsx
```diff
@@ -159,15 +159,12 @@ export const OutcomeTable = (props: Props) => {
...
-  if (withWinningOutcome) {
-    if (balance[1].winningOutcome) {
-      rows = [noRow, yesRow]
-    }
-  }
-
-  return <TableStyled head={renderTableHeader()}>{rows}</TableStyled>
+  const renderTable = () =>
+    balance
+      .sort((a, b) => (a.winningOutcome === b.winningOutcome ? 0 : a.winningOutcome ? -1 : 1))
```
mariano-aguero: This sort exist because we need to order the array by winningOutcome, is a boolean

app/src/services/conditional_token.ts
```diff
@@ -124,7 +131,7 @@ class ConditionalTokenService {
       collateralToken,
       ethers.constants.HashZero,
       conditionId,
-      [1, 2],
+      [1, 2], // TODO: analize if we need to refactor this
```
fvictorio: I think this needs to be replaced with a `indexSets` variable, that is computed like this:

```typescript
const range = length => [...Array(length)].map((x, i) => i)
const indexSets = range(outcomesCount).map(x => 1 << x)
```

I think this comes from the fact that for N outcomes, you can describe a position with a bit array of length N, where something like `1010` means the second and fourth outcomes. But since we don't have conditional positions yet, all our index sets have only one 1. For two outcomes, the valid index sets are `01` and `10` (the `[1, 2]`) array that is being used right now). For N outcomes, you can use the bit shift operator to generate all of them.

app/src/services/market_maker.ts
```diff
@@ -143,6 +146,42 @@ class MarketMakerService {
...
+    const collateralTokenAddress = await this.getCollateralToken()
+
+    const balances = []
+    logger.log(`Outcomes quantity ${outcomeQuantity}`)
+    for (let i = 0; i < outcomeQuantity; i++) {
+      const collectionId = await this.conditionalTokens.getCollectionIdForOutcome(
+        conditionId,
+        1 << i,
+      )
+      logger.log(
```
fvictorio: These logs are too detailed, I don't think they should be at the `log` level. Maybe `debug` is more appropriate.

app/src/components/common/outcome_table/index.tsx
fvictorio: Got it, let's add a comment saying something like `Put winning outcome first`.

fvictorio: Btw, an alternative way of doing this is to use [`_.partition`](https://lodash.com/docs/4.17.15#partition):

```typescript
const [winningOutcomes, losingOutcomes] = _.partition(balances, 'winningOutcome')
if (winningOutcomes.length > 1) {
  logger.warning('There should be only one winning outcome')
}
const sortedOutcomes = [...winningOutcomes, ...losingOutcomes]
```

mariano-aguero: @fvictorio I will add the comment, regarding lodash, we need to install it and maybe is an overkill, what do you think ?

app/src/services/market_maker.ts
```diff
@@ -84,6 +84,7 @@ class MarketMakerService {
     try {
       const overrides = {
         value: '0x0',
+        gasLimit: 750000,
```
mariano-aguero: Please notice this, If we don't add this gasLimit the add funding transaction will fail

app/src/components/market/steps/items/summary_market_step.tsx
```diff
@@ -58,9 +58,9 @@ const SummaryMarketStep = (props: Props) => {
 
   const { marketMakerAddress } = props
 
-  const { question, resolution, category, outcomes } = useQuestion(marketMakerAddress, context)
+  const { question, resolution, category } = useQuestion(marketMakerAddress, context)
   const { marketMakerData } = useMarketMakerData(marketMakerAddress, context)
-  const { marketMakerFunding, collateral, arbitrator } = marketMakerData
+  const { marketMakerFunding, collateral, arbitrator, balance } = marketMakerData
```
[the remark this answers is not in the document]
mariano-aguero: Done!

app/src/hooks/useMarketMakerData.tsx
```diff
@@ -99,33 +103,29 @@ export const useMarketMakerData = (
...
-        outcomeName: OutcomeSlot.No,
-        probability: Math.round((probabilityForNo / 100) * 100),
-        currentPrice: actualPrices.actualPriceForNo,
-        shares: userShares.balanceOfForNo,
-        holdings: marketMakerShares.balanceOfForNo,
-        winningOutcome: winnerOutcome === 0,
-      },
-    ]
+    const balance: BalanceItem[] = []
+    for (let i = 0; i < outcomes.length; i++) {
```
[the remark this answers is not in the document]
mariano-aguero: Change it

app/src/services/market_maker.ts
mariano-aguero: Done!

app/src/components/common/outcome_table/index.tsx
fvictorio: It's OK the way it is.

PR #279, opened 2019-12-18
Fix some feedback for milestone 5

### LINKED ISSUE #277: The initial probabilities for an outcome cannot be zero

mariano-aguero:
Closes #277 

### Includes
- Improve messages when a user redeems and resolve condition
- Add validation in the step 2 in the market wizard creation for probabilities

### DIFF
--- a/app/src/components/market/market_home.tsx
+++ b/app/src/components/market/market_home.tsx
@@ -62,7 +62,7 @@ export const MarketHome: React.FC<Props> = (props: Props) => {
         {RemoteData.hasData(markets) &&
           markets.data.length > 0 &&
           markets.data.slice(0, count).map(item => {
-            return <ListItem key={item.conditionId} data={item}></ListItem>
+            return <ListItem key={`${item.address}_${item.conditionId}`} data={item}></ListItem>
           })}
         {RemoteData.is.success(markets) && markets.data.length === 0 && (
           <NoMarketsAvailable title={'No markets available'} />
--- a/app/src/components/market/profile/closed_market_detail.tsx
+++ b/app/src/components/market/profile/closed_market_detail.tsx
@@ -83,9 +83,19 @@ export const ClosedMarketDetailWrapper = (props: Props) => {
 
   const marketMaker = buildMarketMaker(marketMakerAddress)
 
-  const resolveCondition = () => {
-    // Balances length is the number of outcomes
-    return oracle.resolveCondition(questionId, balances.length)
+  const resolveCondition = async () => {
+    try {
+      setStatus(Status.Loading)
+      setMessage('Resolve condition...')
+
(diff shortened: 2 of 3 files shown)


app/src/components/market/market_home.tsx
```diff
@@ -62,7 +62,7 @@ export const MarketHome: React.FC<Props> = (props: Props) => {
         {RemoteData.hasData(markets) &&
           markets.data.length > 0 &&
           markets.data.slice(0, count).map(item => {
-            return <ListItem key={item.conditionId} data={item}></ListItem>
+            return <ListItem key={`${item.address}_${item.conditionId}`} data={item}></ListItem>
```
fvictorio: This should be fine but I'm curious, how did you get two outcomes with the same `conditionId` (assuming that's why you did this).

app/src/components/market/profile/closed_market_detail.tsx
```diff
@@ -109,13 +119,14 @@ export const ClosedMarketDetailWrapper = (props: Props) => {
...
       if (!isConditionResolved) {
         setMessage('Resolving condition...')
-        await resolveCondition()
+        // Balances length is the number of outcomes
+        await oracle.resolveCondition(questionId, balances.length)
       }
-      setMessage('Redeem payout...')
-      setStatus(Status.Loading)
 
+      setMessage('Redeem payout...')
```
fvictorio: Should we also add a `setMessage('Resolving condition...')` line before the if? Or maybe just move this line before the if?

app/src/components/market/market_home.tsx
mariano-aguero: Mmmm this is happening with rinkeby, you can check the screenshot, Why is this happening ? I think we need to investigate more deeply...

Screenshot_20191219_103141

app/src/components/market/profile/closed_market_detail.tsx
mariano-aguero: I think the message should be added inside the `if` , is only needed when the condition is not resolved. Mmmm you make me think why we need to put this message before the `if` , do you have any concern about ? Please let me know, your thoughts are really appreciated. This is how it looks now...

Screenshot_20191219_103405

I tested and works pretty good

app/src/components/market/market_home.tsx
fvictorio: Weird. Well, it's OK then.

app/src/components/market/profile/closed_market_detail.tsx
fvictorio: Yeah, inside the `if` is better. And it's just because the user will get 2 transactions, and so I think it's better to explain what each one is doing.

PR #288, opened 2019-12-19
Show timezone in the resolution date input when creating new markets

### LINKED ISSUE #281: Show timezone in the resolution date input when creating new markets

mariano-aguero:
Closes #281 

I create a market with the timezone from Phoenix Arizona ( UTC - 7 )
This is the screenshot ...
Screenshot_20191219_090634

And this is how it looks in the market profile ...
Screenshot_20191219_090743

And this is the market profile with Buenos Aires timezone
Screenshot_20191219_130916

The difference between the timezones is four (4) and in the market profile is OK for the buenos aires timezone.

### DIFF
--- a/app/src/components/common/date_field/index.tsx
+++ b/app/src/components/common/date_field/index.tsx
@@ -1,7 +1,9 @@
 import React from 'react'
 import styled from 'styled-components'
-import { IconCalendar } from './img/IconCalendar'
 import DatePicker from 'react-datepicker'
+import moment from 'moment'
+
+import { IconCalendar } from './img/IconCalendar'
 
 interface Props {
   disabled?: boolean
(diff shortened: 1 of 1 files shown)


fvictorio:
Using `Asia/Kolkata` (Indian Standard Time):

image

There are timezones whose offsets are not multiple of 60. The correct label is `UTC+05:30`.

Plus, a lot of people don't use UTC (mostly americans). This is a less solvable problem, but I wonder if there's some library that handles this more automagically.

That being said, I don't think this is a blocker, it's probably good enough. @leolower, your call.

mariano-aguero:
> Using `Asia/Kolkata` (Indian Standard Time):
> 
> image
> 
> There are timezones whose offsets are not multiple of 60. The correct label is `UTC+05:30`.
> 
> Plus, a lot of people don't use UTC (mostly americans). This is a less solvable problem, but I wonder if there's some library that handles this more automagically.
> 
> That being said, I don't think this is a blocker, it's probably good enough. @leolower, your call.

I use moment, and look like this ...

Screenshot_20191220_235806

fvictorio:
Cool! I forgot we had `moment`. I just removed the space after `UTC` because I think that's how it's used. Feel free to merge this.

PR #292, opened 2019-12-20
Add logout button

mariano-aguero:
Look like this ...
Screenshot_20191220_120554


app/src/components/common/button_disconnect_wallet/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+  margin-top: 5.4px;
+  margin-left: 11px;
+`
+
+interface Props {
+  callback: () => void
+}
+
+export const ButtonDisconnectWallet = (props: Props) => {
+  const context = useWeb3Context()
```
fvictorio: You could use `useConnectedWeb3Context` here and avoid all the unnecessary checks below. The problem is that, I think, you won't have the `unsetConnector` callback. I wonder if it would make sense to expose the "raw" web3 context inside the connected web3 context (an extra `web3Context` property maybe?).

app/src/components/common/header/index.tsx
```diff
@@ -72,6 +77,11 @@ export const Header: React.FC = props => {
         <MobileMenuStyled toggleMenu={toggleMenu} isMenuOpen={isMenuOpen} />
         <ConnectedWeb3>
           <NetworkStyled />
+          <LogoutStyled
+            callback={() => {
+              setModalState(false)
```
fvictorio: Is this necessary?

app/src/components/common/button_disconnect_wallet/index.tsx
mariano-aguero: Yeahh, thats why I use this context, because I need to use the unsetConnector, mmm I will try to expose the context

app/src/components/common/header/index.tsx
mariano-aguero: Yes, because the state for the modal is true, and if you logout of the dapp then the modal is show really quickly in the screen . Is not so nice that you try to logout and the dapp display immediately the modal to login, any thoughts realted to this behaviour?

fvictorio: That didn't happen to me, and I don't see why it would. But it's OK, I don't mind having this.

mariano-aguero:
I also add a validation for the list items when the dapp is not connected

app/src/components/common/list_item/index.tsx
```diff
@@ -116,22 +129,43 @@ interface Props extends HTMLAttributes<HTMLDivElement> {
...
             <CalendarIconStyled />
             <Bold>Resolution Date:</Bold>
-            <ResolutionText>{data.resolution ? formatDate(data.resolution) : ''}</ResolutionText>
+            <ResolutionText>{resolution ? formatDate(resolution) : ''}</ResolutionText>
           </ResolutionDate>
           <Separator>-</Separator>
-          <Status resolved={data.status === MarketStatus.Resolved ? true : false}>
-            {data.status}
-          </Status>
+          <Status resolved={status === MarketStatus.Resolved ? true : false}>{status}</Status>
```
fvictorio:
```suggestion
          <Status resolved={status === MarketStatus.Resolved}>{status}</Status>
```

PR #296, opened 2019-12-23
Change fee to 4%

### LINKED ISSUE #285: Change fee to 4%
- Change fee in the market creation form
- Change fee in all calculations on other screens
- Filter out markets which don't have 4%

mariano-aguero:
Closes #285


app/src/common/constants.ts
```diff
@@ -22,3 +21,9 @@ export const FETCH_EVENTS_CHUNK_SIZE = parseInt(
 export const MAX_OUTCOME_ALLOWED = parseInt(process.env.REACT_APP_MAX_OUTCOME_ALLOWED || '4', 10)
 
 export const SINGLE_SELECT_TEMPLATE_ID = 2
+
+export const MARKET_FEE = parseInt(process.env.REACT_APP_MARKET_FEE || '4', 10)
+
+export const MARKET_FEE_WITH_TWO_DECIMALS = MARKET_FEE / Math.pow(10, 2)
```
fvictorio: PP: I would specify the fee with decimals in the environment variable and then use `parseFloat`.

app/src/components/market/market_sell.tsx
```diff
@@ -85,11 +86,12 @@ const MarketSellWrapper: React.FC<Props> = (props: Props) => {
     const holdingsOfOtherOutcomes = holdings.filter((item, index) => {
       return index !== outcomeIndex
     })
+
     const amountToSell = calcSellAmountInCollateral(
       amountShares,
       holdingsOfSoldOutcome,
       holdingsOfOtherOutcomes,
-      0.01,
+      +MARKET_FEE_WITH_TWO_DECIMALS,
```
fvictorio: This casting shouldn't be necessary.

app/src/components/market/market_sell.tsx
```diff
@@ -112,7 +114,7 @@ const MarketSellWrapper: React.FC<Props> = (props: Props) => {
     const pricesAfterTrade = MarketMakerService.getActualPrice(balanceAfterTrade)
 
     setPricesAfterTrade(pricesAfterTrade)
-    setCostFee(mulBN(amountToSell, 0.01))
+    setCostFee(mulBN(amountToSell, +MARKET_FEE_WITH_TWO_DECIMALS))
```
fvictorio: Same comment.

fvictorio:
Besides the comments, I think exporting the fee in three different forms (percentage, over 1 and in wei) is not a good idea. I think the constants should export just one canonical representation of the fee (probably `0.04`) and then the consumer should be the one that interprets it as it sees fit.

Plus, notice that with this approach a fee of `1.5%` cannot be specified.

app/src/common/constants.ts
mariano-aguero: Done!

app/src/components/market/market_sell.tsx
mariano-aguero: I remove the cast

app/src/components/market/market_sell.tsx
mariano-aguero: I remove the casting

mariano-aguero:
> Besides the comments, I think exporting the fee in three different forms (percentage, over 1 and in wei) is not a good idea. I think the constants should export just one canonical representation of the fee (probably `0.04`) and then the consumer should be the one that interprets it as it sees fit.
> 
> Plus, notice that with this approach a fee of `1.5%` cannot be specified.

I refactored the code, and allow to specify a fee with decimals, thanks for your feedback!

PR #302, opened 2019-12-23
Allow to connect to the wallet when the user press the button and create the market

### LINKED ISSUE #293: Allow to connect to the wallet when the user press the button and create the market

mariano-aguero:
Closes #293


app/src/components/market/market_wizard_creator_container.tsx
```diff
@@ -23,70 +27,81 @@ const MarketWizardCreatorContainer: FC = () => {
...
+
+        const collateralService = new ERC20Service(
+          provider,
+          rawWeb3Context.connectorName,
+          collateral.address,
+        )
+
+        const hasEnoughBalanceToFund = await collateralService.hasEnoughBalanceToFund(user, funding)
+        if (!hasEnoughBalanceToFund) {
+          throw new Error('there are not enough collateral balance for funding')
```
mariano-aguero: Mmmm maybe we need to inform this error message to the user, What about to use the `<Message />` component ? Or maybe you are ok with the actual behaviour ?

app/src/components/market/steps/items/funding_and_fee_step.tsx
```diff
@@ -52,22 +52,28 @@ const ErrorStyled = styled.span`
...
     },
-    [context, collateral],
+    [provider, rawWeb3Context, account, collateral],
   )
 
   const collateralBalance = useAsyncDerivedValue('', new BigNumber(0), calculateCollateralBalance)
 
-  const isFundingGreaterThanBalance = funding.gt(collateralBalance)
+  const isFundingGreaterThanBalance =
+    rawWeb3Context.connectorName !== 'Infura' ? funding.gt(collateralBalance) : false
```
mariano-aguero: We don't have the user address, so we don't have the `collateralBalance` of the user

fvictorio:
The problem with this is that you have to click the "Create" button twice: first for connecting the wallet and then again for really creating the market. It's probably something harder to do but maybe it's worth it to do it the right way.

I haven't looked at the code yet; if I can think of a simple way of doing this I'm going to suggest it in another comment.

mariano-aguero:
> The problem with this is that you have to click the "Create" button twice: first for connecting the wallet and then again for really creating the market. It's probably something harder to do but maybe it's worth it to do it the right way.
> 
> I haven't looked at the code yet; if I can think of a simple way of doing this I'm going to suggest it in another comment.

Mmm yeah , I did it this way so we are able to differentiate the action that the user was going to perform. I think  Logging into the application is not the same as creating a market. Maybe we could modify the text of the button suggesting to the user that he/she will perform both actions (an example 'Connect and create market'), remove the modal, and create a logic to connect in the button action. what do you think  @fvictorio ?

fvictorio:
I've thought about this a little more and I think the ideal solution would be to be able to do something like this:

```typescript
if (!context.account) {
  const submitted = await showModal()
  if (!submitted) {
    return
  }
}
```

The problem here is that `showModal` is not trivial to implement (see for example [this](https://github.com/cudr/react-modal-promise/blob/master/src/index.tsx)).

I don't think we need to do this right now, but the UX issue needs to be handled somehow. A compromise solution I think could work is to show `Connect Wallet` as the label of the button the first time and then, when the account is available, do show `Create`.

app/src/components/market/market_wizard_creator_container.tsx
mariano-aguero: Done!

app/src/pages/create_market_page.tsx
```diff
@@ -1,16 +1,37 @@
 import React from 'react'
 
 import { MarketWizardCreatorContainer } from '../components'
+import connectors from '../util/connectors'
 import { ConnectedWeb3, useConnectWeb3 } from '../hooks/connectedWeb3'
+import { useWeb3Context } from 'web3-react/dist'
```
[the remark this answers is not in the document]
mariano-aguero: Done!

app/src/pages/create_market_page.tsx
```diff
@@ -1,16 +1,37 @@
...
 
-const CreateMarketPage = () => {
+const CreateMarketConnectedPage = (props: any) => {
```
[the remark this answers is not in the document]
mariano-aguero: Done!

app/src/services/erc20.ts
```diff
@@ -19,10 +19,19 @@ const erc20Abi = [
 class ERC20Service {
   tokenAddress: string
   provider: any
+  contract: Contract
+  connectorName: string
 
-  constructor(provider: any, tokenAddress: string) {
+  constructor(provider: any, connectorName: string, tokenAddress: string) {
```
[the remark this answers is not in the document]
mariano-aguero: Done

app/src/services/erc20.ts
```diff
@@ -61,27 +62,23 @@ class ERC20Service {
...
       value: '0x0',
     })
     logger.log(`Approve unlimited transaccion hash: ${transactionObject.hash}`)
     await this.provider.waitForTransaction(transactionObject.hash)
   }
 
   getCollateral = async (marketMakerAddress: string): Promise<any> => {
-    const signer: Wallet = this.provider.getSigner()
+    return this.connectorName === 'Infura'
+      ? new BigNumber(0)
```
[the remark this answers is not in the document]
mariano-aguero: Removed

app/src/services/erc20.ts
```diff
@@ -61,27 +62,23 @@ class ERC20Service {
...
+      : this.contract.balanceOf(marketMakerAddress)
+  }
 
-    const erc20Contract = new ethers.Contract(this.tokenAddress, erc20Abi, this.provider).connect(
-      signer,
-    )
+  hasEnoughBalanceToFund = async (owner: string, amount: BigNumber): Promise<boolean> => {
```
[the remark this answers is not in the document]
mariano-aguero: > I would drop this method. First, it couples the service to a particular use case (funding). Second, it's something that is really easy for the callee to do.

I will do this on the component that call the method

mariano-aguero:
> I've thought about this a little more and I think the ideal solution would be to be able to do something like this:
> 
> ```ts
> if (!context.account) {
>   const submitted = await showModal()
>   if (!submitted) {
>     return
>   }
> }
> ```
> 
> The problem here is that `showModal` is not trivial to implement (see for example [this](https://github.com/cudr/react-modal-promise/blob/master/src/index.tsx)).
> 
> I don't think we need to do this right now, but the UX issue needs to be handled somehow. A compromise solution I think could work is to show `Connect Wallet` as the label of the button the first time and then, when the account is available, do show `Create`.

I update the text of the button , please let me know if you need something else or I miss something

app/src/util/market_creation_status_data.ts
```diff
@@ -0,0 +1,105 @@
...
+}
+interface Done {
+  readonly _type: Types.Done
+}
+interface Error {
+  readonly _type: Types.Error
+  readonly error: Error
+}
+
+export type MarketCreationStatusType =
```
fvictorio: Maybe we can use the same name for the type and the helper object, as we do with `RemoteData`. That way you only need to import one name from this module.

fvictorio: Also drop the `Data` suffix.

fvictorio:
This has a problem:

1. Open the app with metamask connected to ganache (but logged out)
2. Create a market an get to the final step
3. Click "Connect wallet" and then "Create"
4. The app tries to use the Rinkeby's addresses instead of the ones of the connected network (ganache)

app/src/util/market_creation_status_data.ts
mariano-aguero: :+1:

mariano-aguero:
> This has a problem:
> 
> 1. Open the app with metamask connected to ganache (but logged out)
> 2. Create a market an get to the final step
> 3. Click "Connect wallet" and then "Create"
> 4. The app tries to use the Rinkeby's addresses instead of the ones of the connected network (ganache)

Fixed

Issue #222, opened 2019-11-19
CATEGORICAL MARKETS - Fund existing market - List all options


leolower:
**Design:** This section needs an update. It's implemented but there's no design, so add a new one. We need to make it look more like the others.

**Some thoughts:**

- Maybe "Amount" should be first? It's the most important thing on this section. Maybe the buttons should be near it? Not sure about this.

- If we don't put "Amount" first, we should follow the order we followed on the other sections: "Outcomes" first, then "Details" (we need to make "Details" look like on the other designs BTW...), and "Amount" last.

- Maybe the wording and order for the buttons should be "Remove Funds" / "Fund".

![Screen Shot 2019-12-11 at 09 52 51](https://user-images.githubusercontent.com/4015436/70623401-cdff7180-1bfc-11ea-9fb8-7818c9b4c113.png)

alongoni:
- I made changes based on structure of Buy/Sell screen. 
- Change button text and order.

![09 - Conditional Exchange - Fund](https://user-images.githubusercontent.com/622217/71097778-55199000-218f-11ea-8230-a64ad878a5f7.png)

gabitoesmiapodo:
- I've been using the "Full" buttons version for "final" actions (for example "Create", "Finish", "Save", etc.), and the "Line" version for intermediate actions ("Next", etc.).

I think we should use the full version for "Remove Funding" and "Fund" to keep consistency, and use the orange version for "Remove Funding" to differentiate the buttons, and also to indicate certain "danger" in performing this action.

- You should add the scroll for multiple options at the top.

- The four items at the beginning of the original screenshot are fixed ("Total funding", etc.), so they all should be present in your design.

alongoni:
![09 - Conditional Exchange - Fund](https://user-images.githubusercontent.com/622217/71176402-91f58d80-2248-11ea-83fa-82482cab2b19.png)

gabitoesmiapodo:
Nice.

PR #303, opened 2019-12-23
(Feature) ability to create multiple options on market creation

### LINKED ISSUE #212: Ability to create multiple options on Market creation
**Design:** We need to update this section's design to provide a better way of adding multiple options / outcomes.

There's a preliminary design, but I think it could be improved.

Maybe there's a better place for the `+` button?

Also add a scroll for the items' list.

**Note:** This issue includes design for #213 

03 - Conditional Exchange - Create 3

### LINKED ISSUE #213: Ability to set the odds for each option
**Design:** We need  for this step to support several new things.

The possibility to add more than 2 outcomes / options should be done here: https://github.com/protofire/gnosis-conditional-exchange/issues/212

But we also need:

- A way to delete each entry.
- By default, adding a new entry resets all the values and makes them equal (50% for each one if there are 2 entries, 33.33% if there are 3 entries, 25% if there are 4, and so on...) Think about how or _if_ we should prevent the user resetting his / her previously entered values every time he / her adds a new entry. The same would happen when an entry is deleted.
- If the sum of all of the odds is less (or more) than 100%, we should clearly inform the user that the total amount is not 100%, and he / she will not be allowed to proceed. **Total must be 100%**. I'll leave this one up to you, I have no suggestions...

03 - Conditional Exchange - Create 3

gabitoesmiapodo:
Depends on #222 

Closes #212 
Closes #213

Added all of the missing elements.

Scroll appears only if there are more than 4 items, so no scroll for now (unless you force it somehow).

There's still some missing logic, so I'm assigning this to @mariano-aguero :

- Total outcomes probabilities are not calculated yet. It should be red if there's an error.
- There's a field at the bottom to add a new outcome, it should add the outcome with the completed name.
- Item remove buttons should work.

Errors don't look nice, but they are OK for now.

**Screenshots:**

Screen Shot 2019-12-23 at 17 17 59


mariano-aguero:
I solved these three items:
- Total outcomes probabilities are not calculated yet. It should be red if there's an error.
- There's a field at the bottom to add a new outcome, it should add the outcome with the completed name.
- Item remove buttons should work.

fvictorio:
Some comments:

- Can we disable the "Add outcome" button if the name of the outcome is empty?
- The total percentage becomes red when it's greater than 100, maybe it should also do it when it's less than 100?

app/src/components/market/steps/items/outcomes_step.tsx
```diff
@@ -60,18 +57,23 @@ const OutcomesStep = (props: Props) => {
...
-  const totalOutcomeProbabilities = outcomes.reduce((total, cur) => total + cur.probability, 0)
-  if (totalOutcomeProbabilities !== 100) {
+  const totalProbabilities = outcomes.reduce((total, cur) => total + cur.probability, 0)
+  if (totalProbabilities !== 100) {
     errorMessages.push('The sum of all probabilities must be equal to 100%.')
   }
 
-  const error = totalOutcomeProbabilities !== 100 || someEmptyName || someEmptyProbability
+  if (outcomes.length < 2) {
+    errorMessages.push('The quantity of outcomes should be equal or greater than 2.')
```
fvictorio:
```suggestion
    errorMessages.push('Please enter at least 2 outcomes')
```

mariano-aguero:
> Some comments:
> 
> * Can we disable the "Add outcome" button if the name of the outcome is empty?
> * The total percentage becomes red when it's greater than 100, maybe it should also do it when it's less than 100?

Both items were done

PR #319, opened 2020-01-02
Selling all shares doesn't seem to work 

### LINKED ISSUE #305: Selling all shares doesn't seem to work
Reproduction steps:

1. Create a binary market with initial odds of 75/25
2. Go to the sell page
3. Select the outcome whose shares you own
4. Click "Add all funds" and then "Sell"

This fails for me.

mariano-aguero:
Closes #305 

@fvictorio 
Increasing the gas limit seems to work, but I don't like it, suggestions are welcome
Some transactions that worked:
https://rinkeby.etherscan.io/tx/0xe89549e0d225e86269c66b5a1ad7f376212313a1e6383462c8a2a73c0f008b63
https://rinkeby.etherscan.io/tx/0x8866a920bb2005fba44768f424799f6c849e14b497a568edbfdb9c6db1d49b83

### DIFF
--- a/app/src/components/market/market_sell.tsx
+++ b/app/src/components/market/market_sell.tsx
@@ -100,7 +100,7 @@ const MarketSellWrapper: React.FC<Props> = (props: Props) => {
       })
 
       const amountToSell = calcSellAmountInCollateral(
-        amountShares.mul(99999).div(100000),
+        amountShares.mul(99999).div(100000), // Because exist some precision error, we need to multiply the amount by 0.99999 and divide by 1
         holdingsOfSoldOutcome,
         holdingsOfOtherOutcomes,
         marketFeeWithTwoDecimals,
--- a/app/src/components/market/market_sell.tsx
+++ b/app/src/components/market/market_sell.tsx
@@ -100,7 +100,7 @@ const MarketSellWrapper: React.FC<Props> = (props: Props) => {
       })
 
       const amountToSell = calcSellAmountInCollateral(
-        amountShares.mul(99999).div(100000), // Because exist some precision error, we need to multiply the amount by 0.99999 and divide by 1
+        amountShares.mul(99999).div(100000), // because of some precision error, we need to multiply the amount by 0.99999
         holdingsOfSoldOutcome,
         holdingsOfOtherOutcomes,
         marketFeeWithTwoDecimals,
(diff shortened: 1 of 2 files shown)


fvictorio:
I still get this error, at least in ganache.

A fix that seems to work is to do this in `market_sell.tsx`:

```typescript
      const amountToSell = calcSellAmountInCollateral(
        amountShares.mul(99999).div(100000),
        holdingsOfSoldOutcome,
        holdingsOfOtherOutcomes,
        marketFeeWithTwoDecimals,
      )
```

Meaning: multiply the amount in the input by 0.99999. It's hacky, but since the problem probably comes from the root finding giving a solution with some precision error, I think this is the best we can do. Could you try this (adding a proper comment) and test it a little more?

Btw, this is only needed in the sell page.

mariano-aguero:
> I still get this error, at least in ganache.
> 
> A fix that seems to work is to do this in `market_sell.tsx`:
> 
> ```ts
>       const amountToSell = calcSellAmountInCollateral(
>         amountShares.mul(99999).div(100000),
>         holdingsOfSoldOutcome,
>         holdingsOfOtherOutcomes,
>         marketFeeWithTwoDecimals,
>       )
> ```
> 
> Meaning: multiply the amount in the input by 0.99999. It's hacky, but since the problem probably comes from the root finding giving a solution with some precision error, I think this is the best we can do. Could you try this (adding a proper comment) and test it a little more?
> 
> Btw, this is only needed in the sell page.

I added this solution to the PR

fvictorio:
Please add an explanatory comment.

app/src/components/market/market_sell.tsx
```diff
@@ -100,7 +100,7 @@ const MarketSellWrapper: React.FC<Props> = (props: Props) => {
       })
 
       const amountToSell = calcSellAmountInCollateral(
-        amountShares,
+        amountShares.mul(99999).div(100000), // Because exist some precision error, we need to multiply the amount by 0.99999 and divide by 1
```
fvictorio:
```suggestion
        amountShares.mul(99999).div(100000), // because of some precision error, we need to multiply the amount by 0.99999
```

mariano-aguero:
@fvictorio I update the base branch to master

PR #320, opened 2020-01-02
Feature/#291 add option to logout from the current wallet

### LINKED ISSUE #291: Add option to logout from the current wallet
We need a button or link to let the user logout from the current wallet.

I'll leave the details and workflow up to you.

alongoni:
Closes #291
 
17 - Conditional Exchange - Disconect wallet

### DIFF
--- a/app/src/components/common/button_disconnect_wallet/index.tsx
+++ b/app/src/components/common/button_disconnect_wallet/index.tsx
@@ -1,47 +1,15 @@
 import React from 'react'
 import styled, { css } from 'styled-components'
+import { Button } from '../button'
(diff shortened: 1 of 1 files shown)


app/src/components/common/button_disconnect_wallet/index.tsx
```diff
@@ -1,47 +1,14 @@
 import React from 'react'
 import styled, { css } from 'styled-components'
```
gabitoesmiapodo:
```suggestion
import styled from 'styled-components'
```
You should delete `{ css }` if it's not used.

PR #339, opened 2020-01-07
Feedback demo milestone 5

mariano-aguero:
#### Includes
- Market detail page: remove the color for the ‘winning so far’ outcome. It is confusing. So far leave everything look the same.
- Create New Market, 2nd step. Change ‘Add Collateral’ label to ‘Add token’
- Sell page: you can add an Amount value higher than your Balance. It should allow to type such number or provide you with an error.

Only  included these items


app/src/components/common/bar_diagram_probabilities/index.tsx
```diff
@@ -77,11 +77,20 @@ interface Props extends HTMLAttributes<HTMLButtonElement> {
   theme?: any
   winningOutcome: boolean
   withWinningOutcome: boolean
+  displayWinningColor?: boolean
```
fvictorio: I would drop this prop, and just show the color when `withWinningOutcome` is true.

app/src/components/common/customizable_tokens_select/index.tsx
```diff
@@ -36,7 +36,7 @@ export const CustomizableTokensSelect = (props: Props) => {
         onTokenChange={onCollateralChange}
         value={value}
       />
-      <Link onClick={() => setModalCollateralState(true)}>Add collateral</Link>
+      <Link onClick={() => setModalCollateralState(true)}>Add token</Link>
```
fvictorio:
```suggestion
      <Link onClick={() => setModalCollateralState(true)}>Add custom token</Link>
```

app/src/components/common/customizable_tokens_select/modal_collateral.tsx
```diff
@@ -75,7 +75,7 @@ export const ModalCollateral = (props: Props) => {
   }
 
   return (
-    <ModalWrapper isOpen={isOpen} onRequestClose={onClose} title={`Add Collateral`}>
+    <ModalWrapper isOpen={isOpen} onRequestClose={onClose} title={`Add Token`}>
```
fvictorio:
```suggestion
    <ModalWrapper isOpen={isOpen} onRequestClose={onClose} title={`Add custom token`}>
```

app/src/components/market/market_buy.tsx
```diff
@@ -152,7 +153,29 @@ const MarketBuyWrapper: React.FC<Props> = (props: Props) => {
     }
   }
 
-  const disabled = (status !== Status.Ready && status !== Status.Error) || cost.isZero()
+  const calculateCollateralBalance = useMemo(
```
fvictorio: This seems to work, but the problem is that it kind of duplicates the logic inside `BalanceToken`. Maybe `BalanceToken` should receive the user's balance instead of computing it? (I like this idea because it would make it more pure).

If we do that, then we should also do it in the sell and fund pages. But that makes sense: this check should be there too.

fvictorio: Just to be clearer: if we make this, `BalanceToken` should receive everything it needs. It shouldn't have any `useEffect` or `useAsyncDerivedValue`.

And btw I'm not sure if we need `useAsyncDerivedValue` for this, since it delays the computation unnecessarily. (I blame me for not indicating this in the name of the hook, but `useAsyncAndDelayedDerivedValue` doesn't sound that good).

app/src/components/common/bar_diagram_probabilities/index.tsx
mariano-aguero: Done!

app/src/components/market/market_buy.tsx
mariano-aguero: I create a hook to get the balance of some collateral. `BalanceToken` component is more pure right now, only receive balance and collateral information and doesn't have any `useEffect` or `useAsyncDerivedValue`.

PR #341, opened 2020-01-09
Use explicit arguments for address truncation

### LINKED ISSUE #335: Change the way the connected address is shown
It should be `0x<4 chars>...<4chars>`

hpmaxi:
Closes #335


app/src/components/common/network/index.tsx
```diff
@@ -68,7 +68,7 @@ export const Network: React.FC = props => {
       <ConnectionStatusWrapper>
         <ConnectionStatusDot />
         <ConnectionStatusText>
-          {truncateStringInTheMiddle(account) || 'No account connected'}
+          {truncateStringInTheMiddle(account, 41, 6, 4) || 'No account connected'}
```
fvictorio: I don't remember the reason for the second parameter. I think just indicating the length of the first part and the second part should be enough: then if the length of the string is less than or equal to the sum of both arguments, you just return it.

For example:

```javascript
truncate("foobarbaz", 3, 2) // "foo...az"
truncate("foobar", 4, 2) // "foobar"
```

@mariano-aguero do you agree?

@hpmaxi can you do this (and also add tests for it)?

mariano-aguero: Yes, I agree, we don't need the second parameter

hpmaxi: I hope there's enough tests

app/src/util/tools.spec.ts
```diff
@@ -177,4 +178,18 @@ describe('tools', () => {
       ).toThrow()
     })
   })
+
+  describe('truncateStringInTheMiddle', () => {
+    it('should not change string with truncate positions great or equal than string length', () => {
+      expect(truncate('foobar', 4, 2)).toBe('foobar')
+      expect(truncate('foobar', 6, 6)).toBe('foobar')
+      expect(truncate('foo')).toBe('foo')
```
fvictorio: I don't know if it makes sense to have default arguments for this method.

mariano-aguero: > I don't know if it makes sense to have default arguments for this method.

Maybe, the last two arguments must be required ?

hpmaxi: I've removed the last two parameters default values to make them explicit on each call of the function

PR #350, opened 2020-01-09
Fill new market info with existing Realitio question

### LINKED ISSUE #342: Fill new market info with existing Realitio question

### LINKED ISSUE #343: Disable inputs that were filled with an existing question

mariano-aguero:
Closes #342 
Closes #343 

Look like this ...
Screenshot_20200109_180443
Screenshot_20200109_180509
Screenshot_20200109_180529

I apply a solution very similar to the one we have with the custom token. We fetch all the information using logs. I had to refactor a lot of existing code, such as the handling of the arbitrator and the categories among other things. Now we have an `Unknow` arbitrator in case the address doesn't exist in the setup of the application.
I update the `arbitratorId` by an arbitrator object, added a lot of functionality in the networks file, also a spinner in the token modal.
All the information related to the question from realitio are blocked, and we allow the user to clear this action in case he/she needs it


app/src/components/common/outcomes/index.tsx
```diff
@@ -109,10 +109,11 @@ interface Props {
   onChange: (newOutcomes: Outcome[]) => any
   errorMessages?: string[]
   totalProbabilities: number
+  disabled: boolean
 }
 
 const Outcomes = (props: Props) => {
-  const { outcomes, totalProbabilities, errorMessages } = props
+  const { outcomes, totalProbabilities, disabled = false, errorMessages } = props
```
fvictorio: The prop is mandatory so the default value doesn't seem necessary.

app/src/components/market/market_wizard_creator.tsx
```diff
@@ -70,6 +56,7 @@ export const MarketWizardCreator = (props: Props) => {
         probability: 50,
       },
     ],
+    questionIsFromRealitio: false,
```
fvictorio: Maybe `existingQuestion` or `loadedQuestion` to make it less coupled to Realitio.

app/src/components/market/steps/items/funding_and_fee_step.tsx
```diff
@@ -50,7 +50,7 @@ const FundingAndFeeStep = (props: Props) => {
   const context = useConnectedWeb3Context()
   const { library: provider, account } = context
 
-  const { values, addCollateralCustom, handleChange, handleCollateralChange } = props
+  const { values, addCollateralCustom, handleChange, handleChangeCollateral } = props
```
fvictorio: Why this change? `handleCollateralChange` sounds much better than `handleChangeCollateral` (and it makes more grammatical sense IMO).

app/src/global.d.ts
```diff
@@ -8,4 +8,4 @@ declare type Maybe<T> = T | null
 
 declare type KnownToken = 'cdai' | 'usdc' | 'dai' | 'weth' | 'owl' | 'chai'
 
-declare type KnownArbitrator = 'kleros' | 'realitio'
+declare type KnownArbitrator = 'kleros' | 'realitio' | 'unknow'
```
fvictorio: `unknown`

app/src/components/common/question_input/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+        disabled={disabled}
+        type="text"
+      />
+      <Wrapper>
+        <Note>
+          <strong>For example:</strong> <i>&quot;Will France win?&quot;</i> is not an acceptable
+          question, but <i>&quot;Will France win the 2020 FIFA World Cup?&quot;</i> is a good one.
+        </Note>
+        {!disabled && <Link onClick={() => setModalQuestionState(true)}>Add question</Link>}
+        {disabled && <Link onClick={onClearQuestionFromRealitio}>Clear question</Link>}
```
fvictorio: If I clear the question, the title of the question is kept in the input (but editable). I'm not sure about that. And, anyway, even if that's OK, if you then add the category, resolution date and arbitrator, the "Next" button is still disabled (if you then edit the question title, it gets enabled).

app/src/components/common/outcomes/index.tsx
mariano-aguero: Done!

app/src/components/market/market_wizard_creator.tsx
mariano-aguero: Done!

app/src/global.d.ts
mariano-aguero: Fixed

app/src/components/common/question_input/index.tsx
mariano-aguero: Done!

mariano-aguero:
> This posts the question in Realitio. It should just reuse the questionId without opening a duplicated question.
> 
> I didn't review the code too much yet, will do after this is fixed.

Ohh sorry for that, fix it!. Now I allow to reuse the loaded question in the first step.

app/src/components/market/market_wizard_creator.tsx
```diff
@@ -56,7 +56,7 @@ export const MarketWizardCreator = (props: Props) => {
         probability: 50,
       },
     ],
-    loadedQuestion: false,
+    loadedQuestion: '',
```
fvictorio: I don't think it's a good idea to use an empty string to signal a missing value. If the goal is to use this variable to indicate that the question is loaded *and* to have the question id, it should probably be renamed and retyped to `loadedQuestionId: Maybe<string>`.

app/src/components/market/market_wizard_creator_container.tsx
```diff
@@ -37,20 +37,31 @@ const MarketWizardCreatorContainer: FC = () => {
...
         setMarketCreationStatus(MarketCreationStatus.postingQuestion())
-        const questionId = await realitio.askQuestion(
-          question,
-          outcomes,
-          category,
-          arbitrator.address,
-          openingDateMoment,
-          networkId,
-        )
+        const questionId = !loadedQuestion
```
fvictorio: I don't mind this a lot, but _maybe_ it would be better to be more explicit:

```
let questionId: string
if (loadedQuestionId) {
  questionId = loadedQuestionId
} else {
  setMarketCreationStatus(MarketCreationStatus.postingQuestion())
  questionId = await realitio.askQuestion(...)
}
setQuestionId(questionId)
```

(or something like that)

app/src/components/market/market_wizard_creator.tsx
mariano-aguero: Done

app/src/components/market/market_wizard_creator_container.tsx
mariano-aguero: Done

PR #352, opened 2020-01-13
Update how 3box is initialized

mariano-aguero:
No issue related. Now we use a handler, the Option B in the documentacion.

See docs for more info:
 https://docs.3box.io/build/plugins/comments


mariano-aguero:
Added improvement on the footer, look like this ... 

Screenshot_20200114_125052

Before the change...
Screenshot_20200114_122350

PR #353, opened 2020-01-13
Improve information shown in buy page

### LINKED ISSUE #304: Improve information shown in buy page
Right now the Buy page looks like this:

image

But the "You spend" part is wrong: if the user enters 1.00 DAI, then they'll spend 1.00 DAI, not 1.04. 0.96 of those DAIs will be exchanged by outcome tokens and 0.04 will be kept as fee.

I think just fixing this by showing 1.00 won't work, because it won't be adding relevant information. Changing it for 0.96 might be better but is perhaps misleading: you won't spend 0.96; you'll spend 1.00.

mariano-aguero:
Closes #304 

Look like this ...
Screenshot_20200113_151300


mariano-aguero:
> The "You will receive" row shouldn't have a bottom border.
> 
> It seems like the **DAI** and **Shares** part on `x.xx DAI` and `x.xx Shares` from the totals should be a lighter color (not sure, but maybe the numbers too).
> 
> `DAI` shouldn't be bold.
> 
> `Shares` should be capitalized.

Good, annndd all done. Look like this ...
Screenshot_20200114_162825

You can check all the changes in this market `http://localhost:3000/#/0xd718e4a04011Ee5f070178dC582cea5F10d6c372/buy` , using the rinkeby network

PR #356, opened 2020-01-14
Update View Market Info block

### LINKED ISSUE #300: Update View Market Info block
It should look like this, and have the _"Copy URL"_ functionality https://zpl.io/VDdgyN3

alongoni:
Closes #300 
**Missing add "Copy URL" functionality**
CE-issue-300


gabitoesmiapodo:
Is this PR for https://github.com/protofire/gnosis-conditional-exchange/issues/300 ?

Don't forget to add the `Closes #300` at the beginning so we don't lose track.

gabitoesmiapodo:
@alongoni 

Nice. It still needs the copy functionality, maybe @mariano-aguero can implement it?

I'll do it later if you can't.

mariano-aguero:
> @alongoni
> 
> Nice. It still needs the copy functionality, maybe @mariano-aguero can implement it?
> 
> I'll do it later if you can't.

Mmmm I think is not necessary , because the component `CopyText` already do that. Also is implemented in the PR. For me is OK...

gabitoesmiapodo:
@mariano-aguero I wasn't aware.

Does it show a toast or something similar when the URL is copied? It would be good to have it if it doesn't already.

mariano-aguero:
> @mariano-aguero I wasn't aware.
> 
> Does it show a toast or something similar when the URL is copied? It would be good to have it if it doesn't already.

Only copy a given text (in this case, the URL) , maybe it would be good to display a toast or a message, right now we have a component to display this kind of messages ... https://app.zeplin.io/project/5b71fb9cd86c791b1612c035/screen/5e0e4b3ad754381882deb642
Maybe we can use any of them...

gabitoesmiapodo:
@mariano-aguero 

I took the liberty of updating the toasts / alerts design (cheating, I know) https://zpl.io/a7xnBGq

Updated the implementation, and added some extra functionality.

Also updated the functionality of the copy button, which shows said alert when used, as shown below.

Jan-23-2020 15-06-56

PR #360, opened 2020-01-16
Use CPK when buying outcomes

### LINKED ISSUE #348: Use CPK when buying outcomes

mariano-aguero:
Closes #348 

#### Includes
- Add CPK library to batch transactions
- Added a service to use with the CPK
- Encode some functions to use with the batch transactions
- Apply cpk when buy outcomes


mariano-aguero:
A successful transaction: https://rinkeby.etherscan.io/tx/0xbeca21f508e321dd2328cec857a0e7c6de842cc4522ccf2c78ced3da5bd47329

app/src/components/market/market_buy.tsx
```diff
@@ -152,27 +153,19 @@ const MarketBuyWrapper: React.FC<Props> = (props: Props) => {
       setStatus(Status.Loading)
       setMessage(`Buying ${formatBigNumber(tradedShares, collateral.decimals)} shares ...`)
 
-      const user = await provider.getSigner().getAddress()
+      // add 10% to the approved amount because there can be precision errors
+      // this can be improved if, instead of adding the 1% fee manually in the front, we use the `calcMarketFee`
+      // contract method and add it to the result of `calcNetCost` result
+      const costWithErrorMargin = cost.mul(11000).div(10000)
```
fvictorio: Since this is sending an extra 10% of collateral to the CPK, but not all of it ends up being used, does this mean that the CPK will keep some tokens?

Did you test if the proxy's collateral balance after buying was 0?

app/src/services/cpk.ts
```diff
@@ -0,0 +1,124 @@
...
+      const txObject = await cpk.execTransactions(
+        [
+          // Step 1: Transfer an amount (cost) from the user to the CPK
+          {
+            operation: CPK.CALL,
+            to: collateralAddress,
+            value: 0,
+            data: ERC20Service.encodeTransferFrom(account, cpkAddress, cost),
+          },
+          // Step 2: Approve an amount(cost) to be used by the market maker (later that amount(cost) will be transferred to the market maker)
```
fvictorio: This comment is wrong, the approval is unlimited.

fvictorio: Very low priority but: the second time a purchase is done, the approval isn't necessary and therefore wasteful. This step should only be added if the allowance of the CPK to the market maker is not enough.

app/src/services/cpk.ts
```diff
@@ -0,0 +1,124 @@
...
+
+      const collateralAddress = await marketMaker.getCollateralToken()
+      const marketMakerAddress = marketMaker.address
+      const conditionalTokensAddress = conditionalTokens.address
+
+      const cpk = await CPK.create({ ethers, signer })
+      const cpkAddress = cpk.address
+
+      // Approve amount of collateral to the CPK
+      const collateralService = new ERC20Service(provider, account, collateralAddress)
```
fvictorio: I think this shouldn't be done by the CPK service. The callee should do it. Plus, doing it that way will remove a parameter from this function.

app/src/services/erc20.ts
```diff
@@ -108,6 +107,30 @@ class ERC20Service {
       symbol,
     }
   }
+
+  static encodeTransferFrom = (from: string, to: string, amount: BigNumber): any => {
```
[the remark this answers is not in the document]
mariano-aguero: Done!

app/src/services/conditional_token.ts
```diff
@@ -125,6 +125,33 @@ class ConditionalTokenService {
 
     await this.provider.waitForTransaction(transactionObject.hash)
   }
+
+  static encodeSafeTransferFrom = (
+    addressFrom: string,
+    addressTo: string,
+    positionId: BigNumber,
+    outcomeTokensToTransfer: BigNumber,
+  ): any => {
```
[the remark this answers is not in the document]

app/src/services/conditional_token.ts
```diff
@@ -125,6 +125,33 @@ class ConditionalTokenService {
...
+    const safeTransferFromInterface = new utils.Interface([
+      'function safeTransferFrom(address from, address to, uint256 id, uint256 value, bytes data) external',
+    ])
```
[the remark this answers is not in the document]

app/src/services/market_maker.ts
```diff
@@ -48,6 +49,7 @@ class MarketMakerService {
     this.conditionalTokens = conditionalTokens
     this.realitio = realitio
     this.provider = provider
+    this.address = address
```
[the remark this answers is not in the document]
mariano-aguero: Done!

app/src/services/conditional_token.ts
```diff
@@ -33,11 +32,14 @@ class ConditionalTokenService {
     } else {
       this.contract = new ethers.Contract(address, conditionalTokensAbi, provider)
     }
-    this.address = address
     this.signerAddress = signerAddress
     this.provider = provider
   }
 
+  getAddress = (): string => {
```
fvictorio: It wasn't necessary to change the _usages_ of this. If you use a getter like in the comment (repeated here):

```typescript
get address(): string => { ... }
```

then the usage continues to be `marketMaker.address` (without parentheses).

That being said, if you prefer explicit getters, that's OK with me.

app/src/services/cpk.ts
mariano-aguero: Done!

app/src/services/cpk.ts
mariano-aguero: Mmmm I'm not sure about this, we need the `cpkAddress` to make the approve, and this exist only in the CPK Service, maybe we need a constructor and a property for the address todo this ? what do you think @fvictorio ?

app/src/components/market/market_buy.tsx
mariano-aguero: Yes, the CPK will keep some tokens, you can check this situation comparing this transactions:
1 tx: Balance of the cpk: 27.398
https://rinkeby.etherscan.io/address/0x48821d9891da0cebd713a163771f8e1f3154f606
Screenshot_20200121_135634

2 tx: Balance of the cpk: 27.631
https://rinkeby.etherscan.io/address/0x48821d9891da0cebd713a163771f8e1f3154f606
Screenshot_20200121_135644

I think the best solution is to remove the  extra 10% amount of collateral  we send to the cpk.
If we remove this extra amount, the balance of the cpk will be 27.631
tx: https://rinkeby.etherscan.io/address/0x48821d9891da0cebd713a163771f8e1f3154f606
Screenshot_20200121_140414

mariano-aguero: I remove the amount with cost, are you ok with this ?

app/src/services/conditional_token.ts
mariano-aguero: Mmmm ok , I apply getter syntax.

mariano-aguero:
@fvictorio, I remove the last two transactions, so right now we will keep the shares in the CPK

app/src/services/cpk.ts
fvictorio: We are going to need to have access that address everywhere so we may as well solve this problem now.

mariano-aguero: @fvictorio I move the allowance logic outside the CPK Service

app/src/services/cpk.ts
```diff
@@ -17,6 +17,13 @@ interface CPKBuyOutcomesParams {
 }
 
 class CPKService {
+  static getCPK = async (provider: any) => {
```
fvictorio: Wouldn't it better to make a constructor, and then make `buyOutcomes` an instance method?

mariano-aguero: > Wouldn't it better to make a constructor, and then make `buyOutcomes` an instance method?

@fvictorio Mmmm I think we can't implement an async constructor(is not allowed, please correct me if I'm wrong), maybe and initialize async method ?

fvictorio: Yep, you are right. A static async factory method is probably the best option:

```
constructor(cpk) {
  this.cpk = cpk
}

static async create() {
  const cpk = await ...
  return new CPKService(cpk)
}

buyOutcomes = async (...) { // instance method that uses this.cpk
```

mariano-aguero: Done!

PR #372, opened 2020-01-20
Show buttons disconnected

### LINKED ISSUE #365: Display the buttons even if the wallet is not connected
Trigger the login modal if anyone clicks

hpmaxi:
Closes #365 

Sell button isn't shown when the user is disconnected because there is a condition for userHasShares being true, also this button is not updating his visibility on connection/disconnection. I don't know if these things are a real problem or not.

The effect used on the closed market is a little more involved to avoid a bug that make the modal appears after clicking disconnect button.


fvictorio:
I never liked the whole "Show buy/sell/fund buttons only if the user has collateral/shares", and I'm pretty sure it wasn't requested. Maybe this is the right place to remove that logic?

cc @leolower @mariano-aguero @gabitoesmiapodo

app/src/components/market/profile/view.tsx
```diff
@@ -30,17 +31,20 @@ interface Props extends RouteComponentProps<{}> {
   status: Status
   marketMakerAddress: string
   funding: BigNumber
+  location: any
```
fvictorio: Why is this being added?

gabitoesmiapodo:
@fvictorio What would the user sell if they have no collateral / shares? Buttons should at least be disabled if you can't perform the action, I think.

But if this is related to being connected / disconnected, I think the user should be able to do everything they want to do and ask them to connect when further progress is not possible anymore without connecting.

A possible workflow could be: User is disconnected; there's a visible "Sell" button; user presses the "Sell" button; they are asked to connect; depending on the amount of shares, the button is enabled or disabled;

I'm not positive if this is 100% correct, just a thought. We may need to define possible workflows and what to do in every scenario.

Not sure if this is a simple thing to implement and / or fix, though.

app/src/components/market/profile/view.tsx
hpmaxi: Removed!

fvictorio:
Ok, so here's the thing.

### What we have now

If the wallet is not connected, no buttons are shown. If the wallet is connected, then buttons are shown only if the user has the proper tokens (collateral for Buy and Fund, outcome tokens for Sell). This is kind of OK I guess? The problem is that if you don't have collateral nor outcomes, then you don't really know what can be done in the app. Showing them but disabling them could work. (Maybe with a hover text explaining why they are disabled? Although I think you [can't add a tooltip on a disabled text](https://stackoverflow.com/questions/2151575/how-do-you-render-tooltip-on-disabled-html-button).)

### What this PR does

With this PR, buttons are always shown if the wallet isn't connected. If you click the "Connect wallet" button in the header, and you don't have tokens, then some (or all) of the buttons disappear. This is weird enough.

If you have collateral but haven't connected your wallet yet, and then you click "Buy", the "Connect Wallet" modal appears and, after connecting, you are redirected to the Buy page. What happens if you don't have collateral? I don't know. I guess you are also redirected to the Buy page (there's nothing preventing you from going there by modifying the URL anyway) but you can't do anything there.

### An alternative

For me, the simplest alternative is to always show the buttons enabled. This makes everything easier. The only problem is that you can enter the Buy page and find out you can't buy. I think that's OK. And if it's not, then we can add some warning at the top saying "You don't have any <collateral symbol>" or "You don't have any outcome tokens to sell".

gabitoesmiapodo:
@fvictorio 

Just a few thoughts:

- We have a tooltip component in place, so it should be possible to use it or adapt it to show extra information when needed. I think the app should be intuitive enough for us not to use the tooltips, though, so it's better to avoid them if possible.

- Yes, for user engagement it should be possible for them to use the app without barriers until they really need to be connected (as with Amazon's cart logging in on checkout, etc.)

- To avoid confusion with the Buy button situation (and similar situations), we could simply display a text warning near the button. No interaction needed, which would be the case with a tooltip. Something along the lines of `"To do this, you need to do this other thing first..."`, with a link to the thing they need to do. It could be a little frustrating for the user, but at least they're not left in the dark. The user should be always aware of what's going on.

- I'm OK with the "buttons always enabled" attitude, except obviously when you have and invalid form and similar scenarios when it shouldn't be possible to go further.

- While I think most of the app should be self explanatory and intuitive, we should give some attention to the Tutorial / Help issue. That might help users a bit too.

mariano-aguero:
My thoughts, I think the market profile section must show  all the buttons enabled. The buy, sell, and fund section should have the responsibility to perform the action or not, depending on the restrictions of each action.

fvictorio:
Closing this because it won't be necessary in the new design.

PR #373, opened 2020-01-20
Use CPK when creating a market

### LINKED ISSUE #349: Use CPK when creating a market

mariano-aguero:
Closes #349 

#### Includes
- Remove old code to create a market
- Add a function in the CPK service to create the market
- First batch transaction, allow to create the question, prepare the condition and create the market maker
- In a normal transaction, approve your collateral to CPK
- In a third batch transaction, move the collateral to the CPK, approve the collateral to the market maker, add funding, and move pool shares back to the original account 

Note: we have a problem when we try to move the pool shares right now, we don't know the amount of shares for every outcome to move with the function safeTransferFrom, and we can't calculate the amounts inside the batch transaction


fvictorio:
Added a commit that fixes a couple of things:

- The market creation now sends only one transaction (two if it's the first time it's used and the proxy needs collateral approval).
- The funding was wrong. The raw array of probabilities was used instead of computing the proper distribution hint. This means that the initial probabilities weren't working when they were not uniform.
- I had to increase the gas limit to 2 millions, which is a lot (6 USD). We can probably use a more tight number, but the gas cost increases with the number of outcomes. Maybe something we can do is to create markets with 2, 3, 4, etc. outcomes, see how much gas each transaction use, and get some kind of approximate formula for the required gas limit.

app/src/services/realitio.ts
```diff
@@ -166,6 +170,77 @@ class RealitioService {
...
+      SINGLE_SELECT_TEMPLATE_ID,
+      questionText,
+      arbitratorAddress,
+      timeoutResolution,
+      openingTimestamp,
+      0,
+    ]
+
+    const askQuestionInterface = new utils.Interface([
+      'function askQuestion(uint256 template_id, string question, address arbitrator, uint32 timeout, uint32 opening_ts, uint256 nonce) public payable returns (bytes32)',
```
fvictorio: Use ABI definition.

app/src/services/realitio.ts
```diff
@@ -166,6 +170,77 @@ class RealitioService {
...
   }
+
+  static encodeAskQuestion = (
+    question: string,
+    outcomes: Outcome[],
+    category: string,
+    arbitratorAddress: string,
+    openingDateMoment: Moment,
+    networkId: number,
+  ): any => {
```
mariano-aguero: Return string

app/src/services/cpk.ts
```diff
@@ -26,6 +27,7 @@ interface CPKCreateMarketParams {
   conditionalTokens: ConditionalTokenService
   realitio: RealitioService
   marketMakerFactory: MarketMakerFactoryService
+  account: string
 }
 
 interface CPKFundMarketParams {
```
mariano-aguero: Remove this, is unnecesary

Issue #336, opened 2020-01-06
Add button to distribute all probabilities uniformly


fvictorio:
This needs a bit of design. We need a button when adding outcomes that just sets all probabilities to `100 / outcomesCount`.

(Actually, a better way to do this would be to keep track of it being used, and then use an empty distribution hint; then disable this when some probability is manually modified)

mariano-aguero:
@alongoni Can you work on a design for this ? Thank you

alongoni:
Add a checkbox "Distribute Unifomly".
![03b - Conditional Exchange - Step 3 (Distribute %)](https://user-images.githubusercontent.com/622217/72526267-c9098f80-3844-11ea-9871-08ea1cfcc389.png)
@fvictorio @gabitoesmiapodo

alongoni:
May be in this way?. Below "100%" and checkbox on right side.
![03b - Conditional Exchange - Step 3 (Distribute %)](https://user-images.githubusercontent.com/622217/72617441-b7dd8300-3917-11ea-9496-7f7f0dc9adcd.png)
@gabitoesmiapodo

gabitoesmiapodo:
Not entirely convinced, but it'll have to do.

fvictorio:
What do you think about putting it below the "Add new outcome" component and left-aligned?

gabitoesmiapodo:
@fvictorio We tried something like that. It would look even more visually disconnected.

I don't like it 100% aesthetically, but at least it's easily connected to the "Probabilities" column.

I think we might have to think about a different approach for this section. Maybe a dropdown or modal with some settings, I don't know yet.

fvictorio:
Ok, let's just do it as that, at least to have the logic working.

@hpmaxi do you want to do this?

PR #378, opened 2020-01-21
Uniform distribution of outcomes

hpmaxi:
Initial logic for #336


fvictorio:
This will need to be a more complex than this.

First, this will be a checkbox that will disable the probability inputs. They will show `100/outcomes` as its probability, but actually their value won't matter.

When the initial funding is done, a `distributionHint` array is sent to the market. If the array is empty, then the probabilities are uniform. Plus, if the funding is done this way, the transaction costs less gas. So, even if the result is pretty similar to just sending `[100, 100, 100]` (or whatever), it's better to send `[]` for this gas reason.

fvictorio:
Oh, another complexity is that if you disable the checkbox, then the resulting probabilities should be the same as the ones shown. Not a big deal, just something to keep in mind.

hpmaxi:
I've repositioned the checkbox on another row to maintain the total probabilities position in the same column as the outcomes probabilities, but I didn't added colors neither a check/tick image.

app/src/components/common/outcomes/index.tsx
```diff
@@ -173,9 +209,46 @@ const Outcomes = (props: Props) => {
     )
   }
 
+  const addNewOutcome = () => {
```
fvictorio: There's a lot of duplication in these methods. A helper `uniform(outcomes)` should help.

app/src/components/common/outcomes/index.tsx
```diff
@@ -209,45 +209,31 @@ const Outcomes = (props: Props) => {
     )
   }
 
+  const uniform = (outcomes: Outcome[]): Outcome[] => {
+    return outcomes.map(o => {
+      o.probability = 100 / outcomes.length
+      return o
```
fvictorio: I'd rather avoid mutation here:

```
return outcomes.map(o => ({
  ...o,
  probability = 100 / outcomes.length
}))
```

(or something like that; didn't test it).

gabitoesmiapodo:
Design is here: https://app.zenhub.com/workspaces/gnosis-conditional-exchange-5d71043b9b6cb500019ae2f7/issues/protofire/gnosis-conditional-exchange/336#issuecomment-575635633

fvictorio:
I think we should just merge this and change the design in the next milestone. If having a raw HTML checkbox is too ugly, we can just hide it (or use a feature flag and hide it only in production).

leolower:
@fvictorio I agree. The UI Design is going to change anyways so it is better to have this merged.

fvictorio:
@mariano-aguero could you please review this? And if it's OK to you, then just merge it.

PR #388, opened 2020-01-24
Use CPK to reduce the number of Tx necessary for market selling

### LINKED ISSUE #381: Use CPK to reduce the number of Tx necessary for market selling
When selling shares or pool tokens the received DAI should be sent back to the wallet interacting with the proxy account and not stay in the proxy

### LINKED ISSUE #383: Display the existing shares in the CPK

mariano-aguero:
Closes #381 
Closes #383


fvictorio:
This is not transferring the collateral back to the user.

mariano-aguero:
> This is not transferring the collateral back to the user.

I added the approve and the transferring

app/src/hooks/useMarketMakerData.tsx
```diff
@@ -89,15 +91,17 @@ export const useMarketMakerData = (
       fee,
       isQuestionFinalized,
     ] = await Promise.all([
-      account
-        ? marketMaker.getBalanceInformation(account, outcomes.length)
+      account && cpk && cpk.address
```
fvictorio: Is `account` necessary here?

app/src/services/cpk.ts
```diff
@@ -256,6 +263,56 @@ class CPKService {
...
+          operation: CPK.CALL,
+          to: collateralAddress,
+          value: 0,
+          data: ERC20Service.encodeApproveUnlimited(account),
+        },
+        {
+          operation: CPK.CALL,
+          to: collateralAddress,
+          value: 0,
+          data: ERC20Service.encodeTransferFrom(this.cpk.address, account, amount),
```
fvictorio: I don't know if this makes sense. The address that will call the `transferFrom` **is** the CPK, so these two transactions can be replaced with just a `transfer`:

```
{
          operation: CPK.CALL,
          to: collateralAddress,
          value: 0,
          data: ERC20Service.encodeTransfer(account, amount),
}

File: app/src/services/cpk.ts [hunk 3]
```diff
@@ -256,6 +263,56 @@ class CPKService {
...
+  }: CPKSellOutcomesParams): Promise<TransactionReceipt> => {
+    try {
+      const signer = this.provider.getSigner()
+      const account = await signer.getAddress()
+
+      const outcomeTokensToSell = await marketMaker.calcSellAmount(amount, outcomeIndex)
+      const collateralAddress = await marketMaker.getCollateralToken()
+
+      const transactions = [
+        {
```
fvictorio: We should only add this transaction if necessary to reduce the gas used.

app/src/hooks/useMarketMakerData.tsx
mariano-aguero: I think `yes` but not in that way. We need to  check for the account and if exist, we can instantiate the CPK, other way will throw an error. Also this is not working with the disconnected status. I fix this in one of the last commits

app/src/services/cpk.ts
mariano-aguero: I added the transfer method to the ERC20 service, and remove the the two transactiosn that you mention

app/src/services/cpk.ts
mariano-aguero: I added a check to add this transaction in case is not already approve for all

app/src/services/cpk.ts
```diff
@@ -294,15 +288,23 @@ class CPKService {
           operation: CPK.CALL,
           to: collateralAddress,
           value: 0,
-          data: ERC20Service.encodeApproveUnlimited(account),
+          data: ERC20Service.encodeTransfer(account, amount),
         },
-        {
+      ]
+
+      const isAlreadyApprovedForMarketMaker = await conditionalTokens.isApprovedForAll(
```
fvictorio: I think it's more readable to move this to the beginning and use a `push`.

app/src/hooks/useMarketMakerData.tsx
```diff
@@ -78,6 +78,11 @@ export const useMarketMakerData = (
 
     const arbitrator = getArbitratorFromAddress(networkId, arbitratorAddress)
 
+    let cpk: Maybe<CPKService> = null
+    if (account) {
+      cpk = await CPKService.create(provider)
```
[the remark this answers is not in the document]
mariano-aguero: Yes, you're right, and this hook is also used in a section with infura as provider (market details). I I will think in a more  `explicit` solution, let's keep this PR open in the meantime

app/src/services/cpk.ts
mariano-aguero: Sure

app/src/hooks/useMarketMakerData.tsx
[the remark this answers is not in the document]
fvictorio: Let's just merge this.

PR #389, opened 2020-01-24
Add realtio question from url

### LINKED ISSUE #384: Add realit.io URL instead of questionId
Instead of adding the questionId, the input will receive the realit.io URL, and we will have to parse to extract the questionId from the URL

hpmaxi:
Closes #384

### DIFF
--- a/app/src/components/common/question_input/modal_question.tsx
+++ b/app/src/components/common/question_input/modal_question.tsx
@@ -60,10 +60,10 @@ export const ModalQuestion = (props: Props) => {
   const [isSpinnerOn, setSpinnerOn] = useState<boolean>(false)
 
   const extractId = (questionURL: string): string => {
-    const reQuestionId = /0x.*/
+    const reQuestionId = /question\/(0x[0-9A-Fa-f]{64})/
     if (!questionURL) return ''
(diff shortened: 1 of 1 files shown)


app/src/components/common/question_input/modal_question.tsx
```diff
@@ -56,11 +56,19 @@ export const ModalQuestion = (props: Props) => {
 
   const { realitio } = useContracts(context)
 
-  const [questionId, setQuestionId] = useState<string>('')
+  const [questionURL, setQuestionURL] = useState<string>('')
   const [isSpinnerOn, setSpinnerOn] = useState<boolean>(false)
 
+  const extractId = (questionURL: string): string => {
+    const reQuestionId = /0x.*/
```
fvictorio: I would like this regex to be a little safer. What about:

```
/question\/(0x[0-9A-Fa-f]{64})/
```

(We know the expected length because [`questionId` is a bytes32](https://github.com/realitio/realitio-contracts/blob/d7e986c9cba3445680430ea942e49920465f56f9/truffle/contracts/Realitio.sol#L252))

Then we extract the matching group `questionMatch[1]`.

PR #391, opened 2020-01-27
Use CPK to reduce the number of Tx necessary for market funding

### LINKED ISSUE #379: Cannot enter to markets with zero funding
Reproduction steps:

1. Create a new market
2. Remove all funding
3. Refresh I guess?

There's a "division by zero" error in the console so I guess the lack of funding is the problem.

### LINKED ISSUE #380: Use CPK to reduce the number of Tx necessary for market funding
Only adding funding, not removing it.

mariano-aguero:
Closes #380 
Closes #379 

#### Includes
- Add support to add funding with the CPK
- Add support to remove funding with the CPK


app/src/services/cpk.ts
```diff
@@ -247,6 +253,98 @@ class CPKService {
...
+          operation: CPK.CALL,
+          to: collateral.address,
+          value: 0,
+          data: ERC20Service.encodeApproveUnlimited(account),
+        },
+        {
+          operation: CPK.CALL,
+          to: collateral.address,
+          value: 0,
+          data: ERC20Service.encodeTransferFrom(this.cpk.address, account, amount),
```
mariano-aguero: Mmmm what do you think about this last transaction @fvictorio ?

app/src/util/tools.ts
```diff
@@ -45,9 +45,15 @@ export const mulBN = (a: BigNumber, b: number, scale = 10000): BigNumber => {
...
-  const prices = holdings.map(holding => product.div(holding).div(denominator))
+  const denominator = holdings
+    .map(h => (h.toString() === '0' ? new Big(0) : product.div(h)))
+    .reduce((a, b) => a.add(b))
+
+  const prices = holdings.map(holding =>
+    product.toString() === '0' || denominator.toString() === '0'
+      ? new Big(0)
+      : product.div(holding).div(denominator),
+  )
```
mariano-aguero: I think this fix this problem https://github.com/protofire/gnosis-conditional-exchange/issues/379

app/src/services/cpk.ts
```diff
@@ -325,6 +332,83 @@ class CPKService {
...
+          operation: CPK.CALL,
+          to: collateral.address,
+          value: 0,
+          data: ERC20Service.encodeTransferFrom(account, this.cpk.address, amount),
+        },
+        {
+          operation: CPK.CALL,
+          to: collateral.address,
+          value: 0,
+          data: ERC20Service.encodeApproveUnlimited(marketMaker.address),
```
fvictorio: We should check if this transaction is necessary.

app/src/services/cpk.ts
```diff
@@ -325,6 +332,83 @@ class CPKService {
...
+    marketMaker,
+  }: CPKFundingParams): Promise<TransactionReceipt> => {
+    try {
+      const transactions = [
+        {
+          operation: CPK.CALL,
+          to: collateral.address,
+          value: 0,
+          data: ERC20Service.encodeApproveUnlimited(this.cpk.address),
+        },
```
fvictorio: Why are you doing this approval?

app/src/util/tools.ts
fvictorio: This is probably fine for now, but the approach is unnecessarily involved. I would just add a check at the beginning of the function: if any holding is 0, then return all zeroes. (In practice, I think that either *all* holdings are zero, or all are non-zero, but anyway).

app/src/services/cpk.ts
mariano-aguero: I added a check for the allowance, and a push if is necessary

mariano-aguero: Right here, this is a transaction with the approveUnlimited https://rinkeby.etherscan.io/tx/0x4ea69c9dd20dbfe268d1a139dbfc3ac4032bb5f55dcd74f7093d2c98c591fd6e#internal

Without the approveUnlimited
https://rinkeby.etherscan.io/tx/0x7b2529fa7f4e8a42f2ba816c0fe90b031511d46d90ef98488eedf00615f3746d#internal

app/src/services/cpk.ts
mariano-aguero: Mmmm is not necessary, I will remove it

app/src/util/tools.ts
```diff
@@ -44,16 +44,16 @@ export const mulBN = (a: BigNumber, b: number, scale = 10000): BigNumber => {
  */
 export const calcPrice = (holdingsBN: BigNumber[]): number[] => {
   const holdings = holdingsBN.map(h => new Big(h.toString()))
+
+  const doesHoldingsAreZero = holdings.every(h => h.toString() === '0')
```
fvictorio: `hasZeroHoldings` or anything else with proper grammar.

mariano-aguero: I will fix this...

PR #395, opened 2020-01-28
Reuse conditionID when we add a realit.io question

### LINKED ISSUE #363: Reuse conditionID when we add a realit.io question
Check if the question was already used to prepare a condition before. If it was, reuse that conditionId (skipping the second step).

mariano-aguero:
Closes #363

### DIFF
--- a/app/src/services/conditional_token.ts
+++ b/app/src/services/conditional_token.ts
@@ -82,7 +82,7 @@ class ConditionalTokenService {
   getQuestionId = async (conditionId: string): Promise<string> => {
     const filter: any = this.contract.filters.ConditionPreparation(conditionId)
 
-    const network = await this.provider.ready
+    const network = await this.provider.getNetwork()
     const networkId = network.chainId
 
     const logs = await this.provider.getLogs({
@@ -185,7 +185,7 @@ class ConditionalTokenService {
   }
 
   getConditionIdFromLogs = async (conditionId: string): Promise<ConditionLog> => {
-    const network = await this.provider.ready
+    const network = await this.provider.getNetwork()
     const networkId = network.chainId
 
     const filter: any = this.contract.filters.ConditionPreparation(conditionId)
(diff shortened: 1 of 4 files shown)


app/src/services/cpk.ts
```diff
@@ -182,18 +182,22 @@ class CPKService {
...
+          logger.log(
+            `ConditionId found: '${conditionId}', Outcome slot count: '${outcomeSlotCount}'`,
+          )
         } catch (err) {
           logger.log(`ConditionId is not already prepared: '${err.message}'`)
         }
       }
 
-      if (!conditionLog) {
+      if (outcomeSlotCount.isZero()) {
```
fvictorio: Maybe we can hide this complexity behind the service and just have a method `conditionExists` or something like that.

app/src/services/conditional_token.ts
```diff
@@ -178,6 +193,40 @@ class ConditionalTokenService {
 
     return conditionId
   }
+
+  getConditionIdFromLogs = async (conditionId: string): Promise<ConditionLog> => {
```
mariano-aguero: This function is not necessary, but I think it does not affect if we keep it, and maybe could be usefull

fvictorio: We either keep this function and remove `getQuestionId`, or we remove this one. We can't keep both of them, they are almost equal.

app/src/services/cpk.ts
mariano-aguero: Done!

app/src/services/conditional_token.ts
mariano-aguero: Remove it

app/src/services/conditional_token.ts
```diff
@@ -194,38 +193,9 @@ class ConditionalTokenService {
...
-
-    const { oracle, questionId, outcomeSlotCount } = event.values
-
-    return {
-      conditionId,
-      oracle,
-      questionId,
-      outcomeSlotCount,
-    }
+  isConditionExists = async (conditionId: string): Promise<boolean> => {
```
fvictorio: Either `conditionExists` or `doesConditionExist`.

PR #401, opened 2020-01-30
Check for funding on account change

### LINKED ISSUE #394: Validate amount of funding during market creation
If the user hasn't connected a wallet, they can go through the market creation wizard but in the last step the submit button is replaced by a "Connect wallet" button. After the wallet is connected, the submit button appears.

In the funding step, if the wallet is not connected, the amount of funding entered is not validated. This means that the user can use an amount greater than the balance of the account they will connect in the last step. When this happens, the transaction just fails.

This should be handled correctly. I think just disabling the submit button and showing an error message (something like `You entered 10 DAI of funding but your account only has 5 DAI`) should be enough. The user should be able to go back and change the amount.

hpmaxi:
Closes #394


app/src/components/market/steps/items/create_market_step.tsx
```diff
@@ -204,9 +197,11 @@ const CreateMarketStep = (props: Props) => {
...
           buttonType={ButtonType.primary}
           disabled={
-            (!MarketCreationStatus.is.ready(marketCreationStatus) &&
-              !MarketCreationStatus.is.error(marketCreationStatus)) ||
-            (!hasEnoughBalanceToFund && !!account)
+            !(
+              MarketCreationStatus.is.ready(marketCreationStatus) &&
+              !MarketCreationStatus.is.error(marketCreationStatus) &&
+              hasEnoughBalance
+            ) && !!account
```
fvictorio: This is too hard to read. I think the problem is that we are merging two buttons into one. A better approach might be something like:

```typescript
account ?
<Button disabled={!MarketCreationStatus.is.ready(marketCreationStatus) || ...} ... >Create</Button>
: <Button ...>Connect Wallet</Button>
```

They are going to have some props in common, but that's probably fine.

An alternative is to extract this in a variable:

```
const isButtonDisabled = account ? (!MarketCreation.is.ready(...) ...) : false
```

I think I prefer the first approach, but I'm ok with either one.

app/src/components/market/steps/items/create_market_step.tsx
```diff
@@ -104,33 +106,24 @@ const CreateMarketStep = (props: Props) => {
   const resolutionDate = resolution && formatDate(resolution)
 
   const buttonText = account ? 'Create' : 'Connect Wallet'
+  const balance = useSelector((state: any) => state.balance && new BigNumber(state.balance))
```
fvictorio: @hpmaxi `balance` is going to be `any` here, right? Can we do better than that?

app/src/components/market/steps/items/create_market_step.tsx
hpmaxi: I opted for split them in two

app/src/components/market/steps/items/create_market_step.tsx
hpmaxi: I added some types to that selector

fvictorio:
Looks fine. I think the next step is to either remove `useCollateralBalance` and replace all its usages with `useSelector`, or to rewrite it to use `useSelector` under the hood. I kind of prefer the second approach, but either one is fine.

If we rewrite it, then there's some implicit semantics here: if you use `useCollateralBalance`, then you get a 0 balance either if the user has no balance or if the balance was not fetched yet. If you use `useSelector` directly, you get a `Maybe<BigNumber>` and so you can know if the value exists. (Notice that we can't distinguish between "the user is not connected" and "the balance is being fetched", but that's fine for now.)

@hpmaxi your call. But either way, let's try to get this merged today.

PR #402, opened 2020-01-31
The app crashes when an unsupported network is used

### LINKED ISSUE #392: App doesn't work when metamask is not installed
The app should be usable when the user doesn't have metamask installed. They should be able to use everything that is available without connection. When the wallet connection modal is shown, the metamask option should be disabled.

### LINKED ISSUE #393: The app crashes when an unsupported network is used
If the user is connected to, for example, Görli, the app just crashes. We should show a message like "Please connect to Mainnet or to one of the supported testnets (Rinkeby, Kovan)"

mariano-aguero:
Closes #392 
Closes #393 

#### Includes
- Check if a web3 (legacy) or ethereum object is inyected in the browser (Metamask installed ? )
- Remove metamask item in the modal connection if metamask is not installed
- Validate wrong network selection in the wallet

### DIFF
--- a/app/src/components/common/modal_connect_wallet/index.tsx
+++ b/app/src/components/common/modal_connect_wallet/index.tsx
@@ -84,6 +84,8 @@ export const ModalConnectWallet = (props: Props) => {
 
   if (context.error) {
     logger.error('Error in web3 context', context.error)
+    localStorage.removeItem('CONNECTOR')
+    props.onClose()
   }
 
   const [walletSelected, setWalletSelected] = useState(Wallet.MetaMask)
--- a/app/src/components/common/modal_connect_wallet/index.tsx
+++ b/app/src/components/common/modal_connect_wallet/index.tsx
@@ -15,13 +15,17 @@ import { lighten } from 'polished'
 
 const logger = getLogger('ModalConnectWallet::Index')
 
-const Item = styled.div`
+const Item = styled.div<{ disabled?: boolean }>`
   border-bottom: 1px solid ${props => props.theme.borders.borderColor};
   cursor: pointer;
   display: flex;
   justify-content: space-between;
   padding: 15px 5px;
-
+  pointer-events: ${props => (props.disabled ? 'none' : 'auto')};
+  &[disabled] {
+    cursor: not-allowed !important;
+    opacity: 0.5;
+  }
   &:hover {
     background-color: ${props => lighten(0.6, props.theme.colors.primary)};
   }
@@ -88,7 +92,11 @@ export const ModalConnectWallet = (props: Props) => {
     props.onClose()
   }
 
-  const [walletSelected, setWalletSelected] = useState(Wallet.MetaMask)
+  const doesMetamaskExist = 'ethereum' in window || 'web3' in window
+
+  const [walletSelected, setWalletSelected] = useState(
+    doesMetamaskExist ? Wallet.MetaMask : Wallet.WalletConnect,
+  )
 
   const onClickWallet = (wallet: Wallet) => setWalletSelected(wallet)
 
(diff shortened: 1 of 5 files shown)


app/src/components/common/modal_connect_wallet/index.tsx
```diff
@@ -155,12 +157,14 @@ export const ModalConnectWallet = (props: Props) => {
...
 
+  const doesMetamaskExist = 'ethereum' in window || 'web3' in window
+
   return (
     <>
       {!context.account && (
         <ModalWrapper isOpen={isOpen} onRequestClose={onClickCloseButton} title={`Choose a Wallet`}>
-          <ItemMetamask />
-          <ItemWallet />
+          {doesMetamaskExist && <ItemMetamask />}
```
fvictorio: I think it's better to disable it rather than hiding it. @leolower @gabitoesmiapodo wdyt?

app/src/components/common/wrong_network_message/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+}
+
+const WrongNetworkMessageContainer = (props: Props) => {
+  return (
+    <>
+      <SectionTitle title="THERE WAS A PROBLEM..." />
+      <ViewCard>
+        <SubsectionTitle>Wrong network configuration</SubsectionTitle>
+        <WrappedText>
+          We couldn&apos;t find the network with the provided wallet configuration.
```
fvictorio: I'm pretty sure (but not 100% sure) that you can use `'` in JSX.

app/src/components/common/wrong_network_message/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+        </WrappedText>
+        <WrappedText>
+          The valid networks to use are{' '}
+          <strong>
+            {Object.keys(networkIds)
+              .join(', ')
+              .toLowerCase()}
```
fvictorio: Let's show the labels instead of the network ids. Let's also filter ganache out since most users won't care.

app/src/components/common/wrong_network_message/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+          </strong>
+          .
+        </WrappedText>
+        <WrappedText>
+          Please check the network configuration in your wallet and come back again.
+        </WrappedText>
+        <ButtonContainer>
+          <ButtonLinkStyled onClick={() => window.location.reload(false)}>
+            Reload with Infura
```
fvictorio: Not sure about this. It seems to me like exposing an implementation detail to the user. Just `Reload` would be fine.

app/src/components/common/modal_connect_wallet/index.tsx
gabitoesmiapodo: Yes.

mariano-aguero: I will need to add some opacity, like in this image, because this component is a Div
Screenshot_20200204_181840

gabitoesmiapodo: That's fine.

mariano-aguero: Done!

app/src/components/common/wrong_network_message/index.tsx
mariano-aguero: We will incur in this linter error 
Screenshot_20200204_183837

app/src/components/common/wrong_network_message/index.tsx
mariano-aguero: The keys are the labels, I will filter ganache.

app/src/components/common/wrong_network_message/index.tsx
mariano-aguero: Rewording

app/src/components/common/wrong_network_message/index.tsx
mariano-aguero: Screenshot_20200204_185433

PR #431, opened 2020-02-04
Get list of collateral tokens from Kleros TCR

### LINKED ISSUE #422: Get list of collateral tokens from Kleros TCR

mariano-aguero:
Closes #422 

#### Includes
- Added a service to fetch tokens from kleros interaction contract
- We can validate the amount of tokens, we have 395 in https://tokens.kleros.io/tokens , and the same quantity in the dropdown to select
- We don't have the contract deployed in the docker, also doesn't exist in rinkeby, so if the user use rinkeby or ganache we choose the old tokens that we have in the networks file.


mariano-aguero:
I refactor the old code to a more uniswap style, we only fetch tokens with a badge. 
So first of all, we fetch addresses of tokens that have the badge. Then we use a[ view contract](https://etherscan.io/address/0xf9b9b5440340123b21bff1ddafe1ad6feb9d6e7f#code) to fetch their submission IDs on the TCR. Later we return all the available token data at once. Warning: some token contracts do not implement the decimals function, I added a fallback to 18 decimals.
To take into account, the TCR contract contains all token submissions, a token submission contains the token's data its current status and request history (name, ticker, address, decimals, etc)

app/src/services/kleros.ts
```diff
@@ -0,0 +1,133 @@
...
+      // With the token IDs, get the information and add it to the object.
+      const fetchedTokens = await this.tokensViewContract.getTokens(klerosTCRAddress, submissionIDs)
+      const tokens = fetchedTokens
+        .filter((tokenInfo: any) => tokenInfo[3] !== ethers.constants.AddressZero)
+        .map(
+          (token: any): Token => {
+            return {
+              symbol: token.ticker,
+              address: token.addr,
+              decimals: token.decimals.toNumber() || 18, // Token does not implement the 'decimals()' function, falling back to the default, 18 decimal places.
```
mariano-aguero: I think we need consensus right here, how to manage this situation. Some tokens like DGD may not implement the decimals function, maybe we can manage some situations with an array of constants or another pattern, what do you think ?

app/src/util/networks.ts
```diff
@@ -75,6 +81,9 @@ const networks: { [K in NetworkId]: Network } = {
       marketMakerFactory: '0x89F54625565792729fd3Fa5F689c46e7612b4E8b',
       conditionalTokens: '0xe6Cdc22F99FD9ffdC03647C7fFF5bB753a4eBB21',
       oracle: '0x5A90132C104CA33652Ee71E2E645D5A4d8E5d7D6',
+      klerosBadge: '0x0',
+      klerosTokenView: '0x0',
+      klerosTCR: '0x0',
```
mariano-aguero: We don't have support for rinkeby :(

app/src/services/kleros.ts
```diff
@@ -0,0 +1,133 @@
...
+import { getContractAddress, getTokensByNetwork, networkIds } from '../util/networks'
+import { Web3Provider } from 'ethers/providers'
+
+const klerosBadgeAbi = [
+  'function queryAddresses(address _cursor, uint _count, bool[8] _filter, bool _oldestFirst) external view returns (address[] values, bool hasMore)',
+]
+
+const klerosTokensViewAbi = [
+  'function getTokensIDsForAddresses(address _t2crAddress, address[] _tokenAddresses ) external view returns (bytes32[] result)',
+  'function getTokens(address _t2crAddress, bytes32[] _tokenIDs ) external view returns (tuple(bytes32 ID, string name, string ticker, address addr, string symbolMultihash, uint8 status, uint256 decimals)[] result)',
```
mariano-aguero: Check this, how to manage a struct (tuple) with human readable abi

app/src/services/kleros.ts
fvictorio: Kleros's Uniswap fork falls back to 18, so I think it's ok to do the same.

app/src/services/kleros.ts
fvictorio: Well, if it works, it works.

fvictorio:
I'll try to deploy this in Rinkeby and our docker image.

app/src/services/kleros.ts
```diff
@@ -0,0 +1,133 @@
...
+        addressesWithBadge = addressesWithBadge.concat(
+          result[0].filter((address: string) => address !== ethers.constants.AddressZero),
+        )
+        hasMore = result.hasMore
+      }
+
+      // Fetch their submission IDs on the TCR.
+      const network = await this.provider.getNetwork()
+      const networkId = network.chainId
+      const klerosTCRAddress = getContractAddress(networkId, 'klerosTCR')
```
fvictorio: I would add this as a constructor parameter and then use it as an instance variable.

app/src/services/kleros.ts
```diff
@@ -0,0 +1,133 @@
...
+            }
+          },
+        )
+
+      const basicOrder = ['dai', 'cdai', 'weth', 'usdc', 'owl', 'chai'].reverse()
+
+      return tokens.sort((tokenA: Token, tokenB: Token) => {
+        let sortBy: number
+
+        const indexOfB = basicOrder.indexOf(tokenB.symbol.toLowerCase())
```
fvictorio: This is going to be extremely pedantic but I'm going to do it anyway: A should go before B.

app/src/services/kleros.ts
```diff
@@ -0,0 +1,133 @@
...
+  tokensViewContract: Contract
+
+  constructor(
+    badgeContractAddress: string,
+    tokensViewContractAddress: string,
+    provider: Web3Provider,
+    signerAddress: Maybe<string>,
+  ) {
+    this.provider = provider
+    if (signerAddress) {
```
fvictorio: I know this is the way we always do it, but I'm wonder if

```typescript
      this.badgeContract = new ethers.Contract(
        badgeContractAddress,
        klerosBadgeAbi,
        provider,
      )
      this.tokensViewContract = new ethers.Contract(
        tokensViewContractAddress,
        klerosTokensViewAbi,
        provider,
      )
      if (signerAddress) {
        this.badgeContract = this.badgeContract.connect(signer)
        this.tokensViewContract =  this.tokensViewContract(signer)
      }
```

isn't better. I'm on the fence with this one so either approach is ok by me.

app/src/services/kleros.ts
mariano-aguero: Done!

app/src/services/kleros.ts
mariano-aguero: Done!

app/src/services/kleros.ts
mariano-aguero: Done!

fvictorio:
I was able to deploy this to ganache, but without tokens. Adding our tokens to the list it's surely doable but I think it's going to take me a lot of time that I don't think it's worth it. Let's continue with the approach of using our hardcoded tokens in ganache for now. If at some point is really important to have this feature in a network other than mainnet, we can just add Kovan to our supported networks.

I will review the code again when the merge conflicts are resolved.

mariano-aguero:
> I was able to deploy this to ganache, but without tokens. Adding our tokens to the list it's surely doable but I think it's going to take me a lot of time that I don't think it's worth it. Let's continue with the approach of using our hardcoded tokens in ganache for now. If at some point is really important to have this feature in a network other than mainnet, we can just add Kovan to our supported networks.
> 
> I will review the code again when the merge conflicts are resolved.

I solved the conflicts ...

PR #452, opened 2020-03-03
Markets overview with subgraph

hpmaxi:
- Add default endpoint to thegraph/omen (rinkeby)
- Change list_item to show new information of each market, when the issue with open/closed markets we'll need to add a condition for the timestamp (Ends in ...)
- useQuery, fetchMore of apollo client used in market_home to query the graph, pagination working.
- marketMakerFactory address changed to use the new one


app/src/components/market/market_home_container.tsx
```diff
@@ -13,25 +14,66 @@ const PAGE_SIZE = 10
...
+  }, [data, loading, error])
+
+  const showMore = () => {
+    fetchMore({
+      variables: {
+        skip: data.fixedProductMarketMakers.length,
+      },
+      updateQuery: (prev, { fetchMoreResult }) => {
+        if (!fetchMoreResult) return prev
+        return Object.assign({}, prev, {
```
fvictorio: I prefer `{ ...a, ...b }` over `Object.assign({}, a, b)`

app/src/components/market/market_home_container.tsx
```diff
@@ -13,25 +14,66 @@ const PAGE_SIZE = 10
...
+          fixedProductMarketMakers: [...prev.fixedProductMarketMakers, ...fetchMoreResult.fixedProductMarketMakers],
+        })
+      },
+    })
+  }
+  const onFilterChange = (filter: any) => {
```
fvictorio: Since this doesn't have extra logic anymore, we can just do `onFilterChange={setFilter}` below.

app/src/components/common/list_item/index.tsx
```diff
@@ -136,21 +139,52 @@ const ListItemWrapper: React.FC<ListItemWrapperProps> = (props: ListItemWrapperP
...
+  const { outcomes, title } = question
+
+  const endsIn = moment(new Date(question.openingTimestamp * 1000)).fromNow()
+
+  useEffect(() => {
+    const setToken = async () => {
+      if (!account) {
+        return
+      }
+      const erc20Service = new ERC20Service(provider, account, collateralToken)
```
fvictorio: How does this work? `account` is the current connected address, not an erc20 :confused:

hpmaxi: `constructor(provider: any, signerAddress: Maybe<string>, tokenAddress: string) {`

The third argument is an erc20, maybe I could pass null on the second argument.

fvictorio: Oh, no, my bad, sorry.

PR #458, opened 2020-03-11
Bug/continous refetch

hpmaxi:


app/src/components/market/market_home_container.tsx
```diff
@@ -48,14 +50,20 @@ const MarketHomeContainer: React.FC = () => {
...
+    setFilter(filter)
+  }
   const showMore = () => {
+    if (!moreMarkets) return
     fetchMore({
       variables: {
         skip: fetchedMarkets.fixedProductMarketMakers.length,
       },
       updateQuery: (prev, { fetchMoreResult }) => {
+        setMoreMarkets(!!fetchMoreResult.fixedProductMarketMakers.length)
```
fvictorio:
```suggestion
        setMoreMarkets(fetchMoreResult.fixedProductMarketMakers.length > 0)
```

app/src/components/market/market_home_container.tsx
```diff
@@ -38,7 +40,7 @@ const MarketHomeContainer: React.FC = () => {
   }, [provider])
 
   useEffect(() => {
-    if (loading) {
+    if (loading && moreMarkets) {
```
fvictorio: I this necessary? The way I see this, when the last fetch is done:

- fetchMore is called, loading is set to true
- data is returned, loading is set to false (and so this condition is false) and fetched data is []
- since data is [], `moreMarkets` is set to false, but loading is still false so this condition continues to be false

Am I missing something?

hpmaxi: No, you're right. It's removed now.

PR #459, opened 2020-03-11
24hs volume

hpmaxi:
This calculates the collateral volume of the last day.


app/src/components/market/market_view_container.tsx
```diff
@@ -30,6 +74,18 @@ const MarketViewContainer: React.FC<Props> = (props: Props) => {
     resolution,
   } = marketMakerData
 
+  useEffect(() => {
+    const get24hsVolume = async () => {
+      const BLOCKS_PER_SECOND = 15
+      const OFFSET = (60 * 60 * 24) / BLOCKS_PER_SECOND
```
fvictorio: Add `Math.round`

app/src/components/market/market_view_container.tsx
```diff
@@ -1,21 +1,65 @@
...
+    variables: { id: marketMakerAddress.toLowerCase() },
+  })
+
+  const { data: volumeBefore, error: errorVolumeBefore } = useQuery(buildQuery24hsEarlier(hash && hash.toLowerCase()), {
+    skip: !!lastDayVolume,
+    variables: { id: marketMakerAddress.toLowerCase() },
+  })
+
+  if (errorVolumeBefore || errorVolumeNow) {
+    setLastDayVolume('-')
```
fvictorio: Log error.

app/src/components/market/market_view_container.tsx
```diff
@@ -1,21 +1,65 @@
...
+  const { library: provider } = context
+
+  const [lastDayVolume, setLastDayVolume] = useState<Maybe<string>>(null)
+
+  const { data: volumeNow, error: errorVolumeNow } = useQuery(GET_COLLATERAL_VOLUME_NOW, {
+    skip: !!lastDayVolume,
+    variables: { id: marketMakerAddress.toLowerCase() },
+  })
+
+  const { data: volumeBefore, error: errorVolumeBefore } = useQuery(buildQuery24hsEarlier(hash && hash.toLowerCase()), {
```
fvictorio: Does this mean that while `hash` is null, the query will be performed with `null`? Doesn't this cause any problems?

app/src/components/market/market_view_container.tsx
```diff
@@ -1,21 +1,65 @@
...
+    variables: { id: marketMakerAddress.toLowerCase() },
+  })
+
+  if (errorVolumeBefore || errorVolumeNow) {
+    setLastDayVolume('-')
+  } else if (volumeNow && volumeBefore) {
+    const now = new BigNumber(volumeNow.fixedProductMarketMakers[0].collateralVolume)
+    const before = new BigNumber(volumeBefore.fixedProductMarketMakers[0].collateralVolume)
+
+    setLastDayVolume(now.sub(before).toString())
```
fvictorio: I think `lastDayVolume` should be a BigNumber. Converting it to string is a view responsibility.

app/src/components/market/market_view_container.tsx
fvictorio: `skip` can be used here, right? Is using `skip: !!lastDayVolume || !hash` below a good solution?

hpmaxi: Yes, seems like a good solution to me.

PR #461, opened 2020-03-13
Add logic for set allowance button

fvictorio:
This PR adds a "Set Allowance" button in the buy and fund pages. This doesn't have any styling.

The button works like this:
- It's always shown if the allowance is 0
- If the allowance is bigger than zero, then the button is shown if the value entered is bigger than the allowance.

### Testing

To manually test this you need to be able to set the allowance to an arbitrary value. You can do this with eth-cli:

```
$ eth repl --rinkeby erc20@0x5592ec0cfb4dbc12d3ab100b257153436a1f0fea --pk <private-key>
> erc20.methods.approve(<cpk-address>, 0).send({from:accounts[0], gas: '1000000'})
```

(To get the address of the CPK for your account, just add some log in the app, for example in `useCpk`)

### Other changes

- Add a try/catch to avoid an error when the wallet is not connected. This error only shows up in development, but when it appears it doesn't let you use the app.
- Add typings for contract-proxy-kit
- Add `react-hooks` to eslint


app/src/components/market/market_buy.tsx
```diff
@@ -247,13 +256,13 @@ const MarketBuyWrapper: React.FC<Props> = (props: Props) => {
...
                 placeholderText={collateral.symbol}
               />
-              <ToggleTokenLock amount={amount} collateral={collateral} context={context} />
             </>
           }
           note={noteAmount}
           title={'Total cost'}
           tooltip={{ id: 'amount', description: 'Shares to buy with this amount of collateral.' }}
         />
+        {showSetAllowance && <button onClick={unlockCollateral}>Set allowance{allowance === null && '...'}</button>}
```
hpmaxi: When allowance is null the button could be disabled

app/src/components/market/market_fund.tsx
```diff
@@ -179,6 +184,16 @@ const MarketFundWrapper: React.FC<Props> = (props: Props) => {
 
   const probabilities = balances.map(balance => balance.probability)
 
+  const unlockCollateral = async () => {
+    if (!cpk) {
+      return
+    }
```
hpmaxi: Unlock also checks for cpk being truthy.

fvictorio: I don't understand this comment. Just for the record, this is mainly so that typescript doesn't complain, because if this method is called then `cpk` is almost surely not null.

app/src/components/market/market_buy.tsx
fvictorio: Fixed, thanks.

PR #463, opened 2020-03-13
Details integration

hpmaxi:
This integrates with 24hs volume and the new design but doesn't have my earnings and my pool earnings. 

Depends on #459

### DIFF
--- a/app/src/components/market/market_view_container.tsx
+++ b/app/src/components/market/market_view_container.tsx
@@ -42,18 +42,19 @@ const MarketViewContainer: React.FC<Props> = (props: Props) => {
 
   const [lastDayVolume, setLastDayVolume] = useState<Maybe<string>>(null)
 
+  // TODO Move all this to another component
   const { data: volumeNow, error: errorVolumeNow } = useQuery(GET_COLLATERAL_VOLUME_NOW, {
     skip: !!lastDayVolume,
     variables: { id: marketMakerAddress.toLowerCase() },
   })
 
   const { data: volumeBefore, error: errorVolumeBefore } = useQuery(buildQuery24hsEarlier(hash && hash.toLowerCase()), {
-    skip: !!lastDayVolume,
+    skip: !!lastDayVolume || !hash,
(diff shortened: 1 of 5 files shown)


app/src/components/market/profile/view.tsx
```diff
@@ -88,12 +104,19 @@ const ViewWrapper = (props: Props) => {
               }
             />
           )}
+          <TitleValue title="24h Volume" value={lastDayVolume ? `${lastDayVolume} ${collateral.symbol}` : '-'} />
```
fvictorio: This should be formatted using the collateral decimals.

app/src/components/market/profile/view.tsx
```diff
@@ -88,12 +104,19 @@ const ViewWrapper = (props: Props) => {
...
           {showExtraDetails ? (
             <>
-              <TitleValue title={'Mocked Title 1'} value={'Mocked Value 1'} />
-              <TitleValue title={'Mocked Title 2'} value={'Longer Mocked Value to have and idea how it looks'} />
-              <TitleValue title={'Mocked Title 3'} value={'Mocked Value 3'} />
-              <TitleValue title={'Mocked Title 4'} value={'Mocked Value 4'} />
+              <TitleValue
+                title={'Total Pool Tokens'}
+                value={totalPoolShares ? formatBigNumber(totalPoolShares, collateral.decimals) : '0'}
```
fvictorio: The pool shares don't necessarily have the same number of decimals as the collateral.

app/src/components/market/market_view_container.tsx
```diff
@@ -88,7 +91,7 @@ const MarketViewContainer: React.FC<Props> = (props: Props) => {
       setHash(hash)
     }
 
-    get24hsVolume()
+    provider && get24hsVolume()
```
fvictorio: PP: I don't like this idiom (`condition && statement-with-side-effects`). Besides... `provider` is not nullable here, is it? If you got an error with it being undefined, then we have a typing problem somewhere.

PR #485, opened 2020-03-26
Markets home remaining filters

### LINKED ISSUE #408: Add new filters

### LINKED ISSUE #455: Complete market list data and filters
- [ ] Use category in filters, and null if 'All' is selected. `category` should be at level 0 of markets nodes
- [ ] Add input text to filter by text, `title` should be at level 0 of markets nodes ( ... question ) 
- [ ] Change queries to receive `isResolved` to fetch open or closed markets. isResolved isn't in the schema at this moment.
- [ ] Add filter for `collateral token`, `arbitrator` and question type in the UI (Only one question type is available, maybe this is not necessary). This need arbitrator (...question) at the market level
- [ ] sortBy volume, seems like a select in the design but we're only sorting by volume
- [ ] Check different error messages when markets fetching results in an empty set. And loading messages (See design with extra filters)
- [ ] Improve types. Create a new type for MarketWithExtraData or something similar.

hpmaxi:
Closes #408 
Closes #455


app/src/components/market/advanced_filters/index.tsx
```diff
@@ -39,57 +41,60 @@ const Options = styled(Dropdown)`
...
+export const AdvancedFilters = (props: Props) => {
+  const context = useConnectedWeb3Context()
+  const { networkId } = context
+
+  const arbitrators = getArbitratorsByNetwork(networkId)
+  const tokens = getTokensByNetwork(networkId)
+
+  const { onChangeArbitrator, onChangeCurrency, onChangeTemplateId } = props
+
+  const currencyOptions: Array<DropdownItemProps> = [{ address: null, symbol: 'All' }, ...tokens].map(
```
hpmaxi: By default all markets are fetched, so I added an 'All' option. 
I'm not sure if it's necessary to maintain the `currentItem` selected in the dropdown on the component that uses AdvancedFilters.

app/src/queries/markets_home.ts
```diff
@@ -13,43 +13,52 @@ const MarketDataFragment = gql`
...
     category
+    answerFinalizedTimestamp
   }
 `
 
-const OPEN = gql`
-  query GetMarkets($first: Int!, $skip: Int!, $sortBy: String) {
-    fixedProductMarketMakers(first: $first, skip: $skip, orderBy: $sortBy) {
-      ...marketData
+export const buildQueryMarkets = (
```
fvictorio, line 20: This needs tests.

app/src/queries/markets_home.ts
```diff
@@ -13,43 +13,52 @@ const MarketDataFragment = gql`
...
-      ...marketData
-    }
-  }
-  ${MarketDataFragment}
-`
-
-export const MARKETS_HOME: any = { OPEN, CLOSED, CORONA, MY_MARKETS }
+    ${MarketDataFragment}
+  `
+  console.log(query)
```
fvictorio, line 62: Delete.

app/src/queries/markets_home.ts
```diff
@@ -13,43 +13,52 @@ const MarketDataFragment = gql`
...
+    isCoronaVersion,
+  } = options
+  const whereClause = [
+    onlyClosedMarkets ? 'answerFinalizedTimestamp_not: null' : '',
+    onlyMyMarkets || isCoronaVersion ? 'creator_in: $accounts' : '',
+    category === 'All' ? '' : 'category: $category',
+    title ? 'title_contains: $title' : '',
+    currency ? 'collateralToken: $currency' : '',
+    arbitrator ? 'arbitrator: $arbitrator' : '',
+    templateId ? 'templateId: $templateId' : !isCoronaVersion ? 'templateId_in: ["0", "2","6"]' : '',
```
fvictorio, line 49:
```suggestion
    templateId ? 'templateId: $templateId' : !isCoronaVersion ? 'templateId_in: ["0", "2", "6"]' : '',
```

app/src/components/market/advanced_filters/index.tsx
```diff
@@ -39,57 +41,60 @@ const Options = styled(Dropdown)`
...
+    {
+      content: 'All',
+      onClick: () => onChangeTemplateId(null),
+    },
     {
       content: 'Binary',
-      onClick: () => {
-        console.warn('Option Binary')
-      },
+      onClick: () => onChangeTemplateId('0'),
```
fvictorio, line 82: Template ids are numbers (actually big numbers, but I don't think that will be necessary soon).

app/src/components/market/advanced_filters/index.tsx
fvictorio: This is not very idiomatic, but it's a problem that seems to be caused by `<Dropdown />`. This will do for now.

PR #490, opened 2020-03-30
(Feature) Market creation styles

### LINKED ISSUE #421: Apply new design

gabitoesmiapodo:
Closes #421

**TO-DO:** 

- [x] Check that everything that used to work still works. I was able to create a market, so that works, but there's new (and old) stuff that should be checked and fixed if necessary.
- [x] Currency selection is new (it's a new, completely different component), so it's not working.
- [x] ~~Details block on the second step is not working, or not fully working.~~ Moved to another issue
- [x] Third step "View Market" should be removed. Redirect to market details instead.
- [x] Date picker's styles are not done yet, but that shouldn't interfere with anything else, so I'm currently working on that.

**Screen captures:**

set max

fill

add

create


gabitoesmiapodo:
Datepicker's styles should be ready now.

Screen Shot 2020-04-01 at 13 06 32

hpmaxi:
Details block issue #499

PR #501, opened 2020-04-02
Fix wording in market list (opening date)

hpmaxi:


app/src/components/common/list_item/index.tsx
```diff
@@ -71,7 +71,9 @@ export const ListItem: React.FC<Props> = (props: Props) => {
     title,
   } = market
 
-  const endsIn = moment(new Date(openingTimestamp * 1000)).fromNow()
+  const endDate = moment(new Date(openingTimestamp * 1000))
+  const diff = endDate.diff(moment())
```
fvictorio, line 75: I think moment has a `isBefore` and `isAfter` methods.

PR #502, opened 2020-04-03
Fix redeem process

fvictorio:
This PR fixes the resolve condition/redeem workflow for nuanced binary markets.

Before the PR, we were assuming that there was a single winning outcome, and we'd show the Resolve/Redeem buttons if the connected user had shares for that outcome.

This is wrong for two reasons. First, now he have nuanced binary markets, that might not have a single winning outcome. Second, any question can have an "undefined" outcome, where all shares are redeemable for 1/N of collateral (where N is the amount of outomes).

This PR fixes that by computing the array of payouts for a given Realitio answer. This is done in OracleService, where a new `getPayouts` function was added. This function does the same that is done on-chain in the RealitioProxy contract, see [here](https://github.com/fvictorio/realitio-gnosis-proxy/blob/ed306cb5e4ea7d831df74d5dafe6ea0ea28ef2ee/contracts/RealitioProxy.sol).

With the payouts, the ClosedMarket view checks if the connected user has some outcome with a non-zero payout, and shows the buttons accordingly.

Another change that was necessary was the redeem function. When we had a single winning outcome, the amount of collateral obtained was equal to the balance of shares for that outcome (because, when you have a single winning outcome, the payout is 1/1). Now the amount of collateral obtained needs to be computed, this is done [here](https://github.com/protofire/gnosis-conditional-exchange/blob/6ad50a4a06d17975273a730c71adf573ab955cc9/app/src/components/market/profile/closed_market_detail.tsx#L118-L124).

The math isn't hard: if the payouts are [3, 1], and the user has 1000 shares of the first outcome, they'll receive 3/4*1000 = 750 of collateral.

Knowing all this, this PR should be easy to review commit by commit.


app/src/components/market/profile/closed_market_detail.tsx
```diff
@@ -140,6 +152,8 @@ export const ClosedMarketDetailWrapper = (props: Props) => {
     disabledColumns.push(OutcomeTableValue.Shares)
   }
 
+  const hasWinningOutcomes = payouts && payouts.some((payout, index) => payout > 0 && balances[index].shares.gt(0))
```
cag, line 155: Couldn't this also be `earnedCollateral.gt(0)`?

app/src/services/realitio.ts
```diff
@@ -157,10 +157,15 @@ class RealitioService {
   }
 
   getWinnerOutcome = async (questionId: string): Promise<number> => {
```
cag, line 159: Is this method still used?

fvictorio: It's still being used for the "Has winning outcomes" filter... which means that that filter won't quite work with nuanced binary questions and undefined answers. I will open an issue for that.

app/src/components/market/profile/closed_market_detail.tsx
fvictorio: Done, thank you!

PR #514, opened 2020-04-07
Extra docs for building

cag:
Wrote up some docs for building the app. Hope they are helpful.


fvictorio:
Added a small change. Other than that, LGTM.

PR #523, opened 2020-04-09
Change expression of formula passed to newton-raphson to ease convergence 

### LINKED ISSUE #522: Bug when selling all shares in a market with more than two outcomes
Step to reproduce:

1. Create a market with more than two outcomes
2. Buy N shares
3. Sell N (or close to N, it happens even for 0.9*N) shares

The error seems to come from the Newton-Raphson function, which doesn't converge in this scenario.

hpmaxi:
Closes #522


app/src/util/tools.ts
```diff
@@ -133,15 +133,19 @@ export const calcSellAmountInCollateral = (
...
+    //   `a` is the amount of outcomes that are being sold
+    //   `r` (the unknown) is the amount of collateral that will be returned in exchange of `a` tokens
+    const firstTerm = otherHoldingsBig.map(h => h.minus(r).minus(r.mul(fee))).reduce((a, b) => a.mul(b))
+    const secondTerm = holdingsBig.plus(sharesToSellBig).minus(r)
+    const thirdTerm = otherHoldingsBig.reduce((a, b) => a.mul(b), holdingsBig)
+    return firstTerm.mul(secondTerm).minus(thirdTerm)
   }
 
-  const r = newtonRaphson(f, 0)
+  const r = newtonRaphson(f, sharesToSellBig, { maxIterations: 100 })
```
cag, line 148: Should use `0` instead of `sharesToSellBig`.

app/src/util/tools.spec.ts
```diff
@@ -172,6 +172,10 @@ describe('tools', () => {
         ],
         '100000000000000000',
       ],
+      [
+        ['18399816000000000000', '139733493703807763', ['11009048601975904608', '17551468438676294710']],
+        '10381992534881175324',
+      ],
```
cag, line 178: Here's an additional test case:

```js
      [
        [
          '200000',
          '100000',
          ['100000', '100000'],
          '0'
        ],
        '38197',
      ],
```

app/src/util/tools.ts
fvictorio: Oh, this was left from a test. Thanks for spotting it!

PR #536, opened 2020-04-14
Feature/ show shares column when buy or sell shares screen is shown

### LINKED ISSUE #515: Show amount of shares owned by user
In the buy and sell pages, when you select an outcome, a message saying "You own X shares" below it.

The UI for this is done with a hardcoded value, but hidden.

See also: https://github.com/protofire/gnosis-conditional-exchange/issues/467

### LINKED ISSUE #534: Show Shares column when buy or sell shares screen is shown
Currently only shown in the pooling interface: it would make sense to use the same table layout in all screens: Market overview, pooling and buy/sell.

gabitoesmiapodo:
Closes #534
Closes #515 

It was requested to show the "Shares" column on the Buy and Sell sections, but this is actually addressed by the design as seen here: https://www.figma.com/file/fDKaHCCfwqmOmOBg5xkSoo/Pages?node-id=774%3A590

It should look like these now:

Screen Shot 2020-04-14 at 17 42 04
Screen Shot 2020-04-14 at 17 43 57

### DIFF
--- a/app/src/components/market/common/outcome_table/index.tsx
+++ b/app/src/components/market/common/outcome_table/index.tsx
@@ -130,7 +130,7 @@ export const OutcomeTable = (props: Props) => {
     const { balance, index } = props
     const shares = new Big(balance.shares.toString())
 
-    if (!payouts || shares.eq(0)) return null
+    if (!payouts || !shares.eq(0)) return null
 
     const redeemable = new BigNumber(shares.mul(payouts[index]).toString())
 
@@ -181,6 +181,7 @@ export const OutcomeTable = (props: Props) => {
     const showSharesAndPriceChange = isOutcomeSelected && !IS_CORONA_VERSION
     const formattedPayout = formatBigNumber(mulBN(shares, payout), collateral.decimals)
     const isWinningOutcome = payouts && payouts[outcomeIndex] > 0
+    const formattedShares = formatBigNumber(shares, collateral.decimals)
 
     return (
       <TR key={outcomeName}>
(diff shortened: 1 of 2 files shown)


app/src/components/market/common/outcome_table/index.tsx
```diff
@@ -199,12 +200,16 @@ export const OutcomeTable = (props: Props) => {
         {disabledColumns.includes(OutcomeTableValue.OutcomeProbability) ? null : withWinningOutcome ? (
           <TDStyled textAlign={TableCellsAlign[1]}>
             <BarDiagram outcomeIndex={outcomeIndex} outcomeName={outcomeName} probability={probability} />
-            {showSharesAndPriceChange && <OwnedShares outcomeIndex={outcomeIndex} value="250.55" />}
+            {!shares.isZero() && disabledColumns.includes(OutcomeTableValue.Shares) && (
```
fvictorio, line 203: Wouldn't this be `!disabledColumns.includes(...)`?

fvictorio: Btw, if that's the case, it's probably better to move it to a `showSharesColumn` variable.

fvictorio: Oh wait, I get it now. This means that the "You own..." message is only shown when the Shares column isn't. I still think it should be extracted to something like `showOwnedSharesMessage`.

gabitoesmiapodo: Yeah, agreed.

PR #537, opened 2020-04-14
Move @types/big.js from devDeps to deps for prod build

cag:
This will allow building after a `yarn install --production` instead of regular `yarn`.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -14,6 +14,7 @@
     "@react-corekit/use-interval": "^1.1.1",
     "@realitio/realitio-lib": "^2.0.1",
     "@reduxjs/toolkit": "^1.2.3",
+    "@types/big.js": "^4.0.5",
     "@types/jest": "24.0.18",
     "@types/lodash.unionby": "^4.8.6",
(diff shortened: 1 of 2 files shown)


cag:
It fails because missing CPK types. Adding those causes an error for some reason:

```
    Types of property 'ethers' are incompatible.
      Type 'typeof import("/home/alan/src/github.com/protofire/gnosis-conditional-exchange/app/node_modules/ethers/ethers")' is not assignable to type 'typeof import("/home/alan/src/github.com/protofire/gnosis-conditional-exchange/app/node_modules/@types/contract-proxy-kit/node_modules/ethers/ethers")'.  TS2345
```

Investigating this...

cag:
Okay, figured it out. Had to run:

```sh
npx yarn-deduplicate yarn.lock
```

To actually dedup `ethers`. Then, the build will find matching `ethers` for types.

fvictorio:
Thanks! I promise I'll start banning `devDependencies` for apps.

cag:
Ah it's fine as long as it's not a build tool or dependency!

PR #539, opened 2020-04-15
Use JSONP for geojs

fvictorio:
This should fix the issue with geolocation when the user has an ad blocker.


fvictorio:
For the record: this doesn't fix the reported problem in all cases, but it does seem to work with an ad blocker enabled more often than when using the JSON endpoint, so I think it's worth merging it.

Issue #547, opened 2020-04-16
List of pending issues before release


fvictorio:
- [x] Show message to users of blacklisted countries [G]
- [x] Show a message when the geolocation can't be obtained (probably because of an ad blocker) [G]
- [x] Hide "Show pool information" button [G]
- [x] Replace "24h volume" with total volume in market details [Maxi]
- [x] Fix WalletConnect [Franco]
- [x] Show "My Shares" column in buy/sell pages [G]
- [x] Show "Answer in Realitio" button when question opens [Franco]
- [x] Disclaimer must appear each time the site is reloaded. [G]
- [x] Update Disclaimer text https://docs.google.com/document/d/141J4I1d2UYC10rssoCIvv1wc1LVgHRkYEMt4oyB3wtU/edit

PR #556, opened 2020-04-16
(Fix) Ad blocking and region restrictions

gabitoesmiapodo:
Part of #547

- If an adblocker is detected, a message advising to disable it should be displayed.
- If the user is connected from a restricted country or the country couldn't be detected, the "Connect" button should be disabled, and a tooltip explaining why should show when trying to use it.

**Related to the above items:** Part of the updated disclaimer explains that users in the US will not be able to use this dApp.

### DIFF
--- a/app/src/components/common/layout/header/index.tsx
+++ b/app/src/components/common/layout/header/index.tsx
@@ -1,6 +1,7 @@
 import React, { useEffect, useState } from 'react'
 import { withRouter } from 'react-router'
 import { NavLink, RouteComponentProps } from 'react-router-dom'
+import ReactTooltip from 'react-tooltip'
 import styled, { css } from 'styled-components'
 import { useWeb3Context } from 'web3-react/dist'
 
@@ -91,6 +92,10 @@ const ButtonDisconnectWalletStyled = styled(ButtonDisconnectWallet)`
   }
 `
 
+const ButtonWrapper = styled.div`
+  ${ButtonCSS}
+`
+
 const ContentsRight = styled.div`
   align-items: center;
   display: flex;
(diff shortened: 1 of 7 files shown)


app/src/components/common/layout/header/index.tsx
```diff
@@ -105,27 +108,74 @@ const HeaderContainer: React.FC<RouteComponentProps> = (props: RouteComponentPro
 
   const { history, ...restProps } = props
   const [isModalOpen, setModalState] = useState(false)
-
-  // hide connect button if country is blacklisted, or the info isn't available yet
-  const hideConnectButton = IS_CORONA_VERSION && (isBlacklistedCountry === null || isBlacklistedCountry === true)
+  const [isAdBlockDetected, setIsAdBlockDetected] = useState(false)
+
+  useEffect(() => {
```
fvictorio, line 113: Let's move this to a custom hook, so that we can do:

```
const isAdBlockDetected = useDetectAdblock()
````

No need to make a new module for it since it will only be used in this file.

**Comment by fvictorio** [comment]
### Ad blocker detected

image

### Country cannot be detected

image

### Blacklisted country

image

File: app/src/components/common/layout/header/index.tsx [hunk 2]
```diff
@@ -105,27 +108,74 @@ const HeaderContainer: React.FC<RouteComponentProps> = (props: RouteComponentPro
...
+    }).then(data => {
+      if (data) {
+        setIsAdBlockDetected(true)
+      }
+    })
+  }, [])
+
+  const disableConnectButton = IS_CORONA_VERSION && (isBlacklistedCountry === null || isBlacklistedCountry === true)
+  const blacklistText = isBlacklistedCountry === true ? 'This action is not allowed in your country' : ''
+  const geolocationErrorText =
```
fvictorio, line 136: I would replace these two lines with:

```
  const tooltipText =
    isBlacklistedCountry === null
      ? "We couldn't detect your country. Maybe an ad blocker is enabled?"
      : isBlacklistedCountry === true
      ? 'This action is not allowed in your country'
      : ''
```

PR #573, opened 2020-04-20
Avoid Set Allowance flashing

### LINKED ISSUE #571: "Set allowance" is briefly shown when Buy page is opened
I'm pretty sure this is caused by us not differentiating between "not loaded" and "false". Maybe the RemoteData type can fix this.

fvictorio:
Closes #571.

### DIFF
--- a/app/src/components/market/sections/market_buy/market_buy.tsx
+++ b/app/src/components/market/sections/market_buy/market_buy.tsx
@@ -13,8 +13,9 @@ import { useCpk } from '../../../../hooks/useCpk'
 import { useCpkAllowance } from '../../../../hooks/useCpkAllowance'
 import { MarketMakerService } from '../../../../services'
 import { getLogger } from '../../../../util/logger'
+import { RemoteData } from '../../../../util/remote_data'
 import { computeBalanceAfterTrade, formatBigNumber } from '../../../../util/tools'
-import { MarketMakerData, OutcomeTableValue, Status } from '../../../../util/types'
+import { MarketMakerData, OutcomeTableValue, Status, Ternary } from '../../../../util/types'
 import { Button, ButtonContainer } from '../../../button'
 import { ButtonType } from '../../../button/button_styling_types'
 import { BigNumberInput, TextfieldCustomPlaceholder } from '../../../common'
@@ -64,7 +65,8 @@ const MarketBuyWrapper: React.FC<Props> = (props: Props) => {
   const [allowanceFinished, setAllowanceFinished] = useState(false)
   const { allowance, unlock } = useCpkAllowance(signer, collateral.address)
 
-  const hasEnoughAllowance = allowance && allowance.gte(amount)
+  const hasEnoughAllowance = RemoteData.mapToTernary(allowance, allowance => allowance.gte(amount))
+  const hasZeroAllowance = RemoteData.mapToTernary(allowance, allowance => allowance.isZero())
 
   // get the amount of shares that will be traded and the estimated prices after trade
   const calcBuyAmount = useMemo(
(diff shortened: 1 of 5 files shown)


app/src/util/types.ts
```diff
@@ -159,3 +159,9 @@ export interface MarketMakerData {
   userEarnings: BigNumber
   userPoolShares: BigNumber
 }
+
+export enum Ternary {
```
fvictorio, line 163: Added this enum for the scenarios where we have a flag whose value can be unknown.

We were using `boolean | null`, but I find it less readable and more error-prone. That being said, using this type has the risk of doing something like `if (flag)` that will always be true. This can be fixed with a linter rule, but it's very slow, so maybe we should run that rule only in the CI.

app/src/util/remote_data.ts
```diff
@@ -30,12 +32,13 @@ const isLoading = (rd: RemoteData<any>): rd is Loading => rd._type === Types.loa
...
 const isFailure = (rd: RemoteData<any>): rd is Failure => rd._type === Types.failure
 const isReloading = (rd: RemoteData<any>): rd is Reloading<any> => rd._type === Types.reloading
+const isAsking = (rd: RemoteData<any>): rd is Loading | Reloading<any> => isLoading(rd) || isReloading(rd)
 
 function hasData<T>(rd: RemoteData<T>): rd is Success<T> | Reloading<T> {
   return isSuccess(rd) || isReloading(rd)
 }
 function getDataOr<T>(rd: RemoteData<T>, defaultValue: T): T {
-  return isSuccess(rd) ? rd.data : defaultValue
+  return hasData(rd) ? rd.data : defaultValue
```
fvictorio, line 41: Not related to this PR, just noticed this and made it more robust.

app/src/hooks/useCpkAllowance.tsx
```diff
@@ -15,7 +16,7 @@ import { useCpk } from './useCpk'
  */
 export const useCpkAllowance = (signer: Signer, tokenAddress: string) => {
   const cpk = useCpk()
-  const [allowance, setAllowance] = useState<Maybe<BigNumber>>(null)
+  const [allowance, setAllowance] = useState<RemoteData<BigNumber>>(RemoteData.notAsked())
```
fvictorio, line 19: Converted allowance to a `RemoteData` to have better information about its state.

app/src/components/market/sections/market_buy/market_buy.tsx
```diff
@@ -164,8 +166,8 @@ const MarketBuyWrapper: React.FC<Props> = (props: Props) => {
   const sharesTotal = formatBigNumber(tradedShares, collateral.decimals)
   const total = `${sharesTotal} Shares`
 
-  const hasZeroAllowance = allowance && allowance.isZero()
-  const showSetAllowance = allowanceFinished || hasZeroAllowance || !hasEnoughAllowance
+  const showSetAllowance =
+    allowanceFinished || hasZeroAllowance === Ternary.True || hasEnoughAllowance === Ternary.False
```
fvictorio, line 170: Set allowance is shown if:

- We are sure that the user has zero allowance
- We are sure that the user doesn't have enough allowance

Same comment aplies in the pooling page.

app/src/components/market/sections/market_buy/market_buy.tsx
```diff
@@ -152,7 +154,7 @@ const MarketBuyWrapper: React.FC<Props> = (props: Props) => {
     cost.isZero() ||
     isBuyAmountGreaterThanBalance ||
     amount.isZero() ||
-    !hasEnoughAllowance
+    hasEnoughAllowance !== Ternary.True
```
fvictorio, line 157: The button will be disabled if the user doesn't have enough allowance, or if we don't know how much they have.

PR #626, opened 2020-04-29
Error messages in pool liquidity, buy and sell

### LINKED ISSUE #611: Show an error message when the entered value is greater than the balance
We disable the button (buy, sell, fund) when the amount entered is bigger than the balance, but there's nothing explaining why.

Implementation detail: maybe this issue is a good excuse to merge the Textfield and TextfieldWithSymbol components.

hpmaxi:
Closes #611 

I'm not sure about the error messages.


gabitoesmiapodo:
Made some changes, it should look like these now

Screen Shot 2020-04-30 at 17 45 15
Screen Shot 2020-04-30 at 17 42 59
Screen Shot 2020-04-30 at 17 43 13
Screen Shot 2020-04-30 at 17 45 27
Screen Shot 2020-04-30 at 17 43 30
Screen Shot 2020-04-30 at 17 43 50
Screen Shot 2020-04-30 at 17 44 07
Screen Shot 2020-04-30 at 17 44 29

fvictorio:
Not sure I like this behavior:

Peek 2020-05-03 19-26

If I enter "28", the number is not shown but the error is. This is not very standard behavior IMO. Also, if I then delete the "2", the error is still shown. I'd say we just allow any input, and show the error when the amount is bigger than the balance.

app/src/components/common/form/big_number_input/index.tsx
```diff
@@ -35,20 +15,50 @@ const Input = styled.input`
...
+  valueFixedDecimals?: number
+}
+
+type Props = OverrideProperties<
+  PropsBigNumber,
+  {
+    max?: BigNumber
+    min?: BigNumber
+    onChange: (value: BigNumberInputReturn) => void
+    onError?: (e: BigNumberInputError) => void
```
fvictorio, line 42: This is overkill. The min/max properties are there mainly for disabling certain numbers, and for the arrows to behave in a sensible way. But error handling should be the parent's component responsibility.

gabitoesmiapodo: Errors are handled outside BigNumberInput, but I think BigNumberInput should inform there are errors to handle (?) so I added some code to do just that.

But yeah, it currently disables user input when there's an out of bounds error, and that's why https://github.com/protofire/gnosis-conditional-exchange/pull/626#issuecomment-623193010 happens.

That behavior was already in place, as a general "out of bounds" error. If we remove that, this is what will happen, and we should fix all of those problems individually:

May-04-2020 17-29-43

PR #629, opened 2020-05-04
(Fix) / add spinner

### LINKED ISSUE #615: Add spinner
There seems not to be any indicator / spinner working when using the categories / filters / search / etc.

Check if it's working, and fix if necessary.

gabitoesmiapodo:
Closes #615

There should be a spinner when changing categories / filters / etc. (nothing was shown before).

### DIFF
--- a/app/src/components/market/sections/market_list/market_home_container.tsx
+++ b/app/src/components/market/sections/market_list/market_home_container.tsx
@@ -32,15 +32,15 @@ const wrangleResponse = (data: GraphMarketMakerDataItem[], networkId: number): M
 
     return {
       address: graphMarketMakerDataItem.id,
-      collateralVolume: bigNumberify(graphMarketMakerDataItem.collateralVolume),
-      collateralToken: graphMarketMakerDataItem.collateralToken,
-      outcomeTokenAmounts: graphMarketMakerDataItem.outcomeTokenAmounts.map(bigNumberify),
-      title: graphMarketMakerDataItem.title,
-      outcomes,
-      openingTimestamp: new Date(1000 * +graphMarketMakerDataItem.openingTimestamp),
       arbitrator: graphMarketMakerDataItem.arbitrator,
       category: graphMarketMakerDataItem.category,
+      collateralToken: graphMarketMakerDataItem.collateralToken,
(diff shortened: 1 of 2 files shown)


app/src/components/market/sections/market_list/market_home_container.tsx
```diff
@@ -106,32 +107,43 @@ const MarketHomeContainer: React.FC = () => {
...
+      setIsFiltering(false)
+    } else if (error) {
+      setMarkets(RemoteData.failure(error))
+      setIsFiltering(false)
     }
-  }, [fetchedMarkets, loading, error, context.networkId])
+  }, [fetchedMarkets, loading, error, context.networkId, markets])
 
   const onFilterChange = useCallback((filter: any) => {
-    setMoreMarkets(true)
```
fvictorio, line 122: Why is this removed?

app/src/components/market/sections/market_list/market_home_container.tsx
```diff
@@ -106,32 +107,43 @@ const MarketHomeContainer: React.FC = () => {
...
     setFilter(filter)
+    setIsFiltering(true)
```
fvictorio, line 128: I wonder if, instead of adding this flag, we could set markets to `RemoteData.loading()` here. Anyway, not a blocker, but it seems like `isFiltering` is kind of redundant.

app/src/components/market/sections/market_list/market_home_container.tsx
gabitoesmiapodo: It was showing the button when it was not needed IIRC.

Generally speaking, my intention was to not to display the button unless there are more than 10 items. The way it was implemented before it was always displayed, and hidden when trying to load more elements and finding there were not more than 10, or no more items to load.

app/src/components/market/sections/market_list/market_home_container.tsx
gabitoesmiapodo: Well, I didn't know we could do that.

Tried to make the UI display a spinner every time there was an ongoing `RemoteData.loading()` but it seemed to trigger only when loading data / more data, not when filtering it.

PR #631, opened 2020-05-05
Configure app based on version passed

### LINKED ISSUE #622: Fork: Only markets traded in GNO tokens
- [ ] Remove market creation
- [ ] Only allow GNO on the list
- [ ] Change name and logo: ConditionalGNO
- [ ] Document all ENV variable for this version, Omen and Corona

hpmaxi:
Closes #622 

This changes only `IS_CORONA_VERSION` for `VERSION`, `VERSION` is used in `constants.ts` file to configure the app. As talked with @fvictorio this could be improved further using some kind of pattern matching to make it clear with options are set for each version.

All occurrences of `IS_CORONA_VERSION` were removed in favor of a constant expressing what is showing or hiding in the design. Also, there are three maps for default arbitrator, default token, and logo to be used in each case. Only the logo change was specified on the issue but I imagine that this version could use a special arbitrator and by sure the GNO token by default.

`SHOW_FOOTER` could be removed as env variable and be expressed as a function of the VERSION used.


app/src/common/constants.ts
```diff
@@ -38,8 +56,42 @@ export const GRAPH_RINKEBY_HTTP =
...
   process.env.REACT_APP_GRAPH_RINKEBY_WS || 'wss://api.thegraph.com/subgraphs/name/gnosis/omen-rinkeby'
 
-// Corona version options
-export const IS_CORONA_VERSION = process.env.REACT_APP_IS_CORONA_VERSION === 'true'
+export const USE_3BOX = !IS_CORONA_VERSION
+export const USE_DISQUS = IS_CORONA_VERSION
+
+const LOGO_MAP: { [versionName: string]: any } = {
+  corona: CoronaMarketsLogo,
+  gno: OmenLogo,
```
hpmaxi, line 64: Use GNO logo

app/src/components/market/sections/market_list/market_home_container.tsx
```diff
@@ -66,7 +71,8 @@ const MarketHomeContainer: React.FC = () => {
   const { account, library: provider } = context
   const feeBN = ethers.utils.parseEther('' + MARKET_FEE / Math.pow(10, 2))
   const query = buildQueryMarkets({
-    isCoronaVersion: IS_CORONA_VERSION,
+    whitelistedCreators: WHITELISTED_CREATORS,
+    whitelistedTemplateIds: WHITELISTED_TEMPLATE_IDS,
```
hpmaxi, line 75: I want to refactor this when we have more details about the gno version

app/src/components/market/sections/market_profile/market_status/open.tsx
```diff
@@ -132,11 +132,11 @@ const Wrapper = (props: Props) => {
...
       <MarketTopDetails marketMakerData={marketMakerData} toggleTitle="Pool Information" />
       {renderTableData()}
-      {isQuestionOpen && IS_CORONA_VERSION && openQuestionMessage}
+      {isQuestionOpen && SHOW_ANSWER_IN_REALITIO && openQuestionMessage}
       <WhenConnected>
         <ButtonContainer>
-          {!IS_CORONA_VERSION && poolButton}
-          {isQuestionOpen ? (IS_CORONA_VERSION ? null : openInRealitioButton) : buySellButtons}
+          {SHOW_POOLING_BTN && poolButton}
+          {isQuestionOpen ? (SHOW_ANSWER_IN_REALITIO ? null : openInRealitioButton) : buySellButtons}
```
hpmaxi, line 139: Maybe we need 3 variables here, and more expressive names, but I wasn't sure about why these buttons were hidden.

app/src/common/constants.ts
fvictorio: Uh?

app/src/common/constants.ts
```diff
@@ -38,8 +56,42 @@ export const GRAPH_RINKEBY_HTTP =
 export const GRAPH_RINKEBY_WS =
   process.env.REACT_APP_GRAPH_RINKEBY_WS || 'wss://api.thegraph.com/subgraphs/name/gnosis/omen-rinkeby'
 
-// Corona version options
-export const IS_CORONA_VERSION = process.env.REACT_APP_IS_CORONA_VERSION === 'true'
+export const USE_3BOX = !IS_CORONA_VERSION
+export const USE_DISQUS = IS_CORONA_VERSION
```
fvictorio, line 60: Since we won't have both comments at the same time, maybe we should have only one variable here, like `USE_DISQUS`, and use 3box when it's false.

app/src/common/constants.ts
```diff
@@ -1,3 +1,21 @@
+import { CoronaMarketsLogo } from '../components/common/logos/corona_markets'
+import { OmenLogo } from '../components/common/logos/omen'
+
+enum Version {
+  GNO = 'gno',
+  CORONA = 'corona',
+  OMEN = 'omen',
+}
+
+export const VERSION = process.env.REACT_APP_VERSION || 'omen'
```
fvictorio, line 10: Check that the given `REACT_APP_VERSION` is one of the allowed values, and make the type of `VERSION` be `Version`.

app/src/common/constants.ts
```diff
@@ -38,8 +56,42 @@ export const GRAPH_RINKEBY_HTTP =
 export const GRAPH_RINKEBY_WS =
   process.env.REACT_APP_GRAPH_RINKEBY_WS || 'wss://api.thegraph.com/subgraphs/name/gnosis/omen-rinkeby'
 
-// Corona version options
-export const IS_CORONA_VERSION = process.env.REACT_APP_IS_CORONA_VERSION === 'true'
+export const USE_3BOX = !IS_CORONA_VERSION
+export const USE_DISQUS = IS_CORONA_VERSION
+
+const LOGO_MAP: { [versionName: string]: any } = {
```
fvictorio, line 62:
```suggestion
const LOGO_MAP: { [K in Version]: any } = {
```

fvictorio:
Also some tests are failing.

app/src/common/constants.ts
hpmaxi: We don't have a logo yet, so I used `OmenLogo` until we have one.

PR #650, opened 2020-05-08
Check for zero supply

### LINKED ISSUE #647: Markets without funding aren't loading
For example, https://conditional.netlify.app/#/0x06abd8f44abab8781c50dc05d02bb07cdd9f08bc

In this particular case, the problem comes when `feesWithdrawableBy` is called. But there might be other issues.

hpmaxi:
Closes #647


leolower:
@hpmaxi the issue remains if you click "Pool liquidity" on the market details. See https://deploy-preview-650--conditional.netlify.app/#/0x06abd8f44abab8781c50dc05d02bb07cdd9f08bc/pool-liquidity

PR #655, opened 2020-05-11
[GNO] Filter by currency instead of whitelisted creators

hpmaxi:


fvictorio:
Looking into something I ended up adding `apollo-link-logger`. It's not exactly related to this PR, but given I've already done it, we can just merge all together.

PR #657, opened 2020-05-12
History chart

### LINKED ISSUE #623: Graph showing the history of Odds over time
A graph showing the odds over time

### LINKED ISSUE #663: Loading of Chart requires an indicator
Hey, I just played around with  https://deploy-preview-657--conditional.netlify.app/ and it seems like the loading of the chart requires a little bit of time. To not have the user feel like it is lagging we should include a loading indicator.

[Playground](https://www.figma.com/proto/GNoehJ64fqfptghsjYqvTJ/Playground-V2?node-id=2291%3A1&viewport=252%2C517%2C0.21183109283447266&scaling=min-zoom)
[Design Specification](https://www.figma.com/file/GNoehJ64fqfptghsjYqvTJ/Playground-V2?node-id=2291%3A1)

hpmaxi:
Closes #623 
Closes #663


app/src/components/market/common/history-chart/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+  useEffect(() => {
+    const getBlocks = async (latestBlockNumber: number) => {
+      const blocksPerDay = 5760
+      const totalDataPoints = 7
+      const granularity = 7
+
+      if (latestBlockNumber) {
+        const blockNumbers = Array.from(new Array(totalDataPoints), (_, i) => i * granularity).map(
+          multiplier => latestBlockNumber - multiplier * blocksPerDay,
+        )
```
hpmaxi, line 91: @fvictorio I changed this, to have only the last week we should change `granularity = 1`, because I wanted to get a chart not so horizontal.

gabitoesmiapodo:
Made some changes to the styling.

It was looking like these:

Screen Shot 2020-05-13 at 16 50 53
Screen Shot 2020-05-13 at 16 51 04
Screen Shot 2020-05-13 at 16 51 12
Screen Shot 2020-05-13 at 16 52 44

Should look like these now:

Screen Shot 2020-05-13 at 18 18 11
Screen Shot 2020-05-13 at 18 17 58
Screen Shot 2020-05-13 at 18 18 16

pimato:
Going to check it out and give feedback @gabitoesmiapodo .

app/src/components/market/common/history_chart/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+    if (latestBlockNumber) {
+      getBlocks(latestBlockNumber)
+    }
+  }, [latestBlockNumber, library, period])
+  return hidden ? null : (
+    <HistoryChart
+      holdingSeries={holdingsSeries}
+      onChange={setPeriod}
+      options={Object.keys(mapPeriod)}
```
fvictorio, line 114: Using a strongly typed `Object.keys` would mean that the type of this prop would be `Period` and we could remove the type assertion in the `onChange` inside the component. I think the function can be implemented like this:

```
function keys<T>(o: T): Array<keyof T> {
  return Object.keys(o) as any
}
```

pimato:
can we have the percentage indicator/bar/y on the right side? @hpmaxi

hpmaxi:
> can we have the percentage indicator/bar/y on the right side? @hpmaxi

Done!

PR #680, opened 2020-05-14
Show omen title and favicon by default

### LINKED ISSUE #662: REACT_APP_VERSION=omen doesn`t show omen favicon and page title correctly
while Loading I see:
Screenshot 2020-05-13 at 16 50 05
If I click on the tab it updates it too:
Screenshot 2020-05-13 at 16 51 02

even with:
Screenshot 2020-05-13 at 16 48 13

fvictorio:
Closes #662.

Notice that the same problem will happen in gnomen and corona (it will show omen's title and favicon until the version is recognized). I don't know if this can be make to work perfectly for all versions, since this is detected on runtime. A hacky solution would be to `sed` the index.html file before building, but that's outside the scope of this issue.


pimato:
Yeah I think at some point the omen project should get rid of the other versions so we will not have that kind of issue.

PR #683, opened 2020-05-15
(Fix) Dropdown buttons should be clickable

### LINKED ISSUE #675: Dropdown buttons should be clickable
Currently:
Screenshot 2020-05-13 at 21 29 12

If I click on More button again, nothing happens. I want to see the button go back to idle clickable mode.

gabitoesmiapodo:
Closes #675


pimato:
Just checked! Works!

PR #712, opened 2020-05-21
Simplify errors in amount inputs

### LINKED ISSUE #673: Inputfield error doesn`t make any sense
this message appears when I am trying to put any number into the inputfield:

Screenshot 2020-05-13 at 21 15 02

If the error wants to imply that the user has no funds we should say **Insufficient balance**

fvictorio:
Closes #673.

Removed all error handling from BigNumberInput. Also the errors are just handled in each component in a not-very-DRY way. This isn't a great approach to form validation, but I think it's enough since we just have one input in each scenario.

@pimato can you check how the inputs behave in this deploy preview?


fvictorio:
I had to `Maybe`fy some values, otherwise the "Insufficiente balance" was showing for half a second when you opened the buy page, and we need to tell apart between "zero" and "unknown".

PR #724, opened 2020-05-22
(Fix) Impossible to add more than 4 outcomes

### LINKED ISSUE #648: Impossible to add more than 4 outcomes
The UI does not allow to add more than 4 outcomes. The "Add new outcome" fields becomes greyed out an unfillable after 4 outcomes are added.
image

gabitoesmiapodo:
Closes #648

- Outcomes amount is environment variable (as it used to be), default is 8 if nothing's set.
- Outcomes amount is properly used now (it used to be amount - 1).
- Max outcomes amount is 256, any number above that is ignored.
- Minimum outcomes amount is 2.
- There are 32 hardcoded outcomes' colors (which seem plenty to me). If more than 32 outcomes are created, the colors array loops from the beginning (the same for 65 outcomes, 97 outcomes, etc.). I don't think anything above - say- 16 or 20 outcomes will look pretty or be easy to remember visually, and charts will look extremely crowded, but the possibility is there.


app/src/theme/utils.ts
```diff
@@ -0,0 +1,22 @@
...
+    logger.error('Error when using theme or theme.outcomes')
+    return defaultReturn
+  }
+
+  const colors = localTheme.outcomes.colors
+  const maxIndex = colors.length - 1
+  const groupIndex = Math.trunc(index / colors.length)
+  const outcomeIndex: number = index - groupIndex - maxIndex * groupIndex
+
+  return colors[outcomeIndex]
```
fvictorio, line 21: Can't you just do `colors[index % colors.length]`?

gabitoesmiapodo: Don't know, that's the best I could come up with.

gabitoesmiapodo: There's also a problem with `colors.length` not being the same as the array's top index (not sure if your comment addresses that or not).

hpmaxi:
Are we using this for chart colors?

gabitoesmiapodo:
@hpmaxi 

> Are we using this for chart colors?

We should be (I think there's some mapping there?), but I'm not sure if I missed something.

PR #726, opened 2020-05-22
hotfix for disabling add if 100 percent is reached

hpmaxi:
For uniform probabilities, there was a need to maintain enabled the add button until each outcome has a probability of 100/MAX_OUTCOMES, but for manual probabilities the same button should be disabled once `totalProbabilities` reach 100% to avoid getting negative values at using `setMax`


app/src/components/market/sections/market_create/steps/outcomes/index.tsx
```diff
@@ -223,7 +224,7 @@ const Outcomes = (props: Props) => {
       </TitleWrapper>
       <NewOutcome uniformProbabilities={uniformProbabilities}>
         <Textfield
-          disabled={disableUniformProbabilities}
+          disabled={disableUniformProbabilities || (!uniformProbabilities && totalProbabilitiesReached)}
```
fvictorio, line 227:
```suggestion
          disabled={disableUniformProbabilities || totalProbabilitiesReached}
```

fvictorio:
@gabitoesmiapodo can you check now?

PR #740, opened 2020-05-29
Sort by liquidity and 24h volume

### LINKED ISSUE #729: Sort market list by liquidity (number of pool tokens)
@cag should add this to the subgraph

Should only be available when you filter by collateral.

### LINKED ISSUE #730: Rename sort by "Opening date" to "Resolution date" and invert results
The closest resolution date should be on top and this should be the default sort option for corona markets (not for Omen)

### LINKED ISSUE #738: Sort by 24 hour volume
This should be the default sort option.

fvictorio:
Closes #729.
Closes #730.
Closes #738.

### DIFF
--- a/app/src/components/market/sections/market_list/market_home.tsx
+++ b/app/src/components/market/sections/market_list/market_home.tsx
@@ -142,6 +142,7 @@ export const MarketHome: React.FC<Props> = (props: Props) => {
   const [category, setCategory] = useState('All')
   const [title, setTitle] = useState('')
   const [sortBy, setSortBy] = useState<Maybe<MarketsSortCriteria>>(props.currentFilter.sortBy)
+  const [sortByDirection, setSortByDirection] = useState<'asc' | 'desc'>(props.currentFilter.sortByDirection)
   const [showSearch, setShowSearch] = useState<boolean>(false)
   const [showAdvancedFilters, setShowAdvancedFilters] = useState<boolean>(false)
   const [arbitrator, setArbitrator] = useState<Maybe<string>>(null)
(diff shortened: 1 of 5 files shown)


pimato:
@fvictorio @hpmaxi what is the general default sorting?

pimato:
Also: 
Screenshot 2020-05-29 at 20 07 44

a market which is closed a month ago should not be shown as the top market sorted by resolution date

fvictorio:
Resolution date is the default only for Corona. For non-corona it's collateral volume as before.

With respect to the closed market, that's because that market was't resolved in Realitio. When it is, it will be shown in closed markets. If it shouldn't show any "ended but not resolved" market, we should open an issue to track that.

pimato:
Based on https://github.com/protofire/gnosis-conditional-exchange/issues/729
the filter option to sort by liquidity should only be available if a collateral token is set as a filter but we still show the option even if all collateral tokens are active:

Screenshot 2020-05-29 at 20 16 15

pimato:
> Resolution date is the default only for Corona. For non-corona it's collateral volume as before.

based on https://github.com/protofire/gnosis-conditional-exchange/issues/738 the default sort for corona should be 24h Volume.

pimato:
> With respect to the closed market, that's because that market was't resolved in Realitio. When it is, it will be shown in closed markets (actually first in pending). If it shouldn't show any "ended but not resolved" market, we should open an issue to track that.

It should def. not be shown in the `open` tab but rather in the `pending` tab.

fvictorio:
Can you open an issue for that?

fvictorio:
Opened #743 to keep track of that.

PR #746, opened 2020-06-01
(Feature) Add Omen footer

### LINKED ISSUE #685: Add Omen footer with version and contract hyperlink similar to Mesa
It is important in terms of security to have links to the codebase and the contracts like we have at Mesa:

Screenshot 2020-05-15 at 15 38 34

gabitoesmiapodo:
Closes #685

Still needs some logic, I think @fvictorio will work on that.


fvictorio:
Done.

PR #758, opened 2020-06-05
(fix) threebox inconsistencies

### LINKED ISSUE #755: 3box comments box UI broken again
Hey guys,

on Omen:
Screenshot 2020-06-04 at 18 32 48

on 3box testing page:
Screenshot 2020-06-04 at 18 34 54

gabitoesmiapodo:
Closes #755 

OK, I removed most of the old custom / inherited code for the 3box styles. Left only a minimum amount so it doesn't break the layout, fixed some overflowing, etc.

But it should look mostly the same than the original now.


pimato:
Hey @gabitoesmiapodo my goal was not to remove all our custom styling but just fix the issue we had. I like the current custom styling more then the default one. The only thing which should be fix is the vertical alignment in the comment textarea:

Screenshot 2020-06-06 at 18 21 34

gabitoesmiapodo:
@pimato OK, just note that there's some probability that things will break every now and then without we changing anything (as it keeps happening, it seems).

All of 3box styles come with the module and we need to override them on a separate file. Any change they make, or any module update could have some consequences, need fixing, etc.

pimato:
> @pimato OK, just note that there's some probability that things will break every now and then even if we don't change anything (as it keeps happening, it seems).
> 
> All of 3box styles come with the module and we need to override them on a separate file. Any change they make, or any module update could have some consequences, need fixing, etc.

ahh interesting. so the reason why the UI got broken was because they updated stuff?

gabitoesmiapodo:
@pimato `"3box-comments-react"` version seems to be fixed now on package.json, so that _could_ do it.

We had these custom styled 3box comments for the previous version of Omen (I mean for the old design), and the only way of customizing this seems to be overriding the base styles. When we updated  the implementation, many things changed and that may have had some consequences (even when we didn't change anything from our 3box customization). So I think that having custom styles is weak at best. We can't be sure If it's going to break on our side or theirs.

For example: there's an issue asking to update the module to the latest version (https://github.com/protofire/gnosis-conditional-exchange/issues/435), and I don't know what that will do to the styles (maybe they have new components, new styles, new classes...).

I think our best bet would be to keep customization at a minimum: basically to fix things that break something on Omen (like the overflowing issue I fixed), or that look terribly bad by default. Total customization could end up being very demanding, and issues will appear every now and then.

pimato:
@gabitoesmiapodo if you resize the avatar circle for each address to the same size as we previously had I am totally fine with removing all custom changes.

gabitoesmiapodo:
@pimato gotcha

PR #759, opened 2020-06-05
Remove duplicated owned shares

KadenZipfel:
Closes https://github.com/protofire/gnosis-conditional-exchange/issues/737


pimato:
Looks good!

fvictorio:
I think you can remove the `owned_shares` file, since it's only being used there.

Also, same comment as in #762 regarding the linter and the import order.

PR #761, opened 2020-06-05
Enable terms and conditions and disclaimer only for corona informatio…

KadenZipfel:
Closes https://github.com/protofire/gnosis-conditional-exchange/issues/741

**Note:** It's entirely possible that I'm misunderstanding what is supposed to be done here, so please consider when reviewing. What I've done is disable the terms and conditions as well as the disclaimer for all `REACT_APP_VERSION` except for `corona`.


fvictorio:
I think that's fine, but notice that instead of reading `process.env.REACT_ENV_VERSION`, we have some variables exported from `common/constants.ts` for that. For example, you can import `IS_CORONA_VERSION` from that module and then do `IS_CORONA_VERSION && <Disclaimer />`

PR #762, opened 2020-06-06
No redirect after buy

KadenZipfel:
As per https://github.com/protofire/gnosis-conditional-exchange/issues/735


app/src/components/market/sections/market_buy/market_buy.tsx
```diff
@@ -38,6 +30,14 @@ import { TransactionDetailsRow, ValueStates } from '../../common/transaction_det
 import { ViewCard } from '../../common/view_card'
 import { WalletBalance } from '../../common/wallet_balance'
 
+import { stripIndents } from 'common-tags'
```
fvictorio, line 33: Moving the imports this way makes the linter fail. Maybe your IDE isn't reading the eslint config?

app/src/components/market/sections/market_buy/market_buy.tsx
```diff
@@ -259,7 +259,7 @@ const MarketBuyWrapper: React.FC<Props> = (props: Props) => {
         </ButtonContainer>
       </ViewCard>
       <ModalTransactionResult
-        goBackToAddress={goBackToAddress}
+        goBackToAddress={undefined}
```
fvictorio, line 262: I think you can just remove the prop instead of passing `undefined`.

PR #781, opened 2020-06-08
Add webpackconfig

AugustoL:
- Use react scripts for `yarn start`, `yarn test` and `yarn eject`
- Use custom build script and webpack config (minimal modifications from react-scripts) to build.
  - Use sha256 hashFunction in webpack.
  - Removed optimizations.
- Added few more dependencies for build script.
- Use only node v12.


app/package.json
```diff
@@ -83,18 +85,17 @@
     "web3-react": "^5.0.4"
   },
   "scripts": {
-    "build": "sed -i '207s/isEnvProduction/false/' node_modules/react-scripts/config/webpack.config.js && GENERATE_SOURCEMAP=false node --max-old-space-size=4000 node_modules/react-scripts/scripts/build.js",
-    "eject": "react-scripts eject",
+    "start": "npx react-scritps start",
```
fvictorio, line 88: This has a typo.

fvictorio: Besides, does it make sense to use `react-scripts` after an eject? I thought that you are supposed to do `node scripts/whatever.js`.

AugustoL: @fvictorio where is the typo? :eyes: 

Good question! I tried! but I had to add a looooot of dependencies and I ran into some problems with the paths. So I only added the changes to make the build with scripts and custom config, and I left the rest to be done in a next PR. But yeah, we will want to move to scripts eventually :)

AugustoL: found the typo, fixed.

pimato:
@AugustoL and me just build the project and checked if we actually generate the same IPFS Hash out of the build folder. It works:

`QmeorzASHRVrJSbGS8jLQRRZ34558QDrowj8DUhyngsntx`

pimato:
@fvictorio @leolower can you guys please do code review?

PR #813, opened 2020-06-16
Only show markets that use known arbitrators

fvictorio:


pimato:
why not remove:

```  realitio: {
    name: 'Realitio Team',
    url: 'https://realit.io/',
    addresses: {
      [networkIds.MAINNET]: '0xdc0a2185031ecf89f091a39c63c2857a7d5c301a',
      [networkIds.RINKEBY]: '0x02321745bE4a141E78db6C39834396f8df00e2a0',
    },
    isSelectionEnabled: false,
  },```

and ```
  unknown: {
    name: 'Unknown',
    url: '',
    addresses: {},
    isSelectionEnabled: true,
  },```

fvictorio:
@pimato We shouldn't show markets that use the Realitio Team as arbitrator?

pimato:
> @pimato We shouldn't show markets that use the Realitio Team as arbitrator?

yeah if we can´t filter by Realit.io team markets we also should not show them!

PR #820, opened 2020-06-17
Remove non omen features

ethereal-nfts:
This PR removes features and variables related to non-omen variants of the app. It simplifies the constants and .env files while maintaining backwards compatibility with old .env files. It also removes variables and conditional statements throughout the app that are no longer required. Also removed non-omen related tests.


yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
fvictorio, line 1: This was probably created accidentally.

PR #821, opened 2020-06-18
(fix) add missing token symbol

### LINKED ISSUE #818: Token description missing in withdraw liquidity result box
in the Deposit section we should add the collateral token description:

currently:
Screenshot 2020-06-17 at 22 34 07

gabitoesmiapodo:
Closes #818 

It should look like this now:

Screen Shot 2020-06-18 at 19 07 24


pimato:
@gabitoesmiapodo the deposited should not be in state.success but rather state.normal (it should not be green)

pimato:
just added it @gabitoesmiapodo  :)

PR #835, opened 2020-06-23
remove earliest word in resolution date and formate date to UTC

pimato:
fixes https://github.com/protofire/gnosis-conditional-exchange/issues/833

@gabitoesmiapodo @fvictorio 

I´ve formated the Resolution date to UTC. It looks like the Datepicker we are using on the market creation page is not using UTC but local time. I believe we need to use UTC as the preferred timezone everywhere.


fvictorio:
@pimato sadly `react-datepicker` doesn't allow setting the timezone. So I had to do an ugly hack: the selected date is interpreted with the UTC offset. So if I select noon, the date picker is actually selecting 3 PM UTC, but then I add the offset and the resulting date is noon UTC. It's ugly, but it seems to work.

pimato:
the main issue I have is that if I pick a date and time via the picker:
Screenshot 2020-06-23 at 18 43 19

but the result looks different:
Screenshot 2020-06-23 at 18 43 24

pimato:
@fvictorio I´ve read about that on stackoverflow, seems like other people did that like that. Seems to work! I would say lets merge and I am gonna do more testing.

PR #844, opened 2020-06-24
Remove google analytics, cookies and general cleanup

pimato:


app/src/common/constants.ts
```diff
@@ -76,7 +76,5 @@ export const LINK_COOKIE_POLICY = process.env.REACT_APP_LINK_COOKIE_POLICY || fa
 export const LINK_FAQ = process.env.REACT_APP_LINK_FAQ || false
 export const DISCLAIMER_TEXT = process.env.REACT_APP_DISCLAIMER_TEXT || false
 
-export const GOOGLE_ANALYTICS_ID = process.env.REACT_APP_GOOGLE_ANALYTICS_ID || null
```
hpmaxi, line 79: You can also remove `LINK_COOKIE_POLICY`

app/src/components/common/form/date_field/index.tsx
```diff
@@ -137,7 +137,7 @@ export const DateField = (props: Props) => {
         calendarClassName="customCalendar"
         dateFormat="MMMM d, yyyy h:mm aa"
         disabled={disabled}
-        minDate={minDate}
+        minDate={new Date()}
```
hpmaxi, line 140: This is giving a warning because you're destructuring `minDate` from props and is not used afterward, maybe to maintain this prop as optional you can give a default value to it (`new Date()`) and maintain the possibility of passing a `minDate` different from now.

app/src/common/constants.ts
hpmaxi: `react-ga` dependency could be removed too.

pimato: Thank you, all removed!

app/src/components/common/form/date_field/index.tsx
pimato: added back!

PR #845, opened 2020-06-24
Add hover resolution

pimato:
adds onHover for local time and ending time:
Screenshot 2020-06-24 at 20 44 45

fixes: https://github.com/protofire/gnosis-conditional-exchange/issues/832


app/src/components/market/common/display_resolution/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+import ReactTooltip from 'react-tooltip'
+import styled from 'styled-components'
+
+import { formatDate } from '../../../../util/tools'
+
+const Wrapper = styled.div`
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  margin: 0px;
```
gabitoesmiapodo, line 12:
```suggestion
  margin: 0px;
```

Maybe use unitless zero.

PR #861, opened 2020-06-30
828 icons

### LINKED ISSUE #828: New token should automatically fetch icon
The DXdao just added USDT to its Omen token registry. It seems like the UI doesn´t know where to get the icon for USDT.
How is Omen fetching the Icons?

Screenshot 2020-06-20 at 19 16 15

namesty:
Aims to close #828.

Worth noting that `currencies_data` (https://github.com/protofire/gnosis-conditional-exchange/blob/master/app/src/components/common/icons/currencies/currencies_data.tsx), which I consider existed for icon purposes only is not being used anymore. This manually updated list also contained names for tokens additional to their symbol (`text` field)

This results in not only icons being fetched dynamically but Token texts are always their symbol now. Example: Warped Ether -> WETH. If we want to keep using names different from the symbol, we would need to keep the manually updated list in parallel or fetch it from an API.

Took 1 hour approx


pimato:
Hey @namesty I don`t see icons on the currency filter picker:
Screenshot 2020-06-30 at 11 55 01

namesty:
Fixed that. Worth noting that CDAI address fetched from ERC20 service differs from the one stored in the known tokens utils (https://github.com/protofire/gnosis-conditional-exchange/blob/master/app/src/util/networks.ts#L122). This is why CDAI's icon is not being fetched on the advanced filter but it is correctly fetched in the fund market tab

PR #866, opened 2020-07-01
Pending markets not visible

### LINKED ISSUE #743: Move markets with resolution date in the past to the "Pending" tab
See https://github.com/protofire/gnosis-conditional-exchange/pull/740#issuecomment-636115163

hpmaxi:
Closes #743


pimato:
works!

PR #877, opened 2020-07-02
Markets not loading

### LINKED ISSUE #812: Markets not loading
reproducable:

1. Go to https://conditional.netlify.com/
2. Filter only Kleros Markets
3. Switch back to all Arbitrator.
4. BUG --> market will not load anymore
Screenshot 2020-06-16 at 13 46 14

mariano-aguero:
Closes #812 

It seems like  the value for loading stuck on true when using useQuery hook.
Adding the `fetchPolicy` option on `network-only` seems to fix it. 
To take into account, network-only is used If you do not want to risk displaying any out-of-date information from the cache,
This policy favors showing the most up-to-date information over quick responses.


pimato:
this is awesome, thank you @mariano-aguero @hpmaxi ! Works now!

PR #879, opened 2020-07-02
Remove myMarkets filter if the user is not connected

### LINKED ISSUE #878: Markets tab is displayed if the user is not connected
With the account not connected, I would expect to either: Not see My Markets tab, or at least see it empty.

Screenshot_20200702_122613

mariano-aguero:
Closes #878 

If the user is not connected, the filter `myMarkets` should not be displayed

### DIFF
--- a/app/src/components/market/sections/market_list/market_home.tsx
+++ b/app/src/components/market/sections/market_list/market_home.tsx
@@ -3,6 +3,7 @@ import styled from 'styled-components'
 
 import { ConnectedWeb3Context } from '../../../../hooks/connectedWeb3'
 import { CategoryDataItem, MarketMakerDataItem } from '../../../../queries/markets_home'
+import { getLogger } from '../../../../util/logger'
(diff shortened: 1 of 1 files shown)


pimato:
@mariano-aguero I checked what happens if the user is connected and disconnects. It switches the filter to "Open" but the market list is still thinking it is the "my markets" filter page. It would be good if the market list gets updated to the open markets.

Screenshot 2020-07-03 at 15 59 05

mariano-aguero:
> @mariano-aguero I checked what happens if the user is connected and disconnects. It switches the filter to "Open" but the market list is still thinking it is the "my markets" filter page. It would be good if the market list gets updated to the open markets.
> 
> Screenshot 2020-07-03 at 15 59 05

Great, let me work on it

mariano-aguero:
@pimato I update the state, when the connection change from connected to disconnected. I think this should be enough, since updating the filter status should fetch the markets again.

pimato:
works!

PR #880, opened 2020-07-02
Insufficient Balance should only be visible if the user has balance lower than his input

### LINKED ISSUE #734: Insufficient Balance should only be visible if the user has balance lower then his input
Right now, the user gets immediately confronted with error messages when the user clicks on buy/sell. 

Currently:
Screenshot 2020-05-27 at 11 10 43

This message should only be visible if the user has balance lower then his input.

mariano-aguero:
Closes #734 

Added some improvements how the message errors are validated 

I uploaded some images of how it looks

With 0 balance and some amount:
Screenshot_20200702_141140

With 0 balance and 0 amount:
Screenshot_20200702_141132

With some balance, and the amount is greater than the balance:
Screenshot_20200702_141028

With some balance and 0 amount:
Screenshot_20200702_140955


pimato:
this is awesome and works like a charm!

PR #900, opened 2020-07-08
Dust amount remains after user sold/bought outcome tokens

### LINKED ISSUE #886: Dust amount remains after user sold/bought outcome tokens
Transaction:
https://etherscan.io/tx/0x5ee238099f984c0349950519b3b4667f83f6af5b1daff44bd247c5f49fb11342

screenshot:
Screenshot (8)

if the user decides to sell all its shares/buy with all assets, there should be no dust!

### LINKED ISSUE #901: Insufficient Balance should only be visible if the user has balance lower than his input for Market funding
same fix as: https://github.com/protofire/omen-exchange/pull/880 but also applied to the fund market view

Screenshot 2020-07-08 at 14 18 45

mariano-aguero:
Closes #886 
Closes #901 

A piece of code that was used to solve a precision problem was removed.
Something to keep into account, if the user enter some amount without all the decimals, can incur in some dust in the shares. 
Perhaps it is possible to alert the user about any dust that remains in the market, especially when the amount entered is done manually and not by clicking on the max link.

test

### DIFF
--- a/app/src/components/market/sections/market_create/steps/items/funding_and_fee_step.tsx
+++ b/app/src/components/market/sections/market_create/steps/items/funding_and_fee_step.tsx
@@ -160,7 +160,7 @@ const FundingAndFeeStep = (props: Props) => {
   const amountError =
     maybeCollateralBalance === null
       ? null
-      : maybeCollateralBalance.isZero()
+      : maybeCollateralBalance.isZero() && funding.gt(maybeCollateralBalance)
       ? `Insufficient balance`
       : funding.gt(maybeCollateralBalance)
       ? `Value must be less than or equal to ${collateralBalanceFormatted} ${collateral.symbol}`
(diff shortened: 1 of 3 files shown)


app/src/components/market/sections/market_create/steps/items/funding_and_fee_step.tsx
```diff
@@ -160,7 +160,7 @@ const FundingAndFeeStep = (props: Props) => {
   const amountError =
     maybeCollateralBalance === null
       ? null
-      : maybeCollateralBalance.isZero()
```
mariano-aguero, line 163: Fixs https://github.com/protofire/omen-exchange/issues/901

pimato:
already closes: https://github.com/protofire/omen-exchange/issues/901

Issue #871, opened 2020-07-02
Make date easy to understand


clesaege:
![image](https://user-images.githubusercontent.com/8873352/86312842-26d33d80-bc13-11ea-9f3c-8bdeae088220.png)

Now we get the hour which is good, but:
- UTC would be better than GMT (GMT will almost always mean the same thing, but it's better to display an international time instead of a local timzeone name).
- the MM/DD/YY format is exotic (only used in the US and a small number of African countries https://en.wikipedia.org/wiki/Date_format_by_country). I'd go with YYYY-MM-DD which is the ISO recommendation or use letters to write the month (actually the later may be better as markets write months in letter).

It's currently possible to understand the format due to the over stating how long it will take to reach the date. So it's not ambiguous for the user looking deep. But we can still have better UX.

PR #904, opened 2020-07-08
Make date easier to understand

KadenZipfel:
From what I've gathered, this should be the most obvious way to interpret the date. YYYY-MM-DD and full text on hover, e.g. July 12th, 2020. The only thing I'm uncertain about is the amount of space between, 'Resolution Date' and the date itself.

Closes https://github.com/protofire/omen-exchange/issues/871


app/src/components/market/common/market_top_details_open/index.tsx
```diff
@@ -56,6 +56,8 @@ const MarketTopDetailsOpen: React.FC<Props> = (props: Props) => {
     setExtraInformation(false)
   }
 
+  console.log(question.resolution)
```
pimato, line 59: shouuld be removed

Issue #804, opened 2020-06-12
Add possibility to create a custom category on market creation


pimato:
with the [new category and market state picker ](https://github.com/protofire/gnosis-conditional-exchange/issues/801) we need to enable the creation of custom categories on the market creation process. Currently only 7 categories are enabled on Omen. We would like to remove a hardcap on categories.

**Current:**
<img width="834" alt="current-category-picker" src="https://user-images.githubusercontent.com/5337809/84508049-06722c00-acc2-11ea-9113-a36de88de3f2.png">

**Version 2:**
<img width="834" alt="new-category-picker" src="https://user-images.githubusercontent.com/5337809/84507958-ec384e00-acc1-11ea-890d-8789b048173b.png">

**Deliverables:**
- Add Inputfield into the the collapsed category picker 
- inputfield placeholder should be `Add Category...`
- Once user defined a new Category the inputfield should go into the select state

**Figma:**
[Design Specification](https://www.figma.com/file/R056tSYMjvfwaFidEBItYy/Widgets-V2?node-id=1%3A953)

**Estimated Time:**
5h

**Total Bounty:**
200$ (5h * 40$ per hour)
300 REP (0.024%) DXdao reputation

**ATTENTION: The estimated time and bounty is just my personal estimations and are not a officially ratified by the collective. Feel free propose a different estimate/bounty**

nickplace:
Hello! I can take this and then take a look at issue #801 . The bounty looks fair to me as well. Cheers!

PR #920, opened 2020-07-10
Add custom category input which collapses categories onBlur

nickplace:
Implements #804


leolower:
The new categories appear on the filters and in the market details but they are missing in the market creation form.

image

pimato:
> The new categories appear on the filters and in the market details but they are missing in the market creation form.
> 
> image

yup, this should be done afterwards. I think this feature should be included in Milestone 0.1.2 nevertheless! Which would also include the changes to actually show all categories in the picker!

pimato:
@hpmaxi could you please do code review here?

app/src/components/market/sections/market_list/market_home.tsx
```diff
@@ -198,6 +198,9 @@ export const MarketHome: React.FC<Props> = (props: Props) => {
   } = props
   const [state, setState] = useState<MarketStates>(MarketStates.open)
   const [category, setCategory] = useState('All')
+  const [counts, setCounts] = useState({
+    open: 0,
+  })
```
hpmaxi, line 203: Maybe this has another use in a future feature but as it is today it could be a single number instead of an object.

nickplace:
> The new categories appear on the filters and in the market details but they are missing in the market creation form.
> 
> image

Screen Shot 2020-07-17 at 10 04 43 PM

I wonder why you're not seeing the new category input? Going to merge upstream changes and clean up this PR on Saturday.

app/src/components/market/sections/market_list/market_home.tsx
nickplace: Good catch. Yeah originally added for all of the different states. Looks like we have a closed count as well now.

pimato:
Hey @nickplace , I am getting an error:

Screenshot 2020-07-22 at 08 53 22

Issue #918, opened 2020-07-10
add more sophisticated routing


pimato:
Currently, everytime the user navigates back to the main market view it will always show the default view omen.eth.link.

I believe this is bad UX as it should navigate back to the state the user configured with. With that we need to introduce a more sophisticated routing:

default market view when visting `omen.eth.link` => `omen.eth.link/#/24h-volume/` (based on the default sort criteria)
if user changed the sort criteria to `Total Volume` => `omen.eth.link/#/volume/`
if user changed the sort criteria to `Newest` => `omen.eth.link/#/newest/`
if user changed the sort criteria to `Ending soon` => `omen.eth.link/#/ending/`

**For filter functions:**
if user filters by Currency like DAI => `omen.eth.link/#/24h-volume/currency/dai`
if user filters by Currency like DAI and by arbitrator => `omen.eth.link/#/24h-volume/currency/dai?arbitrator=kleros`

KadenZipfel:
Currently working on this, and everything is looking good except for the currency route, it will be a lot easier to use the currency address instead of the name. So instead of `omen.eth.link/#/24h-volume/currency/dai` it would be `omen.eth.link/#/24h-volume/currency/0x...`.

If we really want to use the currency name we can but it would require a map of all currencies in each network, so I figured we could just do that later on.

KadenZipfel:
Also, if we use a search query for the arbitrator (`?arbitrator=kleros`), the arbitrator will not be maintained in the route history. Is this the desired effect?

pimato:
> Also, if we use a search query for the arbitrator (`?arbitrator=kleros`), the arbitrator will not be maintained in the route history. Is this the desired effect?

great point! 

we should probably add:
if user filters by arbitrator => `omen.eth.link/#/24h-volume/&arbitrator=kleros`

pimato:
> Currently working on this, and everything is looking good except for the currency route, it will be a lot easier to use the currency address instead of the name. So instead of `omen.eth.link/#/24h-volume/currency/dai` it would be `omen.eth.link/#/24h-volume/currency/0x...`.
> 
> If we really want to use the currency name we can but it would require a map of all currencies in each network, so I figured we could just do that later on.
> 
> Edit: Also we're looking at the same thing for the arbitrator.

totally fine with having the first version that way and we can think about to improve the UX. I am worried that the URL will just get too big!

KadenZipfel:
[...]

PR #934, opened 2020-07-13
Add more sophisticated routing

KadenZipfel:
Closes https://github.com/protofire/omen-exchange/issues/918


pimato:
works perfect <3!

The only small thing is: Do we need the `&` at the end here: 
`/#/ending/category/Coronavirus?tag=Apple&`
?

pimato:
@hpmaxi could you please do code review here?

KadenZipfel:
> The only small thing is: Do we need the `&` at the end here:
> `/#/ending/category/Coronavirus?tag=Apple&`
> ?

The only work around's I can think of would likely result in bugs down the road. AFAIK this is the safest way to do it.

KadenZipfel:
Also it's worth mentioning that one gotcha with this method is that if there is an ampersand in the market state or search query, it will break. Not sure if there's a solution to this other than not using a query in the URL.

app/src/components/market/sections/market_list/market_home_container.tsx
```diff
@@ -166,7 +244,44 @@ const MarketHomeContainer: React.FC = () => {
...
+    if(filter.category && filter.category !== 'All') {
+      route += `/category/${filter.category}`
+    }
+
+    if(filter.state && filter.state !== 'OPEN') {
+      routeQuery += `state=${filter.state}&`
+    }
+
+    if(filter.title) {
+      routeQuery += `tag=${filter.title}&`
```
hpmaxi, line 280: Maybe routeQuery could be an array first with each query, and then do routeQuery.join('&').

Issue #919, opened 2020-07-10
set manual probability in configure market process should be disabled if one outcome was added with 100% probability


pimato:
Currently, if the user clicks on `set manual probability` after the user added only **one outcome** it will basically brake the form as the user can´t add a new outcome and can`t go back to the previous state:

<img width="624" alt="Screenshot 2020-07-10 at 19 44 10" src="https://user-images.githubusercontent.com/5337809/87184309-13e1fc80-c2e8-11ea-9f53-7152ead6a7f4.png">

for a quick fix, we should disable the "set manual probability" if we reached the 100% probability via the added outcomes

PR #939, opened 2020-07-15
Remove 'set manual probability' if outcomes selected

KadenZipfel:
Closes https://github.com/protofire/omen-exchange/issues/919


pimato:
instead of removing I would simple change the blueish hyperlink color to `colors.textColor` to indicate that this link is disabled.

PR #964, opened 2020-07-17
Filter by participants

hpmaxi:
I think this is working fine now if it is then I should disable filters for my markets. Maybe we need to squash these commits.


pimato:
Thanks a lot @hpmaxi for all those fixes!

-after testing it for a bit, it seems like the pagination works without any issues on "Display 4" mode.
- If I switch to Display 8 it shows me 8 markets, but the list is > 8 and it doesn´t allow me to navigate to the right anymore:
Screenshot 2020-07-18 at 11 42 00

Interestingly, if I increase from Display 8 to Display 12 and then go back to Display 8 it works:
Screenshot 2020-07-18 at 11 44 51

Issue #967, opened 2020-07-17
Users should not be able to add two of the same outcome


KadenZipfel:
![image](https://user-images.githubusercontent.com/30579067/87833610-1fc84400-c83d-11ea-9332-45b8f4fba29d.png)

If the inputted outcome is the same as an existing outcome, the '+' add outcome button should be disabled as it is if there is no input.

PR #969, opened 2020-07-17
Prevent users from adding two of the same outcome

KadenZipfel:
Closes https://github.com/protofire/omen-exchange/issues/967


pimato:
great work!
Should we also include lower case word detection here: 

Screenshot 2020-07-18 at 11 49 53

app/src/components/market/sections/market_create/steps/outcomes/index.tsx
```diff
@@ -210,6 +212,12 @@ const Outcomes = (props: Props) => {
     }
   }
 
+  const onOutcomeNameChange = (value: string) => {
+    setNewOutcomeName(value)
+    const val = value.toLowerCase();
+    outcomes.some(i => val.indexOf(i.name.toLowerCase()) >= 0) ? setDuplicateOutcome(true) : setDuplicateOutcome(false)
```
hpmaxi, line 218: This could be clearer, but is just an opinion, something like
```
const isDuplicated = outcomes.map(o => o.name.toLowerCase()).includes(value.toLowerCase())
setDuplicateOutcome(isDuplicated)
```

Issue #974, opened 2020-07-20
add custom trading fee to Funding Market Widget


pimato:
Currently, Omen has a fixed trading fee of 2%. We should allow market creators to decide for themself what fee they want to have:

<img width="749" alt="Screenshot 2020-07-20 at 16 51 55" src="https://user-images.githubusercontent.com/5337809/87951816-5bffdc80-caa9-11ea-82ca-b185da8d0013.png">

if the user clicks on "use custom trading fee":
<img width="653" alt="Screenshot 2020-07-20 at 16 54 16" src="https://user-images.githubusercontent.com/5337809/87952385-0ed03a80-caaa-11ea-8f65-23a7d884d6d4.png">

the user can always use the default market fee of 2% by clicking on "use default trading fee".

KadenZipfel:
Since the designs don't quite match up with our current implementation, @pimato where do you think the trading fee input should be?

Screenshot of current implementation for reference:
![image](https://user-images.githubusercontent.com/30579067/87984202-77a9b980-ca8e-11ea-8dd0-7bd267a93da4.png)

pimato:
@KadenZipfel yeah we may start with the new asset picker first and then move over to the set custom trading fee. If you want to finish this up, just add the trading fee above the Wallet Balance for now like on the wireframes!

KadenZipfel:
Also, should there be a maximum trading fee? If so, what should it be?

auryn-macmillan:
> Also, should there be a maximum trading fee? If so, what should it be?
Yeah, I came to ask this same question.
I'd suggest setting a maximum fee (probably the current fee, 2%), so that you don't end up with people losing money by accidentally participating in markets with outrages fees. That would be a really poor user experience.

PR #980, opened 2020-07-20
Add custom trading fee

KadenZipfel:
Closes https://github.com/protofire/omen-exchange/issues/974


pimato:
@KadenZipfel can we remove those two gaps here:
Screenshot 2020-07-21 at 15 47 14

pimato:
Also the "2" in the inputfield should be a placeholder so the user doesn´t need to remove it to set a different value: 
Screenshot 2020-07-21 at 15 55 04

app/src/components/market/sections/market_create/steps/items/funding_and_fee_step.tsx
```diff
@@ -207,6 +247,20 @@ const FundingAndFeeStep: React.FC<Props> = (props: Props) => {
...
+    setCustomFee(!customFee)
+  }
+
+  useEffect(() => {
+    const setMaxFeeWarning = () => {
+      spread > MAX_MARKET_FEE ? setExceedsMaxFee(true) : setExceedsMaxFee(false)
+    }
+
+    setMaxFeeWarning()
+  })
```
hpmaxi, line 262: This should be giving a warning about the dependency list of this effect, there is no need of a function here.
```
useEffect(() => {
  setExceedsMaxFee(spread > MAX_MARKET_FEE)
}, [spread])
```

app/src/components/market/sections/market_create/steps/items/funding_and_fee_step.tsx
```diff
@@ -207,6 +247,20 @@ const FundingAndFeeStep: React.FC<Props> = (props: Props) => {
     setAllowanceFinished(false)
   }
 
+  const toggleCustomFee = () => {
+    if (customFee) resetTradingFee()
+    if (!customFee) setFee(undefined)
+    setCustomFee(!customFee)
+  }
```
hpmaxi, line 254: This could be clearer, maybe using an if-else instead of two if. Also `setCustomFee`  is always executed because there are no curly brackets.

```
const toggleCustomFee = () => {
  if(customFee) { 
     resetTradingFee()
     setCustomFee(false)
  } else {
    setFee(undefined)
    setCustomFee(true)
  }
}
```

PR #990, opened 2020-07-23
804 custom categories

nickplace:
Clean revision of changes from #920 with fixes from @pimato's comment


app/src/components/common/form/dropdown/index.tsx
```diff
@@ -224,12 +226,19 @@ const Items = styled.div<{
   ${props => props.showScrollbar && DropdownScrollbarCSS}
 `
 
+const SecondaryText = styled.div`
+  color: #757575;
```
KadenZipfel, line 230: This should be `color: ${props => props.theme.colors.textColor};`
Same for other colors.

app/src/components/common/form/dropdown/index.tsx
```diff
@@ -84,6 +84,7 @@ const DropdownVariantCardCSS = css`
   padding: 13px 25px;
   position: relative;
   border: 1px solid #dcdff2;
+  text-transform: capitalize;
```
KadenZipfel, line 87: I don't think this should be capitalized, any reason for it?

app/src/components/market/sections/market_list/market_home.tsx
```diff
@@ -318,9 +331,11 @@ export const MarketHome: React.FC<Props> = (props: Props) => {
     }
   })
 
-  const filterItems: Array<DropdownItemProps> = filters.map(item => {
+  const filterItems: Array<DropdownItemProps> = filters.map((item, index) => {
+    const count = index === 0 ? counts.open : index === 3 ? counts.closed : 0
```
KadenZipfel, line 335: Looks like this only displays the number of open and ended categories, and not for pending or finalized

pimato: @KadenZipfel yeah it is actually not easy to get those numbers from the subgraph I believe!

app/src/components/common/form/dropdown/index.tsx
pimato: it depends what we want to store in the future with the "custom" category feature. Do we want to store the new categories lowercase and then fetch them and transform as capitalize?

KadenZipfel: Yeah it's tricky, especially if you notice when you select a category, it will display the current category in all caps. I'm assuming that's because it's quite hard not to use different values from what's displayed.

app/src/components/common/form/dropdown/index.tsx
```diff
@@ -83,7 +83,7 @@ const DropdownVariantCardCSS = css`
   flex: 1;
   padding: 13px 25px;
   position: relative;
-  border: 1px solid #dcdff2;
+  border: 1px solid ${props => props.theme.colors.tertiary};
```
KadenZipfel, line 86: Though it's the same color value, this should be `props.theme.dropdown.buttonBorderColor`

app/src/components/common/form/dropdown/index.tsx
```diff
@@ -174,7 +176,7 @@ const DropdownVariantCardItemsCSS = css`
 `
 const DropdownVariantCardItemsContainerCSS = css`
   border-radius: 8px;
-  border: 1px solid #dcdff2;
+  border: 1px solid ${props => props.theme.colors.tertiary};
```
KadenZipfel, line 179: Should also be, `props.theme.dropdown.buttonBorderColor`

Issue #988, opened 2020-07-22
Rinkeby markets not loading


KadenZipfel:
Currently, both locally and at dxtest.eth.link, Rinkeby markets are not loading. Instead, the following error is appearing in the console: 

```
Uncaught (in promise) Error: contract not deployed (contractAddress="0x5d3a536e4d6dbd6114cc1ead35777bab948e3643", operation="getDeployed", version=4.0.46)
```

Curiously, pending markets are still loading for some reason.

Looks like the issue is from querying markets with a minimum timeout from this commit: https://github.com/protofire/omen-exchange/commit/0ca42078e12e7df2d6244c74a13fb52da27f3999

PR #991, opened 2020-07-23
Fix Rinkeby markets not loading 

KadenZipfel:
Closes https://github.com/protofire/omen-exchange/issues/988


app/src/queries/markets_home.ts
```diff
@@ -75,11 +78,16 @@ export const queryMyMarkets = gql`
   ${MarketDataFragment}
 `
 
-type buildQueryType = MarketFilters & { whitelistedCreators: boolean; whitelistedTemplateIds: boolean }
+type buildQueryType = MarketFilters & {
+  whitelistedCreators: boolean
+  whitelistedTemplateIds: boolean
+  context: Maybe<ConnectedWeb3Context>
+}
```
hpmaxi, line 85: I think you can use the networkId as prop, or maybe put the timeout as a variable instead of context, I think the first option is better.

Issue #995, opened 2020-07-24
Replace horizontal asset picker with asset dropdown in fund market widget


pimato:
Omen is using a horizontal asset picker which should be replaced with an asset-picker via dropdown:

<img width="777" alt="Screenshot 2020-07-24 at 18 38 09" src="https://user-images.githubusercontent.com/5337809/88414249-d6e43280-cddc-11ea-982d-a96239880bf8.png">

**Figma:**[
Design Playground](https://www.figma.com/proto/GNoehJ64fqfptghsjYqvTJ/Playground-V2?node-id=3452%3A177&viewport=515%2C755%2C0.5033186078071594&scaling=min-zoom)
[Design Specification
](https://www.figma.com/file/GNoehJ64fqfptghsjYqvTJ/Playground-V2?node-id=3452%3A177)

KadenZipfel:
Is there, and should there be, any way to enter max. balance?

pimato:
> Is there, and should there be, any way to enter max. balance?

Yes, 100%!

PR #1002, opened 2020-07-24
Replace asset picker

KadenZipfel:
Closes https://github.com/protofire/omen-exchange/issues/995

### DIFF
--- a/app/src/components/common/form/dropdown/index.tsx
+++ b/app/src/components/common/form/dropdown/index.tsx
@@ -1,11 +1,11 @@
-import React, { DOMAttributes, useCallback, useEffect, useState } from 'react'
-import styled, { css } from 'styled-components'
-
 import { CardCSS } from '../../card'
 import { CommonDisabledCSS } from '../common_styled'
 
 import { ChevronDown } from './img/ChevronDown'
 import { ChevronUp } from './img/ChevronUp'
+
+import React, { DOMAttributes, useCallback, useEffect, useState } from 'react'
+import styled, { css } from 'styled-components'
 
 export enum DropdownPosition {
   left,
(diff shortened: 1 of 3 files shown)


pimato:
Hey @KadenZipfel , I have a few things to address:

Remove the margin-top here:
Screenshot 2020-07-26 at 17 26 41

pimato:
The dropdown should not be active all the time, we should use the same look and feel as all our other dropdowns:
Screenshot 2020-07-26 at 17 27 32

pimato:
The Balance is missing:
Screenshot 2020-07-26 at 18 01 47

should:
Screenshot 2020-07-26 at 18 02 43

pimato:
if a new trading fee is set which is above the limit, the warning message has too much space which can be removed:
Screenshot 2020-07-26 at 18 03 45

pimato:
The message should be below the inputfield and above the set allowance component:
Screenshot 2020-07-26 at 18 05 34

app/src/components/common/form/dropdown/index.tsx
```diff
@@ -361,10 +376,13 @@ export const Dropdown: React.FC<Props> = props => {
             {placeholder && !isDirty ? placeholder : activeItem.content}
             {!!activeItem.secondaryText && <SecondaryText>{activeItem.secondaryText}</SecondaryText>}
           </CurrentItem>
-          <ChevronWrapper>
-            <ChevronDown />
-            <ChevronUp />
-          </ChevronWrapper>
+          <DropdownButtonRight>
+            <CurrentItemExtra>{getItemExtraContent(currentItemIndex)}</CurrentItemExtra>
```
hpmaxi, line 380: I don't know if it is better for styling to have conditional rendered `<CurrentItemExtra>` in case of extraContent being empty, if is the case I would put after `activeItem` a declaration of `extraContent` that calls `getItemExtraContent(currentItemIndex)`.

KadenZipfel: Doesn't seem to be any styling issue if `extraContent` is empty

Issue #996, opened 2020-07-24
replace the current hardcoded categories in create market view with subgraph most used 8 categories


pimato:
the current categories in the create market view are hardcoded. We need to replace the hardcoded list with the most used 8 categories based on our subgraph. Additionally we need to add the function to load more categories:

<img width="748" alt="Screenshot 2020-07-24 at 19 54 37" src="https://user-images.githubusercontent.com/5337809/88420661-858d7080-cde7-11ea-8438-c741d6d1328b.png">

deliverables:
- replace hardcoded categories with eight most used categories based on the omen subgraph
- add a load more button which will be disabled once all categories have been fetched
- align all categories horizontal centered

**Figma**:
[Design Specification](https://www.figma.com/file/R056tSYMjvfwaFidEBItYy/Widgets-V2?node-id=57%3A0)

PR #1007, opened 2020-07-27
Dynamically render categories

KadenZipfel:
Closes https://github.com/protofire/omen-exchange/issues/996

### DIFF
--- a/app/src/components/market/common/categories/index.tsx
+++ b/app/src/components/market/common/categories/index.tsx
@@ -11,7 +11,7 @@ interface Props {
   name: string
   onChange?: any
   selectedCategory: string
-  setFirst: React.Dispatch<React.SetStateAction<number>>
+  setFirst: (n: number) => void
   first: number
   loadMoreButton: boolean
 }
--- a/app/src/components/market/common/categories/index.tsx
+++ b/app/src/components/market/common/categories/index.tsx
@@ -11,6 +11,9 @@ interface Props {
   name: string
   onChange?: any
   selectedCategory: string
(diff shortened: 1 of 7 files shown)


app/src/components/market/common/categories/index.tsx
```diff
@@ -46,10 +50,12 @@ const CategoryInput = styled(Textfield)<{ isSelected: boolean }>`
 `
 
 export const Categories = (props: Props) => {
-  const { categories, name, onChange, selectedCategory, ...restProps } = props
+  const { categories, first, loadMoreButton, name, onChange, selectedCategory, setFirst, ...restProps } = props
   const [selectedCustom, setSelectedCustom] = useState(false)
 
-  const allCategories = CATEGORIES.concat(categories.filter(item => CATEGORIES.indexOf(item) < 0))
+  let allCategories: string[]
+  categories.length > 0 ? (allCategories = categories) : (allCategories = CATEGORIES)
```
hpmaxi, line 57: This seems to be responsability of the component that use Categories, maybe passing categories as undefined and giving a default value here with `const { categories = CATEGORIES, ... } = props`. Or directly doing `categories={categories.length ? categories : CATEGORIES}`.

app/src/components/market/common/categories/index.tsx
```diff
@@ -11,12 +11,16 @@ interface Props {
   name: string
   onChange?: any
   selectedCategory: string
+  setFirst: React.Dispatch<React.SetStateAction<number>>
```
hpmaxi, line 14: I think this type could also be written as `(n: number) => void`  in other places is used with that convention.

Issue #1006, opened 2020-07-27
Set max fee to 5%


KadenZipfel:
Should be reflected by the front-end displaying all markets under 5% fee

PR #1008, opened 2020-07-27
Set max fee to 5% & format custom fee input value

KadenZipfel:
Closes https://github.com/protofire/omen-exchange/issues/1006

### DIFF
--- a/app/src/components/market/sections/market_create/steps/items/funding_and_fee_step.tsx
+++ b/app/src/components/market/sections/market_create/steps/items/funding_and_fee_step.tsx
@@ -262,6 +262,12 @@ const FundingAndFeeStep: React.FC<Props> = (props: Props) => {
     setExceedsMaxFee(spread > MAX_MARKET_FEE)
   }, [spread])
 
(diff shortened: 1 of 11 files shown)


app/src/components/market/sections/market_create/steps/items/funding_and_fee_step.tsx
```diff
@@ -262,6 +262,12 @@ const FundingAndFeeStep: React.FC<Props> = (props: Props) => {
     setExceedsMaxFee(spread > MAX_MARKET_FEE)
   }, [spread])
 
+  const handleFeeChange = (e: ChangeEvent<HTMLInputElement>) => {
+    const t = e.target.value
+    e.target.value = t.indexOf('.') >= 0 ? t.substr(0, t.indexOf('.')) + t.substr(t.indexOf('.'), 3) : t
+    handleChange(e)
```
hpmaxi, line 267: Maybe this is responsibility of handleChange that uses NumberInput, so the processing should be done there. This could be also extracted to another function (e.g. in tools) and tested so you can use different implementations in the future (e.g. using Number, toString and toFixed(3), or regular expressions).

PR #1017, opened 2020-07-29
Do not maintain in history market creation. 

### LINKED ISSUE #1014: After Market creation back button in market details view is routing you back to create market process
The back button should route back to the main market overview page which is default on https://omen.eth.link/

hpmaxi:
If there are no elements in history object then go back home.
Closes #1014

### DIFF
--- a/app/src/components/common/text/section_title/index.tsx
+++ b/app/src/components/common/text/section_title/index.tsx
@@ -46,7 +46,7 @@ export const SectionTitleWrapper: React.FC<Props> = (props: Props) => {
   return (
     <Wrapper {...restProps}>
       {enableGoBack && (
-        <ButtonCircle onClick={() => props.history.goBack()}>
+        <ButtonCircle onClick={() => (props.history.length > 2 ? props.history.goBack() : props.history.replace('/'))}>
           <IconArrowBack />
         </ButtonCircle>
       )}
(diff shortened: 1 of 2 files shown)


app/src/components/common/text/section_title/index.tsx
```diff
@@ -46,7 +46,7 @@ export const SectionTitleWrapper: React.FC<Props> = (props: Props) => {
   return (
     <Wrapper {...restProps}>
       {enableGoBack && (
-        <ButtonCircle onClick={() => props.history.goBack()}>
+        <ButtonCircle onClick={() => (props.history.length > 2 ? props.history.goBack() : props.history.replace('/'))}>
```
KadenZipfel, line 49: This fails if there is any history prior to visiting the create market page. For example, visit a market -> go to buy -> go to create market -> taken to market details page -> back button goes to create market page

KadenZipfel: Although not perfect, considering the complex routing for filters, this may be the best fix at the moment.

hpmaxi: The idea is to replace the history item related to market creation with the detail (replace instead of push), that piece of code is to have a working back button when you entered directly to the market.
I'm not sure if react history gives us enough flexibility to do much more without a custom solution.

hpmaxi: Maybe reverting changes to use a goBack string could work for more complex scenarios, but in that case, it should fallback too to a default route too. Let me know if more changes are needed.

Issue #1015, opened 2020-07-29
Show alert message if user sets a negative fee


pimato:
Currently, if the user wants to set a negative fee he/she can do that. We should restrict the user by showing him a warning message:
<img width="620" alt="Screenshot 2020-07-29 at 12 27 04" src="https://user-images.githubusercontent.com/5337809/88789261-d2ca6300-d196-11ea-87f3-f09ea8bdb42c.png">

KadenZipfel:
At any time if the custom fee is invalid, do we want to display the fee as 2% in green as you have above? E.g. both if the custom fee is negative or >5

pimato:
> At any time if the custom fee is invalid, do we want to display the fee as 2% in green as you have above? E.g. both if the custom fee is negative or >5

The best way forward would be actually to remove the "Earn Trading Fee 2%" if the user wants to set its own custom trading fee as it is redundant information.

[Playground
](https://www.figma.com/proto/GNoehJ64fqfptghsjYqvTJ/Playground-V2?node-id=3452%3A177&viewport=832%2C601%2C0.7815516591072083&scaling=min-zoom)
[Design specification](https://alchemy.daostack.io/dao/0x519b70055af55a007110b4ff99b0ea33071c720a/schemes)

pimato:
this will be eliminated with the dropdown picker!

PR #1024, opened 2020-07-30
Disallow negative fee and funding amount

KadenZipfel:
Closes https://github.com/protofire/omen-exchange/issues/1015

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "app",
-  "version": "1.1.2",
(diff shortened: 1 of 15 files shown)


pimato:
small thing:
Screenshot 2020-07-31 at 11 03 18

Deposit amount value should have default a grey color, only if the user puts something into the inputfield then we show a dark color.

KadenZipfel:
Closing as this will no longer be needed

Issue #1020, opened 2020-07-29
Issue with decimal places and custom fee


KadenZipfel:
Generally when setting a custom fee, users are capped at using max. 2 decimal places. It's possible to get past this by typing '0.000...'. This appears to be bad UX, especially considering if a user types a digit other than 0 at the end of that string, it just reverts back to 0.

pimato:
How would you fix it @KadenZipfel ? One alternative is to not have an inputfield at all but a dropdown which lets you pick:
0% , 0.25%, 0.5% and so on ... I think this could eliminate that issue!

pimato:
<img width="608" alt="Screenshot 2020-07-31 at 15 22 35" src="https://user-images.githubusercontent.com/5337809/89039032-b0c11400-d341-11ea-903b-2e443f438b17.png">

**Figma:**[
Design Specification](https://www.figma.com/file/GNoehJ64fqfptghsjYqvTJ/Playground-V2?node-id=3460%3A9)

KadenZipfel:
> How would you fix it @KadenZipfel ? One alternative is to not have an inputfield at all but a dropdown which lets you pick:
> 0% , 0.25%, 0.5% and so on ... I think this could eliminate that issue!

Agreed!

PR #1031, opened 2020-07-31
Create custom fee dropdown selector

KadenZipfel:
Closes https://github.com/protofire/omen-exchange/issues/1020

### DIFF
--- a/app/src/components/market/common/trading_fee_selector/index.tsx
+++ b/app/src/components/market/common/trading_fee_selector/index.tsx
@@ -31,17 +31,15 @@ interface Props {
 export const TradingFeeSelector: React.FC<Props> = props => {
   const { disabled, onSelect, ...restProps } = props
 
-  const tradingFeeDropdownItems: Array<DropdownItemProps> = []
-
   const onChange = (fee: string) => {
     if (TRADING_FEE_OPTIONS.includes(fee)) onSelect(fee)
   }
 
-  TRADING_FEE_OPTIONS.map(option => {
-    tradingFeeDropdownItems.push({
+  const tradingFeeDropdownItems: Array<DropdownItemProps> = TRADING_FEE_OPTIONS.map(option => {
+    return {
       content: `${option}%`,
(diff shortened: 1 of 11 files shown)


pimato:
@KadenZipfel  can we limit the height to something reasable and have it scrollable?

Currently:
Screenshot 2020-08-01 at 08 24 19

Also, 0% is missing!

app/src/components/market/common/trading_fee_selector/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+      }
+    }
+  }
+
+  tradingFeeOptions.forEach(option => {
+    tradingFeeDropdownItems.push({
+      content: `${option}%`,
+      onClick: () => onChange(option),
+    })
+  })
```
hpmaxi, line 72: This could be done with a map.

app/src/components/market/common/trading_fee_selector/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  const tradingFeeDropdownItems: Array<DropdownItemProps> = []
+
+  const onChange = (fee: string) => {
+    for (const option of tradingFeeOptions) {
+      if (option === fee) {
+        onSelect(option)
+      }
+    }
+  }
```
hpmaxi, line 65: `if(tradingFeeOptions.includes(option) { onSelect(option) }` could work too.

app/src/components/market/common/trading_fee_selector/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+    '3.00',
+    '3.25',
+    '3.50',
+    '3.75',
+    '4.00',
+    '4.25',
+    '4.50',
+    '4.75',
+    '5.00',
+  ]
```
hpmaxi, line 55: This could be done programmatically, there is a range function in `tools` but it doesn't have a step parameter. An "alternative" is to move it to a constants.

Issue #1032, opened 2020-08-03
Redeem Winings view sell and buy button still shown


pimato:
looks like the redeem button is not replacing the sell and buy button on market closed/redeem view:
<img width="605" alt="Screenshot 2020-08-03 at 15 08 12" src="https://user-images.githubusercontent.com/5337809/89186003-7488f000-d59b-11ea-93d1-c1b961b4e643.png">

PR #1036, opened 2020-08-03
Properly display market buttons for finalized and ended markets

KadenZipfel:
Closes https://github.com/protofire/omen-exchange/issues/1032


pimato:
Hey @KadenZipfel ,
after resolving the market and redeeming the winnings it seems like al buttons are again gone:
Screenshot 2020-08-04 at 17 51 55

Issue #1040, opened 2020-08-05
replace realit.io url redirect with reality.eth access


pimato:
Currently:
https://realitio.github.io/#!/question/0xd1e8e274bea94f038aaa30000d2be2a7de5e9db4580edec24f109a374f88b3cd

...it should now go to:
https://reality.eth.link/app/#!/question/0xd1e8e274bea94f038aaa30000d2be2a7de5e9db4580edec24f109a374f88b3cd
(Note the /app in the URL that wasn't there before)

PR #1042, opened 2020-08-05
replace all hyperlinks to ens.link domain

pimato:
closes: https://github.com/protofire/omen-exchange/issues/1040


KadenZipfel:
Code looks good.

I noticed though that the import market modal still has a placeholder for the old URL, which may be worth updating. It also says "Import Market from realit.io", maybe better if it said "Import Market from reality.eth.link" or something like that.

Issue #1001, opened 2020-07-24
Add max deposit functionality to fund market widget


KadenZipfel:
Should be an option to allow users to set the funding to their wallet balance.

pimato:
<img width="787" alt="Screenshot 2020-07-31 at 15 07 15" src="https://user-images.githubusercontent.com/5337809/89037832-879f8400-d33f-11ea-9636-e7ce0cf0d0fc.png">

**Figma:**
[Design Specification](https://www.figma.com/file/R056tSYMjvfwaFidEBItYy/Widgets-V2?node-id=57%3A605)

PR #1047, opened 2020-08-06
Add max deposit functionality to fund market widget

KadenZipfel:
Intends to close https://github.com/protofire/omen-exchange/issues/1001

I'm trying to get this setup, but having some trouble figuring out a way that will work. On clicking the max button, we want to set the value of the input to the `collateralBalance` and trigger the `onchange` event so that `handleChange` will run. The problem seems to be that manually triggering the `onchange` event processes the event with the string value of the input instead of the `BigNumberInputReturn`. I'm not sure of a way to get past this without modifying the `handleChange` function significantly.

Note: The `handleAmount` and `onMax` functions are a bit of a mess as a result of experimentation, ideally we would call the `handleChange` function directly `onchange`.

### DIFF
--- a/app/src/components/common/form/textfield_custom_symbol/index.tsx
+++ b/app/src/components/common/form/textfield_custom_symbol/index.tsx
@@ -112,7 +112,7 @@ const MaxButton = styled.span`
 `
 
 export const TextfieldCustomSymbol = (props: Props) => {
-  const { disabled, formField, shouldDisplayMaxButton, onClickMaxButton, symbol, ...restProps } = props
+  const { disabled, formField, onClickMaxButton, shouldDisplayMaxButton, symbol, ...restProps } = props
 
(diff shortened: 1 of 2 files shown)


pimato:
@KadenZipfel I am getting an error:

Screenshot 2020-08-07 at 13 10 34

KadenZipfel:
> @KadenZipfel I am getting an error after clicking on the Max button:

Yeah, that's part of the problem I'm having getting this setup. I've updated the description to explain the issue.

mariano-aguero:
> > @KadenZipfel I am getting an error after clicking on the Max button:
> 
> Yeah, that's part of the problem I'm having getting this setup. I've updated the description to explain the issue.

Hello @KadenZipfel , I fixed some problems, and modify some name of the variables, could you check if is all ok ? and works as  expected ?

pimato (approved):

Issue #1053, opened 2020-08-13
Show alert if user sets negative amounts


KadenZipfel:
This was meant to be _partly_ fixed with https://github.com/protofire/omen-exchange/pull/1024, but since the original issue was no longer needed, the whole thing was mistakenly scrapped.

If funding amount is set as negative, an alert should appear and the create market button should be disabled. This should be implemented on initial market funding step; depositing and withdrawing liquidity; and buying and selling outcomes.

PR #1054, opened 2020-08-13
Show alert if user sets negative amount

KadenZipfel:
Closes https://github.com/protofire/omen-exchange/issues/1053


pimato:
Looks like I´ve found a bug!
Steps to reproduce:
1) Input `-1`
2) App crashes:

Screenshot 2020-08-28 at 09 46 27

pimato (changes requested):
check bug comment

pimato (approved):

Issue #1058, opened 2020-08-18
Implement hourly daily volume sort


KadenZipfel:
As per https://github.com/protofire/omen-subgraph/pull/41

PR #1059, opened 2020-08-18
Implement hourly daily volume

KadenZipfel:
Closes https://github.com/protofire/omen-exchange/issues/1058

**Note**:
The subgraph changes have been deployed to gnosis/omen-rinkeby, so that can be used for testing. We will have to figure out a plan as far as subgraph deployment and the next release, otherwise we either end up with things breaking in the current version (1.1.4), or things breaking in the new version (1.1.5). Until we decide our approach, I will mark this as a draft so it's not merged.


cag:
Hey, the changes are also accessible on Mainnet via `gnosis/omen`. I believe the current process is to use `gnosis/omen` for staging new subgraph developments and `protofire/omen` for production use.

KadenZipfel:
> Hey, the changes are also accessible on Mainnet via `gnosis/omen`. I believe the current process is to use `gnosis/omen` for staging new subgraph developments and `protofire/omen` for production use.

Ahh yeah, I see it on gnosis/omen now. The problem will be timing the deploy of protofire/omen along with the deploy of omen. Is there a way to schedule a subgraph deploy perhaps? That way we could deploy it at the same time the omen release proposal should be set to deploy.

pimato (approved):

Issue #979, opened 2020-07-20
update 3box comments to the newest version


pimato:

pimato:
@KadenZipfel can you reference the issue which we created for the 3box comment repository?
Gonna remove this from Mileston 1.1.6 for now

PR #1064, opened 2020-08-21
Update 3box comments to most recent version

KadenZipfel:
Intends to close https://github.com/protofire/omen-exchange/issues/979

This is incomplete. I've been struggling to get this working for quite a while, so I'm sharing it to get a fresh set of eyes to look at it. It shouldn't be too hard, but for whatever reason I can't get it to work.


rusuandreirobert:
This currently locks you into a loop where you have to login / sign message, I've got 30 metamask requests to sign the message within 1 minute, the work-around was to first register to 3box.io -> then refresh the localhost page, and it would act normally, I think the registration process is "scewed", 

I think this is because the library tries to have you login "out of the box" 

For example, I went here : https://omen.eth.link/#/0x95b2271039b020aba31b933039e042b60b063800

I login & if I refresh the page I have to login back, the login is under the comment box.

While on localhost, on the same "link", I am always logged in after logging in once, but the questions will not show up.

I think this is a problem from the 3box library, because it detects the total number of comments, "14 comments" which is correct.

Leaving a comment on localhost will not reflect on : https://omen.eth.link/#/0x95b2271039b020aba31b933039e042b60b063800

By checking the live preview of 3box here : https://3box.github.io/3box-comments-react/examples/dist/ it looks like the comment box is showing "commenting as" when having it active, while on this version it doesn't work.

Checking in the example code here : https://github.com/3box/3box-comments-react/blob/master/examples/Example.js

It looks like the comments will load up only when the user is logged in : 

```
  handleLogin = async () => {
    const addresses = await window.ethereum.enable();
    const myAddress = addresses[0];

    const box = await Box.openBox(myAddress, window.ethereum, {});
    const myProfile = await Box.getProfile(myAddress);

    **box.onSyncDone(() => this.setState({ box }));
    this.setState({ box, myProfile, myAddress, isReady: true });**
  }
```

KadenZipfel:
Closing as we will be migrating from 3box to ceramic soon.

Issue #1068, opened 2020-08-24
Make currency dropdown scrollable


KadenZipfel:
With all the new tokens that may be added, see: https://alchemy.daostack.io/dao/0x519b70055af55a007110b4ff99b0ea33071c720a/proposal/0xf87d5db380964738547d92c7562d1b0e7dee6a41d034c672f5037df0e8451731, the currency dropdown will likely be crowded, and should have a fixed height with a scrollbar

PR #1070, opened 2020-08-24
Quickfix for #1068

rusuandreirobert:
Cap at 204px, initially just tought 200px would do the trick, but then I've taken into account the size of each item, looks like it's capped at 48px, 4 items per in the view seems to work really well.

48 * 4 + 12 ( padding top )

#1068


rusuandreirobert:
Okay, changed it so it uses only maxHeight & showScrollBar, the advanced_filters seems to use a generic dropdown, maybe in the future it should be adjusted to use the currency_selector component instead, and have all currency dropdowns come from there.

KadenZipfel:
> maybe in the future it should be adjusted to use the currency_selector component instead, and have all currency dropdowns come from there.

Yeah, good call they should absolutely be consolidated. I'll make an issue for this.

pimato (approved):

Issue #1055, opened 2020-08-17
Better visualizations for small market movements


auryn-macmillan:
Currently, small market movements are very difficult to see in the "Trade History" on Omen. This often makes it appear as if there is less activity than there actually is.

We should improve this view to better visualize small market movements.

One solution might be to reduce the range of the Y axis to sightly higher and lower than the maximum and minimum prices, respectively, in a given range of time.

For example, say that a market is around 50:50 but has had several <1% market movements, the graph might display data between 45% and 55% on the Y axis.

KadenZipfel:
This is a great idea, I'll look into how doable this is with the charting library we're using

PR #1071, opened 2020-08-24
Chart 2 digits

rusuandreirobert:
Related to #1055, it seems like the best solution is to simply chart 2 digits after the comma.

Video preview : https://d.pr/v/6BS2RW

Changing the default of "fixed" to 2 instead of 0 seems to be better, because of : 

`<YAxis orientation="right" tickFormatter={**toPercent**} />`

This function seems to be used strictly within the charting : 

Screenshot 2020-08-24 at 16 20 55

### DIFF
--- a/app/src/components/market/common/history_chart/chart.tsx
+++ b/app/src/components/market/common/history_chart/chart.tsx
@@ -141,7 +141,7 @@ const timestampToDate = (timestamp: number, value: string) => {
   return ts.format('MMM D')
 }
 
-const toPercent = (decimal: number, fixed = 0) => {
+const toPercent = (decimal: number, fixed = 2) => {
   return `${(decimal * 100).toFixed(fixed)}%`
 }
 
--- a/app/src/components/market/common/history_chart/chart.tsx
+++ b/app/src/components/market/common/history_chart/chart.tsx
@@ -201,7 +201,7 @@ export const HistoryChart: React.FC<Props> = ({ holdingSeries, onChange, options
       <ResponsiveContainer height={300} width="100%">
         <AreaChart data={data} margin={{ top: 10, right: 30, left: 0, bottom: 0 }} stackOffset="expand">
           <XAxis dataKey="date" />
-          <YAxis orientation="right" tickFormatter={toPercent} />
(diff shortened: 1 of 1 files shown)


KadenZipfel:
This doesn't seem to make a difference for visualizing graph changes as far as I can tell. Ideally we want to shorten the range of the Y axis, though I'm not sure whether it's possible with this charting library.

rusuandreirobert:
> Let's shorten the range of the Y axis if possible.

seems possible : 

```
<YAxis type="number" domain={['dataMin', 'dataMax']} />
<YAxis type="number" domain={[0, 'dataMax']} />
<YAxis type="number" domain={['auto', 'auto']} />
<YAxis type="number" domain={[0, 'dataMax + 1000']} />
<YAxis type="number" domain={['dataMin - 100', 'dataMax + 100']} />
<YAxis type="number" domain={[dataMin => (0 - Math.abs(dataMin)), dataMax => (dataMax * 2)]} />
```

Source : https://stackoverflow.com/questions/50078787/recharts-set-y-axis-range 

But, seems there's an issue when using : `stackOffset === 'expand'`

`stackOffset === 'expand' ? [0, 1] : other-behaviour`

I've tried : 

`domain={[ 0.01, 0.09 ]}` 
   -> doesn't change anything.

`domain={[ 0.1, 0.9 ]}` 
   -> doesn't change anything.

`domain={[ 1, 9 ]}`

Screenshot 2020-08-25 at 00 22 20

Data using for testing : 

```
  let temp_data = [];

  for( let i = 1; i <= 23; i++ )
      temp_data.push( {
          Greenlit  : 0.8379537053181649 + ( 0.001 * i ),
          Deferred  : 0.16204629468183507 + ( 0.001 * i ),
          date      : "03:" + ( i <= 9 ? '0' + i : i )
      } );
```

Making a commit where I add this attribute `domain={['auto', 'auto']` in the code, to make it simple to continue.

rusuandreirobert:
I still think it would be better to show 2 digits there on the hover & interaction, I don't know how to change the graph, small changes would still be visible there especially if someone hovers over and goes left -> right and notices the digits change.

Issue #1065, opened 2020-08-22
Display future shares on buy and sell


KadenZipfel:
Just as we show the change in share price, we should show the change in 'My Shares'. 
![image](https://user-images.githubusercontent.com/30579067/90945593-9f1bdb00-e3da-11ea-80cc-8804d10af9d1.png)

PR #1078, opened 2020-08-25
Display future shares on buy and sell inputs

KadenZipfel:
Closes https://github.com/protofire/omen-exchange/issues/1065


pimato:
can we not have a transition showing for outcomes which are not currently being bought or sold?

For example:
Screenshot 2020-08-26 at 13 16 54

Outcome No should not have a transition in My Shares

pimato (changes requested):
see comment.

pimato (approved):

Issue #1085, opened 2020-09-01
Replace 24h Volume with Highest liquidity as the default sorting


pimato:
<img width="427" alt="Screenshot 2020-09-01 at 16 55 48" src="https://user-images.githubusercontent.com/5337809/91867596-1b80aa80-ec74-11ea-9333-1840cdea2fd6.png">

based on the small poll in the telegram group chat people expressed interest in having the highest liquidity as the default sorting option!

PR #1097, opened 2020-09-04
Set highest liquidity as default sort param

KadenZipfel:
Closes https://github.com/protofire/omen-exchange/issues/1085


pimato (approved):
Works!

Issue #1099, opened 2020-09-04
Add option to display all markets (valid & invalid)


KadenZipfel:
Currently we only give the option for valid or invalid markets. Instead, we should default to all markets.

pimato:
this was fixed here: https://github.com/protofire/omen-exchange/pull/1066

PR #1101, opened 2020-09-04
Display valid and invalid markets by default

KadenZipfel:
Closes https://github.com/protofire/omen-exchange/issues/1099


pimato:
It looks like the filter component gets removed once the user picks `Market Validity - ALL` or `Market Validity - Valid` but not on `Market Validity - Invalid`.

We should in general keep the filter component visible once the user picks an item on any dropdown

pimato (changes requested):
Please check comments

KadenZipfel:
> It looks like the filter component gets removed once the user picks `Market Validity - ALL` or `Market Validity - Valid` but not on `Market Validity - Invalid`.
> 
> We should in general keep the filter component visible once the user picks an item on any dropdown

The problem with that approach is that the advanced filters will be always be open, even on initial page load. It's also worth noting that the other filters only stay open if the selected item is not 'All', so I've done the same for the market validity in this most recent commit.

pimato:
Hey @KadenZipfel , there is a bug with the pagination:

1. launch omen locally
2. click on the filter button
3. pick on the `market validity - dropdown` => `Valid`
4. pick on the `market validity - dropdown` => `Invalid`
5. pick on the `market validity - dropdown` => `All`
6. BUG: pagination does not work anymore

KadenZipfel:
> Regarding the collapse of the filter components once the user clicks on `All`. Is there really no way to keep the component collapsed? It kind a breaks the UX as we indicate the user that the component stays collapse on basically everything else besides `All`. How much work would it require?

Got it setup. Currently the advanced filters will be collapsed on page load if the selection is 'All', otherwise they will be open on page load. Additionally, all filters will stay open for any selection.

pimato (approved):

PR #1107, opened 2020-09-09
Add Authereum login option

miguelmota:
Demo: https://omen.authereum.com

Screenshot from 2020-09-08 19-49-43


pimato:
Hey @miguelmota , I just tried to make a trade through Authereum and it seems like there are issues with the gaslimit:
Screenshot 2020-09-09 at 15 11 05

KadenZipfel (approved):
Looks like the code is even better than before. Good work!

Issue #1084, opened 2020-09-01
Allow markets to be filter in the My Market view


pimato:
We should allow the user to filter their markets by:
- Ending Soon <--- Should be the default filter option
- Highest Pool Tokens
- Highes Shares

We should allow them to filter by category via the main Category Picker!

Additionally we should allow them to filter by:
- Currency
- Arbitrator

like this:
<img width="698" alt="Screenshot 2020-09-01 at 16 50 34" src="https://user-images.githubusercontent.com/5337809/91866916-546c4f80-ec73-11ea-9fd3-2869ab923219.png">

KadenZipfel:
Some notes:
- With the current subgraph implementation, we can't filter by pool tokens or outcome shares. I've created an issue to add it to the subgraph: https://github.com/protofire/omen-subgraph/issues/53
- We should also be able to filter by curation source, which will also need a subgraph modification: https://github.com/protofire/omen-subgraph/issues/54

If I have the time I'll make PR's for these subgraph changes.

Issue #1083, opened 2020-09-01
Allow searching for My Markets View


pimato:
The search is currently disabled for the my markets view. The search should work exactly like the main market view search works.

KadenZipfel:
This will require a subgraph modification: https://github.com/protofire/omen-subgraph/issues/55

PR #1116, opened 2020-09-14
Add filters to My Markets view

KadenZipfel:
Will close https://github.com/protofire/omen-exchange/issues/1084 and https://github.com/protofire/omen-exchange/issues/1083

Todo:
- [x] Sort by number of pool tokens and outcome shares
  - Depends on: https://github.com/protofire/omen-subgraph/issues/53
  - As per: https://github.com/protofire/omen-subgraph/issues/53#issuecomment-702887425, this can't be done right until ERC-20 wrapped tokens are ready and we figure out a solution
- [x] Filter by curation source
  - Depends on: https://github.com/protofire/omen-subgraph/issues/54
  - **Checked off as it's not possible**
- [x] Enable search
  - Depends on: https://github.com/protofire/omen-subgraph/issues/55
- [x] Deploy subgraph changes to production


pimato:
Hey @KadenZipfel just checking the search functionality and it seems like the search is not getting updated with the search tag. Steps to reproduce:

1) go to localhost
2) pick the my markets picker in the market state picker
3) click on search icon
4) put int "about"
5) Getting this:
Bildschirmfoto 2020-10-03 um 14 22 02
**BUG: it shows all markets again instead of doing the query**

Also, if use `http://localhost:3000/#/ending/curation-source/All%20Sources?state=MY_MARKETS&tag=about`

I am getting:
Bildschirmfoto 2020-10-03 um 14 26 20

which is also not right, because I have a market with the tag `about`: 
Bildschirmfoto 2020-10-03 um 14 27 19

pimato:
Also, I think we should set the default sorting criteria to `Newest`

pimato (changes requested):
see comments.

pimato (approved):

PR #1126, opened 2020-09-24
Upgrade Ethers, CPK, Typescript, create-react-scripts

### LINKED ISSUE #1033: gas limit estimations are way off
I am currently trying to redeem winnings for market:
https://omen.eth.link/#/0xe17a114ebe42ee2065fdc642398fc0e993142b4a

it looks like the gas limit estimations are way to high:
Screenshot 2020-08-03 at 15 11 47

cag:
Upgrading to Ethers v5 to try and fix #1033 by throttling web requests.
That required upgrading to CPK v2.0.0, which contains Ethers v5 support.
That in turn required upgrading Typescript to at least 3.8 for some reason I forgot now.
That requires upgrading the CRA scripts.

I'm still not sure what's going on with the dev server, as I'm having a hard time building because the compiler complains that I haven't listed @ethersproject/provider as a dependency on my machine yet even though it's clearly in the package.json >:o.


cag:
Closing this because it doesn't seem to be directly related to #1033 in the way I originally thought. This is probably something nice to have, but the way I've gone about it here is a bit messy and I'm not happy with it.

PR #1138, opened 2020-10-02
Fix overflow bug in getPayouts function

KadenZipfel:


cag (approved):
Yeah, I'm surprised this scenario didn't get caught. Clearly was a bug.

pimato (approved):

PR #1153, opened 2020-10-06
Reproducible content hash builds

luzzif:
Use `react-app-rewired` to update React Scripts' webpack configuration, achieving reproducible content hash builds artifacts in the process.


pimato (approved):

KadenZipfel (approved):
Code looks great. Have we tested whether we can get a reproducible content hash? Would be a good last step before merging this.

pimato:
> Code looks great. Have we tested whether we can get a reproducible content hash? Would be a good last step before merging this.

@luzzif and me did a run but we can do a run and it worked! We can do another run for the upcoming release `1.1.8`

Issue #1120, opened 2020-09-18
Add adaptive horizontal navigation to Trade Outcome - Widget


pimato:
The trade outcome widget needs a way to navigate between different market operations which will also adopt based on the market state.

## Market State - Open

<img width="834" alt="swap" src="https://user-images.githubusercontent.com/5337809/93618127-2c048500-f9d7-11ea-8c86-2af0b93d521a.png">

## Market State - Finalize

<img width="834" alt="finalize" src="https://user-images.githubusercontent.com/5337809/93619184-818d6180-f9d8-11ea-8aba-be95d9010284.png">

## Market State  - Closed

<img width="834" alt="redeem" src="https://user-images.githubusercontent.com/5337809/93619363-bf8a8580-f9d8-11ea-9f22-2dbf8c5a1a2a.png">

**Figma:**
Different States:
[Design Specification](https://www.figma.com/file/GNoehJ64fqfptghsjYqvTJ/Playground-V2?node-id=4249%3A0)
[Playground](https://www.figma.com/proto/GNoehJ64fqfptghsjYqvTJ/Playground-V2?node-id=4249%3A0&viewport=779%2C491%2C0.139102503657341&scaling=min-zoom)

Navigation:
[Design Specification](https://www.figma.com/file/GNoehJ64fqfptghsjYqvTJ/Playground-V2?node-id=4249%3A0)
[Playground](https://www.figma.com/proto/GNoehJ64fqfptghsjYqvTJ/Playground-V2?node-id=4302%3A1919&viewport=460%2C-348%2C0.38765960931777954&scaling=min-zoom)

Issue #1151, opened 2020-10-06
Market pool liquidity page on closed markets is failing


KadenZipfel:
Example: https://dxtest.eth.link/#/0xe17a114ebe42ee2065fdc642398fc0e993142b4a/pool-liquidity

It looks like somehow `bondTimestamp` is sneaking past the type check as not a BigNumber. `bondAnswer` is just `marketMakerData.question.currentAnswerTimestamp` renamed, so that would be the value to look at. 

Edit: I have yet to look into this properly, but I think the `currentAnswerTimestamp` gets removed when the market closes or something like that.

PR #1155, opened 2020-10-06
Add adaptive horizontal navigation

KadenZipfel:
Closes https://github.com/protofire/omen-exchange/issues/1120

Note: This issue: https://github.com/protofire/omen-exchange/issues/1151 leads to a couple of the closed market views to fail, but since it seems to be separate from this PR, I will resolve it separately. Fix: https://github.com/protofire/omen-exchange/pull/1156

Another note: It wasn't specified in the designs, but in the case that a market is closed and a user doesn't have any winning outcomes, I set the first tab to be called 'Results'.


pimato:
I just found one issue with the `History - Tab`. The title changes to `Purchases Shares`:
Bildschirmfoto 2020-10-07 um 13 18 57

pimato (changes requested):
see comments

pimato (approved):

Issue #1141, opened 2020-10-05
Add Information about liquidity provision back


pimato:
Hey guys,

I just realised that on dxtest.eth.link there is no information about the state of providing liquidity. 

Previous:
<img width="608" alt="Bildschirmfoto 2020-10-05 um 13 31 09" src="https://user-images.githubusercontent.com/5337809/95074594-17322c00-070f-11eb-93fc-d8c474fcb58a.png">

Now that we need to add it back, I took the chance to simplify the UI to:
<img width="862" alt="Bildschirmfoto 2020-10-05 um 13 32 08" src="https://user-images.githubusercontent.com/5337809/95074647-2f09b000-070f-11eb-81da-72c15d392a5e.png">

**Figma:**[
Design Specification](https://www.figma.com/file/R056tSYMjvfwaFidEBItYy/Widgets-V2?node-id=176%3A1716)

PR #1169, opened 2020-10-08
Add liquidity provision data

KadenZipfel:
Closes https://github.com/protofire/omen-exchange/issues/1141

### DIFF
--- a/app/src/components/common/text/title_value/index.tsx
+++ b/app/src/components/common/text/title_value/index.tsx
@@ -3,11 +3,7 @@ import React, { DOMAttributes } from 'react'
 import ReactTooltip from 'react-tooltip'
 import styled from 'styled-components'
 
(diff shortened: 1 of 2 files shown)


pimato:
Hey @KadenZipfel , just realized that my designs have been for the older version where we don´t have the horizontal navigation. I´ve updated the design specification accordingly:
https://www.figma.com/file/R056tSYMjvfwaFidEBItYy/Widgets-V2?node-id=176%3A1716

Also the rerwards number should get highlighted with a greenish color!

pimato (changes requested):
see comments.

KadenZipfel:
> Also the rerwards number should get highlighted with a greenish color!

I have it set so it will only be highlighted green if the amount is greater than 0. Is that good?

PR #1173, opened 2020-10-11
Fix redeem errors caused by numerical issues

cag:
Basically, the redeeming batch also calculates how much will be redeemed and forwards that amount which goes to the proxy back to the owner account. Unfortunately, that amount might also be wrong because of the way that the calculation adds up all the redeemed amounts before truncating occurs, when in reality, the contract truncates the payouts before adding them. You can see this off-by-one effect in [this successful transaction](https://rinkeby.etherscan.io/tx/0x646c050b04fee63fa4019311270d508f0610f6b00149d1bb98149a6e28e17be5) (look in the DAI transfers), which went through because I had other rounded stuff that was off before due to a previous truncated import of payouts from a conversion to a float, causing me to retain DAI dust in my proxy.

@pimato Dunno which issue tracks this, so if you could link it here, would be appreciated.

### DIFF
--- a/app/src/components/market/common/outcome_table/index.tsx
+++ b/app/src/components/market/common/outcome_table/index.tsx
@@ -1,3 +1,4 @@
+import Big from 'big.js'
 import { BigNumber } from 'ethers/utils'
 import React, { useCallback } from 'react'
 import styled, { css } from 'styled-components'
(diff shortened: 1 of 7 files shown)


pimato (approved):
Tested on Mainnet

app/src/components/market/common/outcome_table/index.tsx
```diff
@@ -1,3 +1,4 @@
+import Big from 'big.js'
```
KadenZipfel, line 1: Is there any reason we're using `big.js` instead of the ethers `BigNumber` library that we already use?

KadenZipfel (approved):

app/src/components/market/common/outcome_table/index.tsx
cag: Yeah, `big.js` is for real numbers, whereas `BigNumber` is only for integers. I think in the future maybe I'll have that requirement removed by adding the numerators to the subgraph, but right now, it's to import the payouts, which are BigDecimals, properly.

PR #1181, opened 2020-10-15
import market redesign

liaojikunwork:


pimato:
`Closing Date(UTC)` should be `Closing Date (UTC)`

pimato:
`Verified` should be `Verified by`

pimato:
Bildschirmfoto 2020-10-15 um 20 39 51

onHover does not work correctly

pimato:
Bildschirmfoto 2020-10-15 um 20 43 22

pimato:
Currently if I am in the fund market state like this:
Bildschirmfoto 2020-10-15 um 20 45 44

If I go back I see the `Configure Market - Categorical` view which is **wrong**:
Bildschirmfoto 2020-10-15 um 20 47 00

Omen should keep the question id in the url so we can route back to the original search like this:
Bildschirmfoto 2020-10-15 um 20 45 07

pimato:
The navigation should have the same styling as the market details view.

Currently:
Bildschirmfoto 2020-10-15 um 20 48 57

Should:
Bildschirmfoto 2020-10-15 um 20 49 10

liaojikunwork:
> The navigation should have the same styling as the market details view.
> 
> Currently:
> Bildschirmfoto 2020-10-15 um 20 48 57
> 
> Should:
> Bildschirmfoto 2020-10-15 um 20 49 10

@pimato 
Could you share the figma link of the screenshot you shared?

What I found is different from yours.
Screen Shot 2020-10-15 at 8 39 23 PM

pimato:
@liaojikunwork I am still not happy with the way the navigation was styled. the components move around when you switch between tabs. Please check how @KadenZipfel has implemented/styled the navigation and replicated it 100%.

pimato:
Please remove the underlining and animation from this:
Bildschirmfoto 2020-10-16 um 12 17 34

pimato:
Also, to not have this moving navigation lets rename the navigation items to `Categorical` and `Import Market`

liaojikunwork:
> Also, to not have this moving navigation lets rename the navigation items to `Categorical` and `Import Market`

@pimato 
I changed "Categorical Market" to "Categorical", but "import market" button still moves a bit. (difficult to notice)

pimato:
There is now one issue which I did not count for and needs to be handled. If an user imports a market which has an expired closing date we should prevent this as we would allow the user to proceed with an already closed market. Therefor we should add a warning message and disable the `Continue - Button`if the frontend detects that the closing date is expired.

[Design specification](https://www.figma.com/file/R056tSYMjvfwaFidEBItYy/Widgets-V2?node-id=1209%3A141)

KadenZipfel (approved):
Awesome work 👏

Issue #864, opened 2020-07-01
Can not submit market form after rejecting tx


leolower:
If the user declines the create market transaction, this button cannot be triggered again, it's a bit annoying as you need to fill the fields again.

PR #1187, opened 2020-10-19
fix: disable create market button during market creation

hexyls:
Fixes https://github.com/protofire/omen-exchange/issues/864


pimato (approved):

KadenZipfel (changes requested):
Although this gets the job done, it removes the intended functionality of `MarketCreationStatus`. Can you find a way to do it without removing the `MarketCreationStatus` checks?

hexyls:
`marketCreationStatus` is in ready state by default, so the `!MarketCreationStatus.is.ready(marketCreationStatus)` check is only effective during and after market creation - the `MarketCreationStatus.is.creatingAMarket(marketCreationStatus)` I added handles during but I should probably add a check for a done state too.

imo I don't think we should lock the UI if there is an error (not a good user experience), so I think it is fine to remove the error status check here.

KadenZipfel (approved):

PR #1188, opened 2020-10-19
added metamask check if user is connected for reality link

Mi-Lan:


app/src/components/market/common/additional_market_data/index.tsx
```diff
@@ -64,10 +64,27 @@ interface Props extends DOMAttributes<HTMLDivElement> {
...
+        setConnected(accounts.length !== 0)
+      })
+    }
+    if (windowObj.ethereum) {
+      fetchUserAccounts()
+    }
+  }, [windowObj.ethereum])
   const realitioBaseUrl =
-    windowObj.ethereum && windowObj.ethereum.isMetaMask ? 'https://reality.eth' : 'https://reality.eth.link'
+    windowObj.ethereum && windowObj.ethereum.isMetaMask && connected
```
KadenZipfel, line 85: Should be ok to remove `windowObj.ethereum && windowObj.ethereum.isMetaMask`, and just check that it's `connected`.

KadenZipfel (changes requested):
This is good, but there are other places where we set the reality link which need to be updated. Bonus points if you can consolidate the reality link generating process to single place and display the generated link in each corresponding location.

app/src/components/market/common/additional_market_data/index.tsx
Mi-Lan: i implemented that in this new commit!

Mi-Lan:
> This is good, but there are other places where we set the reality link which need to be updated. Bonus points if you can consolidate the reality link generating process to single place and display the generated link in each corresponding location.

Yeah that makes sense! I made new commit as you can see i made custom hook that encapsulates that logic and i just imported it everywhere i saw that reality link was used...

KadenZipfel (approved):
Good work 👏

Issue #1184, opened 2020-10-17
Allow user to change the probability of each outcome from an imported market


pimato:
With the new Import Market process/designs users can easily import markets from reality.eth.link questions. Currently, if a market is imported, the UI sets the a fixed uniform probability which is not very useful as there is a high probability that an user actually wants to change the probability of the imported market.

With that, we should enable the user to change the probability of each outcome of an imported market.

<img width="714" alt="Configure" src="https://user-images.githubusercontent.com/5337809/96336302-125e6800-107f-11eb-8eef-6409dfb2f179.png">

**Figma**:
[Design Specification](https://www.figma.com/file/R056tSYMjvfwaFidEBItYy/Widgets-V2?node-id=1209%3A141)

PR #1189, opened 2020-10-20
Allow user to change the probability of each outcome from an imported market #1184

liaojikunwork:
Allow user to change the probability of each outcome from an imported market #1184

### DIFF
--- a/app/src/components/market/common/common_styled/index.tsx
+++ b/app/src/components/market/common/common_styled/index.tsx
@@ -201,3 +201,10 @@ export const Breaker = styled.div`
     display: none;
   }
 `
+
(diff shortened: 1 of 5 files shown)


pimato:
Hey @liaojikunwork can we keep `Categorical Market` instead of `Categorical`.

pimato:
Can we remove the background you added here:
Bildschirmfoto 2020-10-20 um 15 47 32

it should look like this (with rounded corners): 
Bildschirmfoto 2020-10-20 um 15 48 03

pimato (changes requested):
see comments

pimato (approved):

KadenZipfel (approved):

PR #1192, opened 2020-10-21
preload list item data

hexyls:
Fetch a bit more data up front when browsing markets so the user can spend less time waiting for stuff to load.

gif before:

from

gif after: 
to


KadenZipfel (approved):
Awesome work 👏

PR #1195, opened 2020-10-21
Fix/1185 add liquidity balance, volume and date to market overview widget

liaojikunwork:


liaojikunwork:
Screen Shot 2020-10-21 at 5 45 30 AM
Screen Shot 2020-10-21 at 5 45 24 AM

pimato:
Please rename  `Time Remaining` to just `Remaining`

pimato:
Also the arrows should be used in the opposite way, default points down, if user clicks on it the arrow should point to the top.

pimato:
is there a reason why the icon here is not aligned with the same height as the description `Total Volume`?
Bildschirmfoto 2020-10-21 um 15 04 33
Bildschirmfoto 2020-10-21 um 15 04 13

pimato (approved):

app/src/components/market/common/market_data/index.tsx
```diff
@@ -89,26 +95,46 @@ export const MarketData: React.FC<Props> = props => {
...
       : '0'
 
+  const totalVolume = formatBigNumber(collateralVolume, currency.decimals)
+
   return (
     <MarketDataWrapper>
       <MarketDataItem>
         <MarketDataItemTop>
-          {moment(resolutionTimestamp).format('DD.MM.YYYY - H:mm zz')} {timezoneAbbr}
+          {liquidity} {currency.symbol}
```
KadenZipfel, line 104: `liquidity` should be wrapped in `formatNumber()` so it's displayed as: `50,010.00` as opposed to the current `50010.00`.

app/src/components/market/common/market_data/index.tsx
```diff
@@ -89,26 +95,46 @@ export const MarketData: React.FC<Props> = props => {
...
         </MarketDataItemTop>
-        <MarketDataItemBottom>Closing Date</MarketDataItemBottom>
+        <MarketDataItemBottom>Liquidity</MarketDataItemBottom>
       </MarketDataItem>
       <MarketDataItem>
         <MarketDataItemTop>
-          {resolutionTimestamp > new Date() ? moment(resolutionTimestamp).fromNow(true) : '0 days'}
+          <MarketDataItemImage src={currencyIcon && currencyIcon}></MarketDataItemImage>
+          {show24H ? dailyVolume : totalVolume} {currency.symbol}
```
KadenZipfel, line 111: `dailyVolume` and `totalVolume` should also be wrapped in `formatNumber()`

app/src/components/market/common/market_data/index.tsx
```diff
@@ -89,26 +95,46 @@ export const MarketData: React.FC<Props> = props => {
...
-          {moment(resolutionTimestamp).format('DD.MM.YYYY - H:mm zz')} {timezoneAbbr}
+          {liquidity} {currency.symbol}
         </MarketDataItemTop>
-        <MarketDataItemBottom>Closing Date</MarketDataItemBottom>
+        <MarketDataItemBottom>Liquidity</MarketDataItemBottom>
       </MarketDataItem>
       <MarketDataItem>
         <MarketDataItemTop>
-          {resolutionTimestamp > new Date() ? moment(resolutionTimestamp).fromNow(true) : '0 days'}
+          <MarketDataItemImage src={currencyIcon && currencyIcon}></MarketDataItemImage>
```
KadenZipfel, line 110: Looks like the currency icon isn't included in the design, so I would omit this unless @pimato has suggested otherwise

KadenZipfel (changes requested):
Great work overall, just some small changes

KadenZipfel:
Looks like we have some text overflow on this market: http://localhost:3000/#/0x002cce850e649f59d60107b56c1536f2065f5b60

image

app/src/components/market/common/market_data/index.tsx
liaojikunwork: ok

liaojikunwork:
> Looks like we have some text overflow on this market: http://localhost:3000/#/0x002cce850e649f59d60107b56c1536f2065f5b60
> 
> image

I think, I need to make another function to shorten the number. e.g: 10.3K, 1.5M, 22.4 ...

pimato:
> > Looks like we have some text overflow on this market: http://localhost:3000/#/0x002cce850e649f59d60107b56c1536f2065f5b60
> > image
> 
> I think, I need to make another function to shorten the number. e.g: 10.3K, 1.5M, 22.4 ...

lets keep it for now how it is. I think we have a better way of dealing with decimals for stablecoins and non-stablecoins. Gonna create an issue for tomorrow about it.

app/src/components/market/common/market_data/index.tsx
pimato: until we did not implemented the Currency picker for an market we should probably keep it for now.

KadenZipfel (approved):

Issue #1198, opened 2020-10-22
Change liquidity value with 1k represents 1,000.00


pimato:
Currently we show just the value with two decimals but to get some more room in width and to simplify the liquidity balance (because no one actually cares abotu 0.50 cents of DAI).

754.35 (can stay as is)

**Everything which is above 1,000.00**
1,000.00 = 1K
10,000 = 10K
10,750 = 10.75K
100,200 = 100.2K

PR #1202, opened 2020-10-22
Change liquidity value with 1k represents 1,000.00 #1198

liaojikunwork:
Screen Shot 2020-10-22 at 8 39 46 AM


liaojikunwork:
Everything which is above 1,000.00
1,000.00 = 1K
10,000 = 10K
10,750 = 10.75K
100,200 = 100.2K

Additionally, M(million), B(billion), and T(trillion) will work.

pimato:
@liaojikunwork can we also get a truncated number in the market details item here:
Bildschirmfoto 2020-10-22 um 14 54 01

liaojikunwork:
Screen Shot 2020-10-22 at 9 00 20 AM

pimato (approved):

KadenZipfel:
Do we want every number in the app to be formatted like this, or only in a select few places like it's in now? Because if we want this done with every number I would suggest just updating the `formatNumber` function to do this itself. 

@pimato what do you think?

app/src/util/tools.ts
```diff
@@ -344,3 +344,20 @@ export const formatNumber = (number: string, decimals = 2): string => {
 
   return `${formattedSubstring}${decimals > 0 ? '.' + fixedInt.split('.')[1] : ''}`
 }
+
+export const formatToShortNumber = (number: string, decimals = 2): string => {
```
KadenZipfel, line 348: We have a `tools.spec.ts` file in which we should add a test for this function

KadenZipfel (changes requested):

pimato:
> Do we want every number in the app to be formatted like this, or only in a select few places like it's in now? Because if we want this done with every number I would suggest just updating the `formatNumber` function to do this itself.
> 
> @pimato what do you think?

No, I think liquidity/volume is enough. In general I would always prefer to see the full balance but because of space and not really required we can ditch it here.

KadenZipfel (approved):

Issue #1193, opened 2020-10-21
Enable creation of scalar markets in the create market process


pimato:
Hey everyone,

this is the first attempt to create a generic UI for the creation of a scalar market on Omen. For a scalar market we need:
- **Market Question** (stored on-chain)
- **Lower Bound/Limit** (stored on-chain)
- **Upper Bound/limit** (stored on-chain)
- **Starting point** (stored on-chain)
- **Closing Date** (stored on-chain)
- **Category** (stored on-chain)
- **Unit of measurement** (Not stored on-chain, needs to be added to the Market Question/Title via a specific identifier like [USD], need to figure out what is the best one)

### Configure Scalar Market
<img width="714" alt="Create-Scalar" src="https://user-images.githubusercontent.com/5337809/96676925-c5281200-136e-11eb-91e3-b5224918c44a.png">

**Figma**: 
[Design Specification
](https://www.figma.com/file/R056tSYMjvfwaFidEBItYy/Widgets-V2?node-id=1209%3A141)

### Fund Scalar Market
<img width="714" alt="Fund-Scalar" src="https://user-images.githubusercontent.com/5337809/96677020-01f40900-136f-11eb-883d-200b23d7f1f8.png">

Please review the provided playgrounds and design specification first!

**Figma:**
[Playground](https://www.figma.com/proto/aTivxuHmp2l6Sn8nho1Ngu/Playground?node-id=115%3A8443&viewport=635%2C104%2C0.7277852892875671&scaling=min-zoom)
[Design Specification
](https://www.figma.com/file/aTivxuHmp2l6Sn8nho1Ngu/Playground?node-id=115%3A8443)

pimato:
Can we always put the dot horizontal aligned to the info box?
<img width="1231" alt="Bildschirmfoto 2020-10-23 um 12 29 28" src="https://user-images.githubusercontent.com/5337809/96993340-77a4d400-152b-11eb-9b36-e76cac21b3f0.png">

PR #1205, opened 2020-10-22
Scalar market creation front-end

KadenZipfel:
Closes https://github.com/protofire/omen-exchange/issues/1193

Note: This PR only addresses the front-end of the scalar market creation. 

The git flow for scalar markets will be to create a separate PR for each part of the process (front-end and back-end separate) and merge all PR's into the scalar-markets branch once approved. This way the process can stay modular without having unfinished work on the master branch.

Also, I've decided to omit the border-radius change to buttons as it will likely cause a ton of merge conflicts, so it's probably much easier to put it in its own issue.

There's also a mobile-view hiccup on the funding view, but that's probably also better solved in a separate issue as it also affects categorical markets in the same way.


pimato:
Can we always put the dot horizontal aligned to the info box?
Bildschirmfoto 2020-10-23 um 12 29 28

pimato:
Picky comment but when I start my journey with setting the Lower Bound and use a key to switch to the next input it brings me the Starting Point inputfield instead of going to right. Maybe that is an easy fix.

pimato:
When I filled out my scalar market:
Bildschirmfoto 2020-10-23 um 12 35 09

click continue, and go back again:
Bildschirmfoto 2020-10-23 um 12 35 41

Missing filled out data:
- Market question
- Closing Date
- Category

pimato (changes requested):
see comments

pimato (approved):

Issue #1216, opened 2020-10-26
New Filter/Search Next/Back button


pimato:
<img width="714" alt="Markets" src="https://user-images.githubusercontent.com/5337809/97159400-16733f80-177b-11eb-81a8-29f93d9b2efe.png">

- Note: We want to keep the Display market item size dropdown (it is not visibile in those mockups)

Figma:
[Playground](https://www.figma.com/proto/aTivxuHmp2l6Sn8nho1Ngu/Playground?node-id=361%3A18130&viewport=377%2C200%2C0.21331609785556793&scaling=min-zoom)
[Design Specification](https://www.figma.com/file/aTivxuHmp2l6Sn8nho1Ngu/Playground?node-id=361%3A17626)

PR #1219, opened 2020-10-26
New Filter/Search Next/Back button #1216

liaojikunwork:


pimato:
the divider should be removed.
Also if a button is active the icon should not be blue as it is right now.

pimato (dismissed):
check comment

pimato (approved):

app/src/theme/index.js
```diff
@@ -76,6 +76,7 @@ const theme = {
   colors: {
     activeListItemBackground: '#fafafa',
     alert: '#E57373',
+    borderColorDar: '#9FA8DA',
```
KadenZipfel, line 79: Looks like a typo here :p

liaojikunwork: Just fixed and pushed again.

KadenZipfel (approved):

PR #1229, opened 2020-10-26
fix: clear old results when sortBy filter changes

hexyls:
Prevents a quick flash of old results when the sortBy filter changes.


pimato (approved):

liaojikunwork (approved):

KadenZipfel (approved):
👏

PR #1253, opened 2020-10-28
Scalar market creation back-end

KadenZipfel:
Note: It's expected that the created scalar market will not work on the front-end as that will be done by this PR: https://github.com/protofire/omen-exchange/pull/1306

### DIFF
--- a/app/src/services/cpk.ts
+++ b/app/src/services/cpk.ts
@@ -2,12 +2,13 @@ import CPK from 'contract-proxy-kit/lib/esm'
 import EthersAdapter from 'contract-proxy-kit/lib/esm/ethLibAdapters/EthersAdapter'
 import { ethers } from 'ethers'
 import { TransactionReceipt, Web3Provider } from 'ethers/providers'
-import { BigNumber, formatUnits } from 'ethers/utils'
+import { BigNumber } from 'ethers/utils'
+import { Zero } from 'ethers/constants'
 import moment from 'moment'
 
 import { getLogger } from '../util/logger'
 import { getCPKAddresses, getContractAddress } from '../util/networks'
-import { calcDistributionHint } from '../util/tools'
+import { calcDistributionHint, clampBigNumber } from '../util/tools'
 import { MarketData, Question, Token } from '../util/types'
 
 import { ConditionalTokenService } from './conditional_token'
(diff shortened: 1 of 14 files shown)


app/src/services/cpk.ts
```diff
@@ -274,6 +274,165 @@ class CPKService {
...
+      })
+
+      // Step 4: Transfer funding from user
+      transactions.push({
+        to: collateral.address,
+        data: ERC20Service.encodeTransferFrom(account, this.cpk.address, marketData.funding),
+      })
+
+      // Step 4.5: Calculate probabilities for distributionHint
+      const a = scalarLow && scalarHigh && scalarStart && ((scalarStart - scalarLow) / (scalarHigh - scalarLow)) * 100
```
cag, line 396: I would write it such that if `scalarStart` exists, then

```js
// this might be a useful utility to have somewhere
function clampBigNumber(x, min, max) {
  if (x.lt(min)) return min;
  if (x.gt(max)) return max;
  return x;
}

const domainSize = scalarHigh.sub(scalarLow)

a = clampBigNumber(scalarHigh.sub(scalarStart), ethers.constants.Zero, domainSize)
b = clampBigNumber(scalarStart.sub(scalarLow), ethers.constants.Zero, domainSize)

// scaling to 100 is optional if distributionHint is calculated like below
// but these lines would do it, just for reference:
// a = Number(a.toString()) * 100 / Number(domainSize.toString());
// b = Number(b.toString()) * 100 / Number(domainSize.toString());
```

In fact, we can get the `distributionHint` directly, skipping `calcDistributionHint`:

```
const distributionHint = [b, a];
```

cag (commented):
Overall conceptually seems correct, but there're a few details to consider.

pimato:
Bildschirmfoto 2020-11-07 um 11 26 58

is there a chance to get the starting point box aligned horinzotally centered to the slider?

pimato (commented):
see comment

KadenZipfel:
> Bildschirmfoto 2020-11-07 um 11 26 58
> 
> is there a chance to get the starting point box aligned horinzotally centered to the slider?

I've linked to this in https://github.com/protofire/omen-exchange/pull/1306 as the change will have to be made there instead

Mi-Lan (approved):

pimato (approved):

liaojikunwork (approved):

Issue #1251, opened 2020-10-28
Restyle Deposit/Withdraw buttons in the pool section


pimato:
Currently:
<img width="625" alt="Bildschirmfoto 2020-10-28 um 16 57 22" src="https://user-images.githubusercontent.com/5337809/97462080-aad9cf80-193e-11eb-865d-7c8a4a3e0524.png">

Should:
<img width="819" alt="Bildschirmfoto 2020-10-28 um 16 57 44" src="https://user-images.githubusercontent.com/5337809/97462132-b6c59180-193e-11eb-8624-cc6a9dac5b8f.png">

PR #1255, opened 2020-10-29
Restyle Deposit/Withdraw buttons in the pool section #1251

liaojikunwork:
Screen Shot 2020-10-28 at 8 53 42 PM


KadenZipfel (approved):

app/src/components/button/button_tab/index.tsx
```diff
@@ -5,7 +5,8 @@ import { ButtonCSS, ButtonProps, ButtonType } from '../button_styling_types'
 
 const Wrapper = styled.button<ButtonProps>`
   ${ButtonCSS}
-  border-radius: 6px;
+  border-radius: 8px;
```
Mi-Lan, line 8: @liaojikunwork @pimato  should we maybe change the buttonCSS class itself ?

Mi-Lan (approved):
GTG! But we should add separate issue for changing whole ButtonCSS so we don't just override code like this it can create messy codebase really fast..

pimato (approved):

PR #1258, opened 2020-10-29
Fix/#1232

liaojikunwork:


app/src/components/market/common/market_top_details_open/index.tsx
```diff
@@ -1,16 +1,33 @@
...
+import { CurrencySelector } from '../currency_selector'
 import { MarketData } from '../market_data'
 import { ProgressBar } from '../progress_bar'
 import { ProgressBarToggle } from '../progress_bar/toggle'
 
+const SubsectionTitleLeftWrapper = styled.div`
+  display: flex;
+  align-items: center;
+  & > * + * {
+    margin-left: 20px;
```
pimato, line 21: margin-left: 12px

pimato (changes requested):
check comments

app/src/components/market/common/currency_selector/index.tsx
```diff
@@ -66,22 +77,24 @@ export const CurrencySelector: React.FC<Props> = props => {
...
-            return
-          },
+  tokens
+    .filter(({ address }) => filters.length === 0 || filters.indexOf(address.toLowerCase()) >= 0)
+    .forEach(({ address, image, symbol }, index) => {
+      currencyDropdownData.push({
+        content: image ? <TokenItem image={image} text={symbol} /> : symbol,
+        extraContent: balance,
+        onClick: !disabled
+          ? () => {
```
Mi-Lan, line 87: could this ternary operator be done without repeating whole arrow function? onClick: ()=>{!disabled?onChange(address):return}

liaojikunwork:
```
() => {
          if (!disabled) onChange(address)
        }
```
I think this is enought

app/src/hooks/useGraphMarketsFromQuestion.tsx
```diff
@@ -0,0 +1,66 @@
+import { useQuery } from '@apollo/react-hooks'
+import gql from 'graphql-tag'
+import { useEffect, useState } from 'react'
+
+import { Status } from '../util/types'
+
+const query = gql`
```
Mi-Lan, line 7: so i saw that we have almost the same useGraphMarketFromQuestion hook that just has id as response instead of id and collateral...so i don't know if it's maybe better idea to add one extra field in that query than to create new file just for one more piece of data...and query name should be changed not to be same as the other one...
Screenshot 2020-10-29 at 10 47 06

liaojikunwork: @Mi-Lan , the query is similar, but the purpose of each function is different logically.
So I created another similar one.

Mi-Lan: @liaojikunwork ok i just thought that this approach is not according to DRY principle and to maybe handle that difference within the one hook if it's 90% same code but i guess you already thought about it so all is good

liaojikunwork: @Mi-Lan , let me remove `useGraphMarketIdFromQuestion`

Mi-Lan (approved):

pimato (approved):

KadenZipfel (commented):
The code looks good, but a caveat I noticed with this method of implementation is that the back button won't work the same after you switch between currencies.

Issue #1257, opened 2020-10-29
Style Header buttons connect/Exit/create market/Address-Dropdown


pimato:
Currently:
<img width="137" alt="Bildschirmfoto 2020-10-29 um 08 24 51" src="https://user-images.githubusercontent.com/5337809/97538183-495a4500-19c0-11eb-93f0-ba381a9279a8.png">
<img width="352" alt="Bildschirmfoto 2020-10-29 um 08 24 47" src="https://user-images.githubusercontent.com/5337809/97538186-49f2db80-19c0-11eb-9fb8-ba921c12de72.png">
<img width="286" alt="Bildschirmfoto 2020-10-29 um 08 24 37" src="https://user-images.githubusercontent.com/5337809/97538187-4a8b7200-19c0-11eb-8434-e920eb7cb88e.png">

should:
<img width="1088" alt="Bildschirmfoto 2020-10-29 um 08 25 39" src="https://user-images.githubusercontent.com/5337809/97538218-570fca80-19c0-11eb-9dfb-b6358cbe4a45.png">

PR #1259, opened 2020-10-29
Style Header buttons connect/Exit/create market/Address-Dropdown #1257

liaojikunwork:


pimato (approved):

Mi-Lan (commented):
According to design left and right padding should be 17px instead of 20px and top and bottom paddings are not specified although they dont change much

liaojikunwork:
> According to design left and right padding should be 17px instead of 20px and top and bottom paddings are not specified although they dont change much

@Mi-Lan 
Currently, that's part of design and mobile responsive is implemented.
So I didn't change the paddings.

Mi-Lan (approved):

KadenZipfel (approved):

Issue #676, opened 2020-05-13
Include ETH as funding asset by using the gnosis proxy kit


pimato:
Currently there is no way to let users just use Ether as funding asset. To improve UX we should include that functionality.

PR #1262, opened 2020-10-30
wip: support native ether

hexyls:
Closes https://github.com/protofire/omen-exchange/issues/676

Includes native asset support for ethereum (mainnet, rinkeby), sokol, and xdai.

**xDAI**
Replaces: https://github.com/protofire/omen-exchange/pull/1464

PR does not include verification, we can add that later once the kleros/curate-xdai subgraph is up.

1. Add xDai as a new network to your wallet:
     Name: xDai
     RPC: https://rpc.xdaichain.com/
     Chain ID: 100
     Symbol: xDai
     Block explorer: https://blockscout.com/poa/xdai
2. Use https://bridge.xdaichain.com/ to convert dai on Ethereum to xDai
3. Use native xDai to fund/buy
4. Alternatively, send some xDai to the wrapped xDai address: 0xe91D153E0b41518A2Ce8Dd3D7944Fa863463a97d


pimato:
Hey @hexyls can you include Ether into the UI so I can test it?

for the time being, we should include both tokens Ether/Wrapped Ether in the the asset balance dropdown so the users can do whatever they want.

hexyls:
Funding with native ether should work at this point but does not.

Some notes:

- contract-proxy-kit [converts tx values to numbers](https://github.com/gnosis/contract-proxy-kit/blob/master/src/utils/transactions.ts#L72) which [ethers.js does not like](https://docs.ethers.io/v5/api/utils/bignumber/#BigNumber--notes-safenumbers).
- In my tests I confirmed that once a proxy's implementation is upgraded to Gnosis Safe v1.2.0 you can successfully send ether to the proxy and wrap it in a single transaction.
- However as soon as the multisend contract is involved contract-proxy-kit will estimate that the transaction will revert.

Will continue to explore...

hexyls:
Yep there is a problem with the MultiSend contract. Again it is the [same issue](https://github.com/gnosis/safe-contracts/issues/227) that both Gnosis Safe v1.1.1 and the current Proxy factory have. 

I deployed a new MultiSend contract on Rinkeby and was able to get multisend with ether wrapping working. 

However the conversion of tx values to numbers in contract-proxy-kit is still a problem.

hexyls:
Today I monkey patched the cpk library to fix the transaction value encoding issue.

Once [this proposal](https://alchemy-staging-rinkeby.herokuapp.com/dao/0x1eb3bcae24d42fde2f90e91d26313ac3946c4764/proposal/0x937297b31cdacd3daa684754751455e62a9478c6b7b7d063b5a54aff1c576a10) passes to add / index WETH markets on Rinkeby we should be ready to test market creation in native ether.

hexyls:
Today I added native ether funding for existing market pools. Market buys is next.

Some notes:

- You can deploy and upgrade a new proxy in a single transaction which is great. This makes the UX for funding with native ether equivalent to other assets where you must set an unlimited allowance first.

hexyls:
Native ether is now supported for create market, fund market, and buys on Rinkeby for existing users with a previously deployed proxy.

I discovered a new issue with deploying a new proxy and upgrading in a single transaction and will look into that next week.

app/src/components/market/common/upgrade_proxy/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+export const UpgradeProxy: React.FC<UpgradeProxyProps> = (props: UpgradeProxyProps) => {
+  const { context, finished, loading, upgradeProxy, ...restProps } = props
+  const nativeAsset = getNativeAsset(context.networkId)
+  const state = loading ? ButtonStates.working : finished ? ButtonStates.finished : ButtonStates.idle
+
+  return (
+    <Wrapper {...restProps}>
+      <Title>Enable {nativeAsset.symbol}</Title>
```
pimato, line 51: To keep the user experience in Omen consistent we should use the same Allowance component as we do with any other asset on Omen. This should be `Set Allowance`.

app/src/components/market/common/upgrade_proxy/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+      <DescriptionWrapper>
+        <Description>
+          This permission allows the smart contracts to interact with your {nativeAsset.symbol}.
```
pimato, line 54: `This permission allows Omen smart contracts to interact with your ETH. This has to bedone for each new token.`

pimato (changes requested):
The Set Allowance component should be consistent:
Bildschirmfoto 2020-12-11 um 11 49 23

KadenZipfel (commented):
Looks really good at first glance. Gonna do a deeper review later today when I'm feeling fresher, but a couple questions:

So generally this works by upgrading your CPK and when using ETH, it's just wrapped before being deposited, and thus the markets work the same as if they just accepted WETH?

Does the cpk update change your cpk address and/or affect any of your ability to participate with markets you have outcome shares or pool tokens?

hexyls:
Yeah that is exactly right, the cpk proxy address stays the same even after upgrading the master copy (the CPK library still uses the old mastercopy address to calculate the cpk proxy address under the hood). There shouldn't be a change to old markets after the upgrade 👍

KadenZipfel:
I'm trying to create a market with ETH, and think I've run into a bug: Instead of navigating to the market page after transaction confirmation, it's just hanging on load. Then if I go to the page manually with the `predictedMarketMakerAddress` that page also hangs, but if I find the contract on etherscan it seems to be deployed properly, so not sure what's happening.

app/src/util/networks.ts
```diff
@@ -67,6 +70,8 @@ interface KnownTokenData {
   order: number
 }
 
+export const pseudoNativeAssetAddress = '0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE'
```
KadenZipfel, line 73: Out of curiousity, is there any particular reason you capitalized the address like this?

KadenZipfel (approved):
Great work!

app/src/util/networks.ts
hexyls: I copied this from the 0x Project I think 🤔 and this is the checksum of that address, for us we could use any value here though cause we don't use the address in any transactions.

KadenZipfel (approved):
LGTM!

pimato (approved):

pimato:
I found issues with the asset dropdown ( https://github.com/protofire/omen-exchange/issues/1499 ) , but should be fixed in a separat PR! Looks good to me!

Mi-Lan (commented):
Looks good ! Tested xdai markets they work good, native eth works great also!
- I would suggest for the xDai instead of verify section having this error we hide the whole section for xDai this could be created as seprate issue so we publish this feature
Screenshot 2021-01-05 at 22 09 22
- Also i found that multiple currencies on xDai don't work, below is how it looks and I added xDai to it...
Screenshot 2021-01-05 at 22 14 20

Mi-Lan (approved):
LGTM!

PR #1290, opened 2020-11-02
Feature/#1278

Mi-Lan:


app/src/components/market/sections/market_create/steps/items/import_market_content.tsx
```diff
@@ -82,11 +82,16 @@ const CommentLabel = styled.div`
...
 
-const TitleValueVertical = styled(TitleValue)`
+const TitleValueVertical = styled(TitleValue)<{ verified?: boolean }>`
   flex-direction: column;
   justify-content: flex-start;
   text-transform: capitalize;
 
+  &:last-child {
+    p {
+      ${({ verified }) => (verified ? '' : `margin-top:-3px;`)};
```
liaojikunwork, line 92: Why is `margin-top: -3px` necessary?

liaojikunwork (changes requested):
see comments

liaojikunwork:
@Mi-Lan , about that Alert Icon, please export from Figma and replace it.
Current one is a bit big, I think.

app/src/components/market/sections/market_create/steps/items/import_market_content.tsx
Mi-Lan: @liaojikunwork well because in the difference in size of buttons that issue you mentioned on keybase with icon not vertically aligned should be fixed with this...

liaojikunwork: I know, but it's not a professional way to fix this issue.
Please export alert svg on Figma again( it will be 17*17) and replace svg file.
On mobile, icon size is different, so just setting `margin-top: -3px` is not a good way, I think.

app/src/components/market/common/additional_market_data/index.tsx
```diff
@@ -136,7 +136,7 @@ export const AdditionalMarketData: React.FC<Props> = props => {
           }
           isError={!verified}
         >
-          {verified ? <IconVerified size={'24'} /> : <IconAlert size={'24'} />}
+          {verified ? <IconVerified size={'24'} /> : <IconAlert size={'20'} />}
```
KadenZipfel, line 139: Why change the size here?

Mi-Lan: @KadenZipfel because with new svg size was not correct

KadenZipfel: Well if the size is passed in as a prop then you can set it as whatever size right? So if it's normally supposed to be `size={'24'}` then should it be kept the same?

Mi-Lan: @KadenZipfel you are right but svg size isn't same and dynamic viewBox size was also discarded

Mi-Lan: @KadenZipfel what do you suggest should i change it to 24 ?

Mi-Lan:
@KadenZipfel Here is how smaller one looks
Screenshot 2020-11-02 at 23 16 28
Here is how bigger one looks
Screenshot 2020-11-02 at 23 17 27

app/src/components/market/common/additional_market_data/index.tsx
KadenZipfel: I would set it as 24 since it was the previous size used, but I agree it does look a little off. See if you can find the size used for that icon in that spot on figma and go with whatever is used there

liaojikunwork:
Screen Shot 2020-11-02 at 7 14 18 PM
Screen Shot 2020-11-02 at 7 14 27 PM

@Mi-Lan , when you export icons on Figma, please select the group(not the vector only).
And the size of icons will be 20

Mi-Lan:
> Screen Shot 2020-11-02 at 7 14 18 PM
> Screen Shot 2020-11-02 at 7 14 27 PM
> 
> @Mi-Lan , when you export icons on Figma, please select the group(not the vector only).
> And the size of icons will be 20

updated that and put 20 in size since that seems to be more according to figma.https://www.figma.com/file/R056tSYMjvfwaFidEBItYy/Widgets-V2?node-id=1209%3A141

liaojikunwork (approved):

pimato:
Hey @Mi-Lan ,
I just checked my designs and it seems they were outdated, sorry for that.
I have updated the designs now and hopefully we have a more consistent visualisation of those meta icons.

**Figma**:
[Design Specification](https://www.figma.com/file/R056tSYMjvfwaFidEBItYy/Widgets-V2?node-id=1209%3A141)

pimato:
Hey Milan,

the non-verified and verified icons looking great. Only the Kleros icons needs to be updated to 24x24px as it is currently:
Bildschirmfoto 2020-11-04 um 08 25 09
This new version of it is a 24x24px:
[Kleros.svg.zip](https://github.com/protofire/omen-exchange/files/5486127/Kleros.svg.zip)

pimato (changes requested):
see comments

pimato:
margin should be reduced to 8px, currently 10px:
Bildschirmfoto 2020-11-04 um 13 28 11

pimato (changes requested):
see comments

pimato (approved):

liaojikunwork (approved):

KadenZipfel (approved):

PR #1292, opened 2020-11-02
changed more button so the dropdown is on the right for mobile

Mi-Lan:


app/src/components/common/form/more_menu/index.tsx
```diff
@@ -38,6 +38,9 @@ const ItemsContainer = styled.div<{ isOpen: boolean }>`
   display: ${({ isOpen }) => (isOpen ? 'block' : 'none')};
   padding: 8px;
   border-radius: 12px;
+  @media (max-width: ${props => props.theme.themeBreakPoints.md}) {
+    right: -123px;
```
liaojikunwork, line 42: This is not a good way to fix this issue.

Instead of `right: -123px`, what about using the following one?

```
right: unset;
left: 0;
```

liaojikunwork (changes requested):
See comments

app/src/components/common/form/more_menu/index.tsx
Mi-Lan: @liaojikunwork ohh didn't know it could be done that way! Will change!

liaojikunwork (approved):

KadenZipfel (approved):

pimato:
on chrome I still see this:
Bildschirmfoto 2020-11-02 um 19 44 31

Mi-Lan:
> on chrome I still see this:
> Bildschirmfoto 2020-11-02 um 19 44 31

@pimato aha but this issue is regarding dropdown of three dots not the layout...maybe we can create separate issue for that?

pimato:
> > on chrome I still see this:
> > Bildschirmfoto 2020-11-02 um 19 44 31
> 
> @pimato aha but this issue is regarding dropdown being out of view not the layout...maybe we can create separate issue for that?

agreed, added here: https://github.com/protofire/omen-exchange/issues/1296

pimato (approved):

PR #1295, opened 2020-11-03
Fix/1272 ui not updating after transaction

liaojikunwork:


pimato:
@liaojikunwork after a user is done with the buy/sell deposit/withdraw, is total liquidity and total volume/24 volume getting updated without a page refresh?

I tried it with buy/sell and deposit/withdraw and it does not look like it:
Bildschirmfoto 2020-11-03 um 21 22 44

Bildschirmfoto 2020-11-03 um 21 23 36

on rinkeby!

pimato (changes requested):
see comments

Mi-Lan:
> @liaojikunwork after a user is done with the buy/sell deposit/withdraw, is total liquidity and total volume/24 volume getting updated without a page refresh?
> 
> I tried it with buy/sell and deposit/withdraw and it does not look like it:
> Bildschirmfoto 2020-11-03 um 21 22 44
> 
> Bildschirmfoto 2020-11-03 um 21 23 36
> 
> ISSUE/BUG: The volume has not been updated
> 
> on rinkeby!

yeah for me also on rinkeby it doesn't update the values for liqudity and volume..haven't tested on mainnet though might be issue with rinkeby...And pool section seems to work good

liaojikunwork:
@pimato , @Mi-Lan , after pulling my PR, could you do hard-refresh and check pls?

Mi-Lan:
> @pimato , @Mi-Lan , after pulling my PR, could you do hard-refresh and check pls?

it seems that for this question nothing is updating for some time now http://localhost:3000/#/0x801d77068666ee9eec3074425456e7ae0376ca48

@liaojikunwork  but with other question (http://localhost:3000/#/0x901c1c45ba009f67726fb22dcf39a7ebc424ee32) volume updates good but liquidity still needs refresh

liaojikunwork:
> still

Is Liquidity updated after you reload the page?

liaojikunwork:
> > @pimato , @Mi-Lan , after pulling my PR, could you do hard-refresh and check pls?
> 
> it seems that for this question nothing is updating for some time now http://localhost:3000/#/0x801d77068666ee9eec3074425456e7ae0376ca48
> 
> @liaojikunwork but with other question (http://localhost:3000/#/0x901c1c45ba009f67726fb22dcf39a7ebc424ee32) volume updates good but liquidity still needs refresh

@Mi-Lan  is Liquidity updated after you reload the page?

pimato (approved):

KadenZipfel:
Just tried a transaction and I'm stuck in this state a couple minutes after the transaction has been confirmed
image

Mi-Lan:
> Just tried a transaction and I'm stuck in this state a couple minutes after the transaction has been confirmed
> image

yeah it also seemed to me like loading screen was longer there... and it got away as soon as liqudity state changed

liaojikunwork:
> Just tried a transaction and I'm stuck in this state a couple minutes after the transaction has been confirmed
> image

@KadenZipfel , if input value is less than balance, `sell` button should be disabled, right?

Mi-Lan (approved):

pimato (approved):

app/src/util/tools.ts
```diff
@@ -361,3 +361,39 @@ export const formatToShortNumber = (number: string, decimals = 2): string => {
...
+  if (keys1.length !== keys2.length) return false
+
+  for (let keyIndex = 0; keyIndex < keys1.length; keyIndex += 1) {
+    const key = keys1[keyIndex]
+    if (typeof obj1[key] !== typeof obj2[key]) return false
+    if (!isObjectEqual(obj1[key], obj2[key])) return false
+  }
+
+  return true
+}
```
KadenZipfel, line 394: What's the source of this function? It seems to work properly, but I'd be more comfortable knowing that it's battle tested

KadenZipfel (commented):
Overall it seems to be a good implementation, and it seems The Graph limits us in terms of options so using `setTimeout` is probably our best option. So long as `isObjectEqual` is safe to use, I'm comfortable merging this

app/src/util/tools.ts
liaojikunwork: @KadenZipfel , I built this function myself and tested in many cases.

Mi-Lan (approved):

app/src/util/tools.ts
KadenZipfel: I would say that we should find a battle tested version of this, but now that I think of it, in the very unlikely case that there's an exception which fails, the worst case scenario would be very minor so I think it's fine.

KadenZipfel (approved):

Issue #1313, opened 2020-11-05
Updating the CPK library version to v2.2.0-alpha.1


pimato:
The Omen squad/multisig is not able to operate with Omen (create and provide liquidity to markets) because the current CPK does not allow us to add Omen as a gnosis safe application on the gnosis safe interface https://gnosis-safe.io/ . This should be fixed with the newest alpha release [here](https://github.com/gnosis/contract-proxy-kit/releases/tag/v2.2.0-alpha.1). Goal of the update is to get Omen running as a gnosis-safe application.

PR #1317, opened 2020-11-06
Update contract-proxy-kit to v2.2.0-alpha.1

hexyls:
Closes https://github.com/protofire/omen-exchange/issues/1313


liaojikunwork:
@hexyls , all checks have failed.
Could you resolve pls?

liaojikunwork:
@hexyls , still checks have been failed.

Could you try "npm run test" locally and push?

hexyls:
Thanks for the heads up @liaojikunwork! All test are passing locally so that is ok.

The checks are saying that it "Cannot find module 'levenary'" during the build / deploy process which is strange because I don't get that error when building locally. I will add levenary as a dependency and see if that helps.

hexyls:
That seems to have done it!

Thanks for your help @liaojikunwork 👌

liaojikunwork (approved):

cag:
Hey, I've looked a bit more into this and I think this is a good first step towards resolving #1313. However, in order to use this as a Safe App, some [additional instructions](https://github.com/gnosis/safe-apps-sdk#testing-in-the-safe-multisig-application) have to be followed. I found that changing the `manifest.json` to contain the following content allowed the app to be added as a Safe App:

```json
{
  "name": "Omen",
  "description": "Decentralized Prediction Markets on Ethereum",
  "iconPath": "omen_logo_thumbnail.png",
  "providedBy": { "name": "DXdao", "url": "https://dxdao.eth/" }
}
```

This change isn't enough though. I haven't confirmed why, but I have a working hypothesis that involves race conditions (yay). Anyway, this might be a bug with the CPK.

hexyls:
Thanks heaps for the info @cag!

I took a look at this today and I think this is because we create many CPK instances in Omen. 

I think config.listeners in safe-apps-sdk is [being overridden](https://github.com/gnosis/safe-apps-sdk/blob/master/src/index.ts#L142) by newer instances of CPK, so old instances are not getting the onSafeInfo update. 

If this is the case we can probably resolve this by using a single instance of CPK across the app.

cag:
Yeah, that actually makes a lot of sense! Actually the fact that the CPK gets constantly recreated has always bothered me about this app a bit, so this would be a good fix to have. Not sure if that's in the scope of this PR though, but since the issue mentions getting Safe App integration as a goal, maybe? Up to you.

pimato:
> Thanks heaps for the info @cag!
> 
> I took a look at this today and I think this is because we create many CPK instances in Omen.
> 
> I think config.listeners in safe-apps-sdk is [being overridden](https://github.com/gnosis/safe-apps-sdk/blob/master/src/index.ts#L142) by newer instances of CPK, so old instances are not getting the onSafeInfo update.
> 
> If this is the case we can probably resolve this by using a single instance of CPK across the app.

Sounds like a solution! Lets go with it!

hexyls:
@cag awesome, yep will add to this PR

hexyls:
Alright this is ready to go! The netlify deploy preview blocks cross site requests so I deployed this build to https://omen-gamma.vercel.app/ and you can use that instead. Try it out by adding it as an app to https://rinkeby.gnosis-safe.io/

pimato:
so I´ve played around with https://omen-gamma.vercel.app/ on Mainnet (https://gnosis-safe.io/) and on Rinkeby (https://rinkeby.gnosis-safe.io/).

On Rinkeby I am getting this just a loading indicator:
Bildschirmfoto 2020-11-12 um 12 14 35

On Mainnet it seems to work, but I can`t connect through the gnosis safe web3provider, rather it allows me only to connect throughout metamask/walletconnect. I reached out to the Gnosis Safe team and it looks like, there is a cpk functions which allows us to detect if it is a gnosis safe:
Bildschirmfoto 2020-11-12 um 12 11 32

pimato (dismissed):
see comment

KadenZipfel (approved):

hexyls:
Some notes after a convo with @pimato in keybase

What works now:
1. Sign into the gnosis-safe interface, create a safe
2. Add Omen as an app, sign in as the same account that created the safe
3. When you make transactions that involve a proxy, Omen correctly uses the gnosis-safe proxy instead of the CPK proxy

What we want:
1. Sign into the Gnosis Safe interface, create a safe
2. Add Omen as an app, Omen auto signs in from the perspective of the gnosis-safe proxy (not the owner of the proxy)
3. When you make transactions you are making them from owner of the proxy (through the SDK) but from the perspective of the proxy

We'll need to:
1. Add a Safe connector that can integrate with web3-react (I started looking into this on Friday)
2. Add some checks in various places that assume the perspective of the proxy owner not the proxy itself

hexyls:
Added a Safe connector today, only tested on Rinkeby so far. The gnosis gas estimator is being funky, so if you want to test this out make sure you add a ~400,000 gas limit for transactions to be successful.

pimato:
@hexyls it looks like we are using an outdated buggy version of web3-react(`^5.0.4`). We should update the library first and see if there are still issues with the connector.

hexyls:
We are blocked on this until this issue is resolved: https://github.com/gnosis/safe-react/issues/1512

Wallets like Metamask are currently vastly underestimating how much gas a transaction needs because of the internals of the gnosis safe smart contracts.

A way around this would be to hardcode gas limits when Omen is running as a Safe App, however this is not possible until safe-react is updated, followed by safe-apps-sdk and finally CPK.

hexyls:
It looks like the next release of safe-apps-sdk will include a safeTxGas param we can potentially use to force correct gas estimations: https://github.com/gnosis/safe-apps-sdk/pull/52

hexyls:
Figured out how to monkey patch CPK to support `SEND_TRANSACTIONS_V2` so we don't have to wait for the next release of safe-apps-sdk.

Tomorrow I will go through and test this properly 👌

pimato:
Hey @hexyls, just tried to create a market:
Bildschirmfoto 2020-11-18 um 12 47 33

once I clicked on `Create Market`:
Bildschirmfoto 2020-11-18 um 12 47 42

Note: It looks like the gnosis safe app is getting our `txOption` but it signals that the estimation fails.

I´ve set the gas estimation to `500000` on Metamask:

Bildschirmfoto 2020-11-18 um 12 48 13

Bildschirmfoto 2020-11-18 um 12 48 17

but it sadly did not work:
Bildschirmfoto 2020-11-18 um 12 57 20

this is the reverted transaction:
https://etherscan.io/tx/0xbcabc49c1438faaa57a17d92d28982eb17ee097ad57bd287751671c27cb032d6/advanced

hexyls:
@pimato thanks for trying it out and yeah this is expected behaviour.

There are a few things going on here:

- safeTxGas is a hard minimum gas value enforced by the smart contract, anything lower will revert the transaction
- We have to set safeTxGas because:
    1. There is a bug within the gnosis interface that causes gas estimations fail (this is why it says "Error while estimating gas. The transaction may fail." in your screenshot above, this is a bug and the transaction will not fail if it has enough gas)
    2. Failed gas estimations lead to wallet providers doing their own gas estimations which are too low and cause internal transactions to fail
    3. Setting safeTxGas forces gas estimations from wallet providers to meet a minimum gas requirement

hexyls:
Setting safeTxGas does work well on Metamask, however I have found on Rainbow Wallet that they do not always respect this value when estimating gas limits for transactions. This is the best we will get until the Gnosis interface is updated to allow us to set gas limits directly.

hexyls:
Just a reminded of how to test this branch:

Once you have created a safe on https://rinkeby.gnosis-safe.io/ you can add this build as an app using https://omen-gamma.vercel.app

The app auto signs in from the perspective of the safe rather than the owner of the safe, so to buy / sell etc you will need to send funds to the safe itself first.

pimato:
Hey @hexyls,
the omen squad is currently trying to withdraw liquidity from this market: https://omen.eth.link/#/0x88fabdbc004f507cc796f170f87a6352b5db3ce0

but it no markets are showing up on the my markets view. Also no pool shares are showing up in the withdraw section in the market details pool view. 

Interestingly, if we use this PR from Alan (https://github.com/protofire/omen-exchange/pull/1317) via wallet connect everything shows up but the safeTxgas is not configured so a lot of transactions are out of gas errors.

We need to withdraw the liquidity as soon as possible.

app/src/components/market/sections/market_verify/market_verify.tsx
```diff
@@ -100,6 +100,8 @@ const MarketVerifyWrapper: React.FC<Props> = (props: Props) => {
 
   if (!loading && errorMessage) return <GenericError>{errorMessage || 'Failed to fetch curation data'}</GenericError>
 
+  const verificationBtnDisabled =
+    loading || typeof selection !== 'number' || !ovmAddress || verificationState != 1 || (cpk && cpk.cpk.isSafeApp())
```
KadenZipfel, line 104: Why can't we verify with the safe app?

app/src/services/cpk.ts
```diff
@@ -1,3 +1,4 @@
+import { txs } from '@gnosis.pm/safe-apps-sdk/dist/txs'
```
KadenZipfel, line 1: What do you think about creating a separate service file for safe transactions instead of modifying the cpk service?

app/src/components/market/sections/market_create/steps/items/funding_and_fee_step.tsx
```diff
@@ -279,7 +279,8 @@ const FundingAndFeeStep: React.FC<Props> = (props: Props) => {
     isNegativeDepositAmount
 
   const showSetAllowance =
-    allowanceFinished || hasZeroAllowance === Ternary.True || hasEnoughAllowance === Ternary.False
+    !cpk?.cpk.isSafeApp() &&
```
KadenZipfel, line 282: Do you not need to set allowance in the safe app?

app/src/components/market/sections/market_verify/market_verify.tsx
hexyls: `requestVerification` does [not use the CPK](https://github.com/protofire/omen-exchange/blob/master/app/src/services/cpk.ts#L425) but rather the signer/provider directly. When running as a safe app we don't have direct access to the signer.

app/src/components/market/sections/market_create/steps/items/funding_and_fee_step.tsx
hexyls: No the allowance is usually to allow the cpk to transfer user funds. The safe is the user here and is using its own funds.

app/src/services/cpk.ts
hexyls: Open to it for sure but not sure what the benefit would be?

KadenZipfel: Yeah, now looking at the code again I don't think it's quite necessary

Mi-Lan (commented):
Great job! I just found that when i want to request verification via gnosis safe interface it doesen't work...what could this be?
@hexyls 
Screenshot 2020-12-03 at 19 31 37

hexyls:
Thanks for the review @Mi-Lan you have a good eye! [Check out this convo](https://github.com/protofire/omen-exchange/pull/1317#pullrequestreview-542386760) above for some context on why this is disabled 👍

Mi-Lan (approved):

pimato (approved):

Issue #1271, opened 2020-10-31
My Markets Ending soon is showing the soonest ending markets at the end of the list


pimato:
last page for me:
<img width="624" alt="Bildschirmfoto 2020-10-31 um 22 31 08" src="https://user-images.githubusercontent.com/5337809/97790451-f3a4b900-1bc8-11eb-8158-1db1b6954b56.png">

first page for me:
<img width="617" alt="Bildschirmfoto 2020-10-31 um 22 32 53" src="https://user-images.githubusercontent.com/5337809/97790462-0919e300-1bc9-11eb-9bf2-12855b2c20f6.png">

KadenZipfel:
So I guess intended effect would be to not display any markets which are already closed?

liaojikunwork:
@pimato , do you want to list soonest ending markets at first, and closed markets at the end?

pimato:
> @pimato , do you want to list soonest ending markets at first, and closed markets at the end?

yeah it would definitely be an improvement.

> So I guess intended effect would be to not display any markets which are already closed?

if we removed the closed markets users can`t see in which markets he/she participated with.

liaojikunwork:
@pimato , so on this issue, I should do that improvement only?

liaojikunwork:
@pimato , I checked the code.
Currently, markets are sorted by openingTimestamp.
Should update to be sorted by closingTimestamp? (only for my markets, or for all options)

pimato:
> @pimato , I checked the code.
> Currently, markets are sorted by openingTimestamp.
> Should update to be sorted by closingTimestamp? (only for my markets, or for all options)

We should change the sorting only for the `my markets - view`. Will we see markets which are not closed yet at the beginning with it?

PR #1330, opened 2020-11-10
My Markets Ending soon is showing the soonest ending markets at the end of the list #1271

liaojikunwork:


pimato (approved):
it is def. an improvement as now active markets are showing first.

Mi-Lan (approved):

KadenZipfel (approved):
The code looks good, but the functionality is a bit misleading. What is expected is to see the markets that are closing soonest but really you see the markets which have the latest closing date. This being said, previously we were displaying already closed markets first, so maybe this is the lesser of two evils considering the limitations of the subgraph.

Issue #1352, opened 2020-11-13
Currency dropdown broken after replacing market address


pimato:
Steps to reproduce:

1. Go to https://dxtest.eth.link/#/0x06c942fd737bc53571a34667b2ae3b669f1f502d
2. Replace the market address in the url with a market which is not open like this `0xa776a48b4a62d57da6084c20858e31e654c829d0` 
3. BUG:
<img width="621" alt="Bildschirmfoto 2020-11-13 um 15 15 57" src="https://user-images.githubusercontent.com/5337809/99082194-1caf4b80-25c4-11eb-9f53-cf0b2e498b7c.png">
<img width="151" alt="Bildschirmfoto 2020-11-13 um 15 15 41" src="https://user-images.githubusercontent.com/5337809/99082199-1e790f00-25c4-11eb-906c-7b6e2325ef05.png">

PR #1363, opened 2020-11-16
Currency dropdown broken after replacing market address #1352

liaojikunwork:


pimato (approved):

Mi-Lan (approved):

app/src/hooks/useGraphMarketsFromQuestion.tsx
```diff
@@ -50,13 +51,10 @@ export const useGraphMarketsFromQuestion = (questionId: string): Result => {
     if (!questionId) setMarkets([])
   }, [questionId])
 
-  if (data && data.question && data.question.indexedFixedProductMarketMakers.length > 0 && markets.length === 0) {
```
KadenZipfel, line 53: Don't we want to leave this check: `data.question.indexedFixedProductMarketMakers.length > 0` so we can error if no markets are returned?

app/src/hooks/useGraphMarketsFromQuestion.tsx
```diff
@@ -50,13 +51,10 @@ export const useGraphMarketsFromQuestion = (questionId: string): Result => {
...
+  if (data && data.question && !isObjectEqual(markets, data.question.indexedFixedProductMarketMakers)) {
     setMarkets(data.question.indexedFixedProductMarketMakers)
   } else if (data && data.question && !data.question.indexedFixedProductMarketMakers.length) {
-    return {
-      markets: [],
-      status: Status.Error,
-    }
+    setMarkets([])
```
KadenZipfel, line 57: Shouldn't we still return an error here?

liaojikunwork: @KadenZipfel , this section doesn't mean error.
It means "no markets" is returned.

app/src/hooks/useGraphMarketsFromQuestion.tsx
liaojikunwork: @KadenZipfel , if no markets are returned, indexedFixedProductMarketMakers will be an empty array and isObjectEqual will return true.
Don't need to check `data.question.indexedFixedProductMarketMakers.length > 0`

KadenZipfel (approved):

PR #1374, opened 2020-11-17
changed styles of verification process icons

Mi-Lan:


pimato:
This:
Bildschirmfoto 2020-11-17 um 19 10 41

Should be a hyperlink to the kleros curate market page

pimato (changes requested):
see comments

pimato:
Also, can we have the onHover only when the user is actually hovering above the text. Currently the hover effect shows also if I am far away from it:
IMG_4353

pimato (approved):

liaojikunwork (approved):

app/src/components/market/sections/market_verify/option/kleros_curation.tsx
```diff
@@ -131,6 +164,10 @@ export const KlerosCuration: FC<Props> = (props: Props) => {
   let KlerosNotice
   let KlerosRightColumn
 
+  const urlSwitch = (url: string) => {
+    const win = window.open(url, '_blank')
+    win && win.focus()
+  }
```
KadenZipfel, line 170: The problem with this approach is that [browser support is unknown](https://caniuse.com/?search=window.open). I think we're better off staying on the safe side and using what we know works for sure

KadenZipfel (changes requested):

KadenZipfel (approved):

PR #1384, opened 2020-11-19
Scalar market sell

KadenZipfel:


pimato:
I´ve added comments to the scalar markets buy PR: feature/scalar-market-buy which should be fixed first.

pimato:
Is there a reason why the playing with the slider is not working here?

KadenZipfel:
> Is there a reason why the playing with the slider is not working here?

Yeah, I wasn't sure how it would be used in the sell view so I just messily added it in for now. Still a good amount of changes to be made on this PR

KadenZipfel:
So the last thing (hopefully) that needs to be modified for this PR is the slider. Personally, I think it should just show the current prediction and the new prediction (based on the amount being sold), and I think it should also be disabled, i.e. not draggable. @pimato what do you think?

pimato:
> So the last thing (hopefully) that needs to be modified for this PR is the slider. Personally, I think it should just show the current prediction and the new prediction (based on the amount being sold), and I think it should also be disabled, i.e. not draggable. @pimato what do you think?

having the slider-prayground here could allow the user to evaluate to really close his position or not. It will also allow him/her to see the results for question "if I don´t sell now, how much can I earn more?".

What do you think?

KadenZipfel:
> having the slider-prayground here could allow the user to evaluate to really close his position or not. It will also allow him/her to see the results for question "if I don´t sell now, how much can I earn more?".
> 
> What do you think?

I agree. Since this will require trade data, and be a little more complex to integrate, I've created this issue https://github.com/protofire/omen-exchange/issues/1422 to integrate it in v1.2.2. In the meantime I think the best bet would be to have a more minimal implementation with just the current and new prediction. Sound good?

hexyls (approved):

app/src/theme/overrides/react_tooltip_styles_override.tsx
```diff
@@ -77,4 +77,25 @@ export const ReactTooltipStylesOverride = css`
...
+      letter-spacing: 0.1px;
+      text-align: left;
+      background-color: ${theme.colors.mainBodyBackground};
+      opacity: 1;
+      border-radius: ${theme.borders.commonBorderRadius};
+      box-shadow: 0px 0px 4px rgba(0, 0, 0, 0.05);
+      border: 1px solid ${theme.borders.tooltip};
+      padding: 5px 8px;
+    }
+  }
```
Mi-Lan, line 100: @KadenZipfel seems that this class is same as defaultTooltip made above just the padding difference, we should probably reuse that

app/src/components/market/sections/market_sell/scalar_market_sell.tsx
```diff
@@ -1,12 +1,340 @@
...
+        positionIndex,
+        amountToSell.mul(-1),
+        amountShares.mul(-1),
+      )
+
+      const pricesAfterTrade = MarketMakerService.getActualPrice(balanceAfterTrade)
+      const potentialValue = mulBN(amountToSell, 1 / (1 - marketFeeWithTwoDecimals))
+      const costFee = potentialValue.sub(amountToSell)
+
+      const newPrediction = pricesAfterTrade[1] * ((upperBound || 0) - (lowerBound || 0)) + (lowerBound || 0)
```
Mi-Lan, line 134: @KadenZipfel i was looking at code and it seems that calcSellAmount function for scalar market and in  market_sell is almost identical...could this function maybe be exported and that change accounted for like some extra parameter which handles that small difference?

Mi-Lan (changes requested):
In general works great! Also would be good to have at least figma link if issue is not linked...

app/src/components/market/sections/market_sell/scalar_market_sell.tsx
KadenZipfel: Good catch. I've outsourced it to this issue: https://github.com/protofire/omen-exchange/issues/1429 for the next release to ensure we can get scalar markets out for v1.2.1.

Mi-Lan (approved):
GJ!

pimato (approved):

PR #1385, opened 2020-11-19
Scalar market pool

KadenZipfel:


pimato (changes requested):
Button withdraw is disabled because I don`t have any pool tokens, but after clicking Max button the withdraw button is enabled, see:
Bildschirmvideo-aufnehmen-2020-1

pimato:
Also, it seems like the DAI Description in the balance component is jumping around from right to left:
Bildschirmvideo-aufnehmen-2020-1 (1)

KadenZipfel:
> Also, it seems like the DAI Description in the balance component is jumping around from right to left:
> Bildschirmvideo-aufnehmen-2020-1 (1)

I'm pretty certain that this is only because there's no icon for the tokens on rinkeby, so on mainnet it shouldn't be a problem.

hexyls (approved):
👍

app/src/theme/overrides/react_tooltip_styles_override.tsx
```diff
@@ -77,4 +77,25 @@ export const ReactTooltipStylesOverride = css`
...
+      letter-spacing: 0.1px;
+      text-align: left;
+      background-color: ${theme.colors.mainBodyBackground};
+      opacity: 1;
+      border-radius: ${theme.borders.commonBorderRadius};
+      box-shadow: 0px 0px 4px rgba(0, 0, 0, 0.05);
+      border: 1px solid ${theme.borders.tooltip};
+      padding: 5px 8px;
+    }
+  }
```
Mi-Lan, line 100: @KadenZipfel seems that we should delete this class with tooltip since it is defined above in the dry fashion...

Mi-Lan (changes requested):
Just this small change!

Mi-Lan (approved):
LGTM!

pimato:
So it looks like I found a reproducable bug:

1) Create Market and fund it with USDC:
Bildschirmfoto 2020-12-22 um 21 28 57

2) Deposit 50 USDC 

3) Withdraw all

4) Deposit 50 USDC again

CRASH --> 
Bildschirmfoto 2020-12-22 um 21 36 43

pimato (changes requested):
see comment

pimato (approved):
works!

PR #1399, opened 2020-11-25
button styling seems to be overwritten

Mi-Lan:


app/src/components/market/sections/market_verify/option/kleros_curation.tsx
```diff
@@ -238,7 +238,7 @@ export const KlerosCuration: FC<Props> = (props: Props) => {
             successful challenge.
           </DescriptionText>
           <RightButtonWrapper>
-            <RightButton buttonType={ButtonType.secondary}>
```
KadenZipfel, line 241: Looks like there's also an instance of `<RightButton buttonType={ButtonType.secondary}>` at line 291. Is it the correct button type for that button?

Mi-Lan: changed it! good catch!

pimato (approved):

KadenZipfel (approved):

Issue #1393, opened 2020-11-23
Make positions visible in market details view


pimato:
With the current Omen interface users don`t have a way to observe their market positions:

![market-details](https://user-images.githubusercontent.com/5337809/100005524-5dc80c80-2dc9-11eb-9609-53108c8d22d8.png)

Notes: 
- The table should only be visible if user has > 0 positions.

**Figma:**
[Design Specification
](https://www.figma.com/file/R056tSYMjvfwaFidEBItYy/Widgets-V2?node-id=1209%3A141)

PR #1411, opened 2020-12-03
Scalar market details

KadenZipfel:
Closes: https://github.com/protofire/omen-exchange/issues/1393


pimato:
When I do a trade and go back, I need to refresh the page to see my newly bought position. But everything else looks great!

hexyls:
Yeah I think it would be cool to update everything simultaneously if possible, even if it comes with a slight delay. Thoughts?

**Sliders**
 - [x] pool
 - [x] create
 - [ ] buy
 - [ ]  sell
 
I've asked Kaden on Keybase to school me on how prediction calculations for buy and sell sliders work. So will come back to those soon.

- [x] Correct payout/profit/loss values (as far as I can tell)

hexyls (approved):

KadenZipfel:
> I did notice that the transaction details table did not fully clear after a buy, here is 0.04 DAI left over in max loss / max profit after a purchase:
> Screen Shot 2020-12-21 at 1 43 13 pm

Yeah, glad you noticed. That is fixed in: https://github.com/protofire/omen-exchange/pull/1457

> Random thing I noticed on line 377 in market_scale (I wish github would let me comment on any line 🙄):
> 
> ```
> setTotalShortPrice(
>    trades
>      .filter(trade => trade.outcomeIndex === '0')
>      .map(trade => Number(formatBigNumber(trade.collateralAmount, collateral.decimals, collateral.decimals)))
>      .reduce((a, b) => a + b),
> )
> ```
> 
> If that filter removes all trades the following reduction will cause a "Reduce of empty array with no initial value" error.

Good find! I'll add check to prevent that error.

app/src/theme/overrides/react_tooltip_styles_override.tsx
```diff
@@ -77,4 +77,25 @@ export const ReactTooltipStylesOverride = css`
...
+      letter-spacing: 0.1px;
+      text-align: left;
+      background-color: ${theme.colors.mainBodyBackground};
+      opacity: 1;
+      border-radius: ${theme.borders.commonBorderRadius};
+      box-shadow: 0px 0px 4px rgba(0, 0, 0, 0.05);
+      border: 1px solid ${theme.borders.tooltip};
+      padding: 5px 8px;
+    }
+  }
```
Mi-Lan, line 100: @KadenZipfel you are gonna kill me 🗡️

Mi-Lan (changes requested):

Mi-Lan (changes requested):
Screenshot 2020-12-22 at 21 41 07
seems that this shouldn't be here ?

pimato:
After market creation, there is a missing divider:
Bildschirmfoto 2021-01-05 um 22 47 07

pimato:
Also, there is now a missing divider between the slider and the order form too:
Bildschirmfoto 2021-01-05 um 22 50 55

KadenZipfel:
The problem that I keep coming back to without managing to find a quick solution is that liquidity provision mixes with trading in a way that the subgraph isn't yet equipped, causing pretty big bugs.

If you provide liquidity at a non-5050 probability, you're buying outcome shares on one side behind the scenes. Since you have not actually done any buy transactions, the position table doesn't yet render, but if you do buy some more of those shares, your profit will be calculated based on the valuation of all the shares you own minus just the price that you paid to buy those additional shares. This can even lead to the total cost being counted as negative, which results in profit/loss to be <-100% (which is obviously impossible in reality).

Notice how both positions show green profit/loss and a positive absolute profit yet <-100%
image

**Possible solutions:**
1. We can hide the table, if the user has provided liquidity in that market.
    - Caveat: Users that have provided liquidity and are used to this flow might think that there is a bug, or even that their position has been liquidated somehow.
2. We can remove the profit/loss column
3. We can update the subgraph to track behind the scenes purchases of outcome shares

Personally I think we should do a combination of 2 and 3. No profit/loss column for now, and if we can get the necessary subgraph update, we can reintroduce the column.

pimato:
I agree! 

Lets move ahead with:
2. Removing the profit/loss column until 3. is implemented and live on the subgraph
3. Create issues which describes clearly what needs to be added on the omen-subgraph repo.

Mi-Lan (approved):

pimato (approved):

Issue #1326, opened 2020-11-08
Infinite loading of newly created markets


KadenZipfel:
After creating a market, when we're redirected to the market page it hangs indefinitely. I assume this is because the subgraph data is not available when we initially load the market. To fix this it might be best to wait a certain amount of time, and if the market still hasn't loaded, refetch the subgraph data.

PR #1413, opened 2020-12-04
fix/#1326

hexyls:
closes https://github.com/protofire/omen-exchange/issues/1326


pimato (approved):

KadenZipfel (approved):

Mi-Lan (approved):
@hexyls Very useful feature! I see also that this can be implemented for useKlerosCuration component...but not scope of this pr.

Issue #1390, opened 2020-11-23
Enhanced My Market View


pimato:
![my-market](https://user-images.githubusercontent.com/5337809/99964590-d2805400-2d93-11eb-9e7f-a50ae8efbe8c.png)

The current my market view on Omen is not sufficient for keeping track of market creations, liquidity provision and shares overview. This Bounty addresses several important improvements:

- **A read-only widget which shows:** 
  - the user address or ENS
  - An overview of total won/lost predictions
  - An overview of total amount won/lost in predictions

- Separate user owned shares overview with pool tokens/liquidity provision overview.
- sort list of markets which users owns shares of by:
  - Closing soon
  - Highest Shares
- sort list of markets users provides liquidity by:
  - Closing soon
  - Highest liquidity
  - Highest profit

- The list item of the `My Shares` should include the total amount of Shares the users owns of the market.
- The list item of the `My Liquidity` should include the total amount of liquidity and the earnings of the market.

**Figma:**
[Playground](https://www.figma.com/proto/aTivxuHmp2l6Sn8nho1Ngu/Playground?node-id=697%3A1&viewport=352%2C429%2C0.6951354742050171&scaling=min-zoom)
[Design Specification](https://www.figma.com/file/aTivxuHmp2l6Sn8nho1Ngu/Playground?node-id=697%3A1)

anon-good-boy:
commenting to claim/work on this issue

PR #1414, opened 2020-12-04
feature/#1390

anon-good-boy:
PR for #1390


anon-good-boy:
Pushing initial (early) draft of this feature. 

Thus far, I have modified useMarketMakerData to expose an additional hook that fetches all of my (current user) markets along with all of my holdings from those markets. I did this by mostly updating/leveraging the existing `useBlockchainMarketMakerData` and the `useGraphMarketMakerData` hooks.

PR #1417, opened 2020-12-08
Update "Profit" to "Revenue"

alexvansande:
Using "profit" here is misleading and can lead to a very bad experience for the user. "Profit" implies the net result from your trade, meaning that if you bought at X, you sell at Y and your revenue is (Y-X). If you are selling at a loss, because you bought the odds disfavorably, then profit would be negative. In fact seeing the word "profit" might lead someone to think that they can still turn a profit, encourage them to sell it and they will be very very disappointed when they realize that "profit" was actually a 95% loss as that was all they would have gotten. I mean this in a totally hypothetical scenario, of course, nothing of the sorts would ever happen to me or any friend I know. 😢

So I suggest changing the wording to revenue. Or maybe "total value".

### DIFF
--- a/app/src/components/market/sections/market_sell/market_sell.tsx
+++ b/app/src/components/market/sections/market_sell/market_sell.tsx
@@ -245,7 +245,7 @@ const MarketSellWrapper: React.FC<Props> = (props: Props) => {
             <TransactionDetailsRow
               emphasizeValue={potentialValue ? potentialValue.gt(0) : false}
               state={ValueStates.success}
-              title={'Profit'}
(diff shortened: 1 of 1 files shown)


KadenZipfel (approved):

pimato:
I agree! The main issue is that we are showing the profit value wrong. The goal was since day one to show the actual profit so bought total - sell total. I am declined in changing the description but we should actually fix the broken way we calculate the profit value.

I´ve created an issue for this: https://github.com/protofire/omen-exchange/issues/1426

Issue #1403, opened 2020-12-01
Scalar market history chart


KadenZipfel:
The history chart doesn't automatically work for scalar markets. This is because the data that's being fetched for categorical market history doesn't exist for scalar markets. 

Instead of grabbing `outcomeTokenAmounts`, we should be grabbing `outcomeTokenMarginalPrices[0]`. Additionally, instead of returning percentage values for outcomes, we should be returning the proper scale, i.e. 100USD-400USD instead of 0%-100%, with only one outcome, representing the current price. 

This may require a design, though I think it would be simple enough to implement without a design just by adjusting the Y axis values and using only one outcome.

PR #1430, opened 2020-12-09
Scalar market history chart

KadenZipfel:
Closes https://github.com/protofire/omen-exchange/issues/1403

It might be good to later migrate to a different type of chart and or design, but for the initial release I figured this is pretty good.


hexyls (approved):
Looks good! 

One small thing I noticed is that the top yaxis tick can get a little bit cut off with larger values. 
Screen Shot 2020-12-10 at 11 36 26 am

pimato (approved):

Mi-Lan (approved):

Issue #1443, opened 2020-12-13
Retrieve `transaction` from `cpk.createScalarMarket` and `waitForBlockToSync`


KadenZipfel:
The `cpk.createScalarMarket` function should incorporate the `transaction` object as `cpk.createMarket` does, and wait for the transaction before navigating to the market page.

PR #1459, opened 2020-12-15
Wait for transaction before navigating to new scalar market

KadenZipfel:
Closes https://github.com/protofire/omen-exchange/issues/1443

### DIFF
--- a/app/src/components/market/sections/market_create/market_wizard_creator_container.tsx
+++ b/app/src/components/market/sections/market_create/market_wizard_creator_container.tsx
@@ -51,13 +51,17 @@ const MarketWizardCreatorContainer: FC = () => {
         }
 
         if (isScalar) {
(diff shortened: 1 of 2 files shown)


pimato (approved):
works!

pimato (approved):

hexyls (approved):
Subgraph is still busted so can't fully test but there is no reason this wouldn't work 👍

Mi-Lan (changes requested):
@KadenZipfel hmmm for me it seems not to work...when i create market and transaction gets confirmed it seems not to catch that and loader for creating market seems just to sit there indefinitely....

Mi-Lan (approved):
@KadenZipfel didn't know that i needed to have kadenzipfel/omen-rinkeby that subgraph....so when i used that one it worked like charm! GJ

Issue #1456, opened 2020-12-15
After creating/adding a market to an existing market/questionID market details view is not showing the currency dropdown


pimato:
Steps to reproduce:

1. Create a new market 1 with currency A
2. Click on the menu button to add a new currency to market 1
3. Create a new market 2 with currency B
4. BUG: User gets redirected to a market details view which is not showing the currency dropdown to switch between currencies:
<img width="605" alt="Bildschirmfoto 2020-12-15 um 15 35 23" src="https://user-images.githubusercontent.com/5337809/102228605-2d1c5400-3eeb-11eb-9cf2-23a7144442ff.png">

Doing a refresh:
<img width="614" alt="Bildschirmfoto 2020-12-15 um 15 47 30" src="https://user-images.githubusercontent.com/5337809/102230186-dadc3280-3eec-11eb-83ee-dc3a11ace5ac.png">

PR #1468, opened 2020-12-21
fix/#1456

hexyls:
closes https://github.com/protofire/omen-exchange/issues/1456 

`useQuery` returns cached results when variables are unchanged, even when mounted across different components. Updating the fetchPolicy to 'cache-and-network' will make sure we get updated values.


pimato (approved):
Works!

KadenZipfel (approved):
Perfect solution 🔥

Mi-Lan (approved):

Issue #1336, opened 2020-11-11
Add new Settings Page with ability to change and add custom RPC nodes


pimato:
With the current Infura outage, it is clear that Omen is not build to be a cockroach as Omen was not reachable at all. Our goal is to make Omen resilient from Infura or even subgraph outages. The first step is to allow Omen and its users to decide which RPC node it uses to connect to Ethereum. The RPC node configuration should be part of the new Settings page:

<img width="714" alt="settings" src="https://user-images.githubusercontent.com/5337809/98824988-4174b900-2434-11eb-8ed3-88317666e901.png">

Notes:
- Allow to pick curated RPC nodes powered by the DXdao (starting with Infura and an RPC node from the DXdao)
- Allow to define custom RPC node 
- Allow to switch back to default
- Any settings change should be stored in local storage so users don´t need to switch all the time they are using Omen.

Figma:
[Playground](https://www.figma.com/proto/aTivxuHmp2l6Sn8nho1Ngu/Playground?node-id=673%3A61563&viewport=272%2C305%2C0.7164108753204346&scaling=min-zoom)
[Design Specification](https://www.figma.com/file/aTivxuHmp2l6Sn8nho1Ngu/Playground?node-id=673%3A61563)

pimato:
All free available options from this side should be pickable:
https://ethereumnodes.com/

PR #1470, opened 2020-12-21
Feature - Add new Settings Page with ability to change and add custom RPC nodes

kboomr:
#1336


pimato:
@kboomr UI looks great so far, could you please add the functionality now?

pimato (changes requested):
see comment

Issue #1472, opened 2020-12-21
Disable 'continue' button if inputs invalid


KadenZipfel:
Currently, the user can continue to the funding step even if their inputs don't fit the requirement of: `lowerBound < startingPoint < upperBound'. Instead, their should be a check which disables the 'continue' button.

A good precursor to https://github.com/protofire/omen-exchange/issues/1444

PR #1477, opened 2020-12-22
Disable continue button if invalid inputs

KadenZipfel:
Closes https://github.com/protofire/omen-exchange/issues/1472


hexyls (approved):
🎉 nice catch!

pimato (approved):
a good temporary solution until we have the designs ready

Mi-Lan (approved):
Works good!

PR #1480, opened 2020-12-23
fix safe app

hexyls:
Fix for the breaking change introduced in safe-react v2.17.0

To test, add this deployment as a safe app: https://omen-staging.automate.vercel.app


app/src/util/cpk.ts
```diff
@@ -65,58 +59,15 @@ function standardizeTransaction(tx: Transaction): StandardTransaction {
...
-        messageId: SEND_TRANSACTIONS_V2,
-        requestId,
-        data,
-      }
-
-      this.txCallbacks.set(requestId, { confirm, reject })
-
-      window.parent.postMessage(message, '*')
-    })
+  async sendTransactions(transactions: StandardTransaction[], options?: ExecOptions): Promise<TransactionResult> {
```
germartinez, line 62: Is this necessary? Doesn't it have the same behaviour as the `cpk.execTransaction()` method when running as a Safe app inside the Multisig web interface?

app/package.json
```diff
@@ -10,7 +10,7 @@
     "3box-comments-react": "0.0.8",
     "@apollo/react-hooks": "^3.1.3",
     "@fvictorio/newton-raphson-method": "^1.0.5",
-    "@gnosis.pm/safe-apps-sdk": "0.4.2",
+    "@gnosis.pm/safe-apps-sdk": "1.0.1",
```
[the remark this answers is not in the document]
germartinez: Maybe it was implemented before the `contract-proxy-kit` worked as a Safe-app?

germartinez:
Hello, I'm Germán, from Gnosis.
I'm working on the `contract-proxy-kit`

I hope this message is not too abrupt or rude. Just wanted to help after being pointed to this PR after an error was reported from the dxdao regarding the messageId `SEND_TRANSACTIONS_V2` :)

I saw that the `contract-proxy-kit` and the `safe-apps-sdk` are both added as dependencies, which feels a bit confusing to me as the CPK already imports the Safe Apps SDK and I left a couple of comments.

Thank you! :)

app/src/hooks/useSafeApp.tsx
```diff
@@ -10,7 +9,7 @@ export const useSafeApp = () => {
   useEffect(() => {
     if (!safeInfo) {
       safeSdk.appsSdk.addListeners({
-        onSafeInfo: (safeInfo: SafeInfo) => setSafeInfo(safeInfo),
+        onSafeInfo: safeInfo => setSafeInfo(safeInfo),
```
mikheevm, line 12: @germartinez I'm also not sure about this LOC. New safe apps SDK doesn't use event-driven architecture anymore, not sure if it will work

germartinez: `contract-proxy-kit` `v2.2.0-alpha.1` is using `safe-apps-sdk` `v0.4.1`. It should be updated when the next CPK release comes out :)

app/package.json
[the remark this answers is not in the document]
hexyls: Hey @germartinez thank you for dropping by and helping out! I really appreciate it :)

`@gnosis.pm/safe-apps-sdk` was originally added as a dependency because the version of `contract-proxy-kit` we were using did not allow access to the underlying `getBySafeTxHash` method that we need [to get transaction details](https://github.com/protofire/omen-exchange/blob/master/app/src/services/cpk.ts#L134).

app/src/hooks/useSafeApp.tsx
hexyls: Omen is using two versions of `safe-apps-sdk` in this PR, one that uses the old event-driven architecture that comes with `contract-proxy-kit` and the other to send transactions with safeTxGas support through `safe-apps-sdk` v1.0.1. This is only until `contract-proxy-kit` supports v1+ of `safe-apps-sdk` 👍

app/src/util/cpk.ts
hexyls: Yeah need this just to send safeTxGas with the newer version of safe-apps-sdk 👍

app/src/util/cpk.ts
```diff
@@ -65,58 +59,15 @@ function standardizeTransaction(tx: Transaction): StandardTransaction {
 
 // Omen CPK monkey patch
 class OCPK extends CPK {
```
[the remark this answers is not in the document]
hexyls: We use it in [`createCPK`](https://github.com/protofire/omen-exchange/blob/master/app/src/util/cpk.ts#L154) which is imported into the [`connectedCPK`](https://github.com/protofire/omen-exchange/blob/master/app/src/hooks/connectedCpk.tsx#L5) hook.

KadenZipfel (approved):

Mi-Lan (approved):

pimato (approved):

Issue #1511, opened 2021-01-07
Fix incorrect max profit calculation


KadenZipfel:
Currently the max profit value in the buy view takes into account a selling fee, but it should be omitted since the value returned should be the max profit if the outcome shares are redeemed for the max value after market close, hence there should not be a fee included in the calculation.

PR #1515, opened 2021-01-07
Update scalar slider

KadenZipfel:
Closes https://github.com/protofire/omen-exchange/issues/1511

This PR changes the slider from the scalar market buy view to return payout/profit/loss values based on the redeem value at resolution, instead of returning the sell value. I've also added tooltips for clarity.


pimato (approved):

Mi-Lan (approved):
Nicely done!

Issue #1499, opened 2021-01-05
Asset picker not correct default asset, icon and balances not shown!


pimato:
This gif shows the issue:
![Bildschirmvideo-aufnehmen-2021-0](https://user-images.githubusercontent.com/5337809/103698741-145f0580-4fa2-11eb-91a7-04d23be2673b.gif)

Notes:
- Base asset should be shown on default, not the wrapped equivalent
- If xDai asset is picked, Icon is not shown
- Balances are not shown in the dropdown

PR #1537, opened 2021-01-13
Fix/#1499

hexyls:
closes https://github.com/protofire/omen-exchange/issues/1499


pimato:
Hey @hexyls ,

it looks like scalar markets have issues with native assets ( might be an issue introduce with my merge from master?):

Steps to reproduce:
1)Go to http://localhost:3000/#/0x07ec54f9484acd0c4b20b2722b44972c4d51f062
2) Click on the pool tab
3) Put any value below your balance in the inputfield:
4) BUG: Deposit Button is disabled:
Bildschirmfoto 2021-01-13 um 22 09 48

We have the same issue with Buy/Sell outcome tokens!

pimato:
Also: https://github.com/protofire/omen-exchange/issues/1540

It looks like the Max button is having issues with native assets :/

pimato (changes requested):
see comments

Mi-Lan (changes requested):
@hexyls Hmm seems that there is this error, its instant so it's probably something on front-end side...heres how to reproduce it 
Screenshot 2021-01-14 at 13 01 46
1) Go to xDai chain
2) Got to this market http://localhost:3000/#/0x866d69718a4c9e79e86315264187198da52b161e
3) Add some amount and click deposit

https://user-images.githubusercontent.com/33226956/104588715-15aac500-5669-11eb-8766-2e28637c4cb6.mov

app/src/components/market/common/currency_selector/index.tsx
```diff
@@ -80,10 +84,16 @@ export const CurrencySelector: React.FC<Props> = props => {
...
+      const selected = currency && currency.toLowerCase() === address.toLowerCase()
       currencyDropdownData.push({
-        content: <TokenItem image={image} text={symbol} />,
-        extraContent: currency && currency.toLowerCase() === address.toLowerCase() ? balance : '',
+        content: <TokenItem image={image} key={symbol} text={symbol} />,
+        extraContent: selected ? balance : '',
+        secondaryText: !tokenBalance
+          ? ''
+          : Number(tokenBalance) > 0
+          ? formatBigNumber(new BigNumber(tokenBalance), decimals, 5)
```
pRoy24, line 95: Do we need to format 5 decimals?
The dropdown width is short and the display breaks if there are actually 5 decimals.
image

hexyls: Great catch! This seems to cause issues with larger balances too. I ended up keeping the decimals for consistency but removed the hard width set on the secondary text 👍

KadenZipfel (approved):

pimato (approved):
Found a small issue regarding the dropdown picker but feel this has a lower priority: https://github.com/protofire/omen-exchange/issues/1546

Approved!

Mi-Lan (approved):
Works good now!

Issue #1542, opened 2021-01-15
Add a subgraph out of sync warning


hexyls:
Uniswap has a warning they display to users when their subgraph is out of sync by 30 blocks, might be useful to copy this.

PR #1544, opened 2021-01-15
Move waitForBlockToSync to CPK service

hexyls:
Moves `waitForBlockToSync` into the CPK's `waitForTransaction` to simplify and cover all transactions.

Didn't completely remove `useGraphMeta` because we might use the data later in https://github.com/protofire/omen-exchange/issues/1542 or elsewhere.


pimato (approved):

KadenZipfel (approved):

Mi-Lan (approved):
Works good hexyls good job! Also saw some other places where we use waitForTransaction function?

Issue #1550, opened 2021-01-18
Show Network modal if user is not connected to the right network


pimato:
![network-modal](https://user-images.githubusercontent.com/5337809/104967270-3e251d00-59e3-11eb-9404-a41e34951991.png)

`how do I connect to xDai?` Should open up a new tab to https://www.xdaichain.com/for-users/wallets/metamask/metamask-setup

Figma:
[Design Specification](https://www.figma.com/file/R056tSYMjvfwaFidEBItYy/Widgets-V2?node-id=1209%3A141)

PR #1551, opened 2021-01-19
Mainnet <> xDai

KadenZipfel:
Closes: https://github.com/protofire/omen-exchange/issues/1550

To test locally, checkout `451fd71` and spin up `localhost:3000` and `localhost:3001` at the same time.


pimato:
Did some simple testing and works great so far!

hexyls (approved):
This is great 🙌

pimato:
works great :)

pimato (approved):

Mi-Lan (approved):
Very cool feature! Only thing i could suggest is different background message instead of Market not found since we know that it is because of wrong network we can be more specific.
Screenshot 2021-01-20 at 11 34 26

KadenZipfel:
> Very cool feature! Only thing i could suggest is different background message instead of Market not found since we know that it is because of wrong network we can be more specific.
> Screenshot 2021-01-20 at 11 34 26

Good call! I'll put together an issue for this

Issue #1516, opened 2021-01-07
Set Outcome Widget on xDai is showing Balance in Ether but should be xDai


pimato:
<img width="611" alt="Bildschirmfoto 2021-01-08 um 00 43 13" src="https://user-images.githubusercontent.com/5337809/103957259-9b4ee200-514a-11eb-9381-03eb6f5024e5.png">

PR #1552, opened 2021-01-19
Fix/#1516

hexyls:
closes https://github.com/protofire/omen-exchange/issues/1516


KadenZipfel (approved):

Mi-Lan:
Was unable to test it since we can't create market on xDai and no markets had option to bond....

hexyls:
@Mi-Lan a good way to test this is to checkout the branch, pop into [open.tsx](https://github.com/protofire/omen-exchange/blob/master/app/src/components/market/sections/market_profile/market_status/open.tsx#L253), and hardcode `currentTab` to `MarketDetailsTab.finalize`.

pimato (changes requested):
@hexyls I´ve tried to bond some xdai on an outcome and it does not seem to work:

http://localhost:3000/#/0x1ec1bf127ceb4dd6c1d8640a65668d13b9eb09d9

Bildschirmfoto 2021-01-20 um 11 09 20

Mi-Lan (changes requested):
@hexyls i'm getting same error as geronimo...Also should the amount on xDai be changed? Bonding 0.1 xDai is really nothing in my opinion? 0.1ETH in xDai makes more sense...
Screenshot 2021-01-20 at 13 19 58

hexyls:
@Mi-Lan @pimato I think this error is because we are currently indexing markets that are using the old Realitio contracts. Should be resolved when the subgraph is updated 👍 In general this PR is cosmetic only so it should not impact functionality.

@Mi-Lan that is a great point about the bond amount! Updated to 10 xDAI 🙌

Mi-Lan (approved):

KadenZipfel (changes requested):

KadenZipfel (approved):
LGTM!

app/src/components/market/sections/market_bond/market_bond.tsx
```diff
@@ -61,8 +64,10 @@ const MarketBondWrapper: React.FC<Props> = (props: Props) => {
   const [isModalTransactionResultOpen, setIsModalTransactionResultOpen] = useState(false)
   const [outcomeIndex, setOutcomeIndex] = useState<number>(0)
   const probabilities = balances.map(balance => balance.probability)
+  const initialBondAmount =
+    networkId === networkIds.XDAI ? parseUnits('10', nativeAsset.decimals) : parseUnits('0.01', nativeAsset.decimals)
   const [bondEthAmount, setBondEthAmount] = useState<BigNumber>(
```
pRoy24, line 69: Just a nomenclature change, we should probably change the state and setter to [bondNativeAssetAmount, setBondNativeAssetAmount]

app/src/components/market/sections/market_bond/market_bond.tsx
```diff
@@ -153,7 +158,10 @@ const MarketBondWrapper: React.FC<Props> = (props: Props) => {
         <div>
           <>
             <CurrenciesWrapper>
-              <EthBalance value={`${formatNumber(formatBigNumber(ethBalance, TokenEthereum.decimals, 3), 3)}`} />
+              <EthBalance
```
pRoy24, line 161: Similar this component should be renamed to <NativeAssetBalance ... />

pRoy24 (commented):
Looks good, suggested a few nomenclature changes since the asset state variables and components to Native... instead of Eth... since these are no longer specific to ETH but general to the native asset on the chain.

pimato (approved):

PR #1557, opened 2021-01-19
UI Settings (#1556)

Mi-Lan:
Co-authored-by: kboomr <boor2k@gamil.com>
Co-authored-by: Thomas Pulber <corkus@pm.me>
closes https://github.com/protofire/omen-exchange/issues/1336


Mi-Lan:
@pimato Would it make sense to disable save button if the url selected is not okay? Or offline?

pimato:
active inputfield styling not according our design:
Bildschirmfoto 2021-02-19 um 16 19 45

should look like:
Bildschirmfoto 2021-02-19 um 16 19 53

pimato:
Only when user switches to not default rpc, button should be active:
Bildschirmfoto 2021-02-19 um 16 22 17

Currently, button is always active, even the user has nothing changed!
`Save` Button should only be active if user changed rpc.

app/src/components/settings/settings_view/index.tsx
```diff
@@ -0,0 +1,254 @@
...
+        </ButtonRound>
+        <ButtonRow>
+          <ButtonRound
+            onClick={() => {
+              setCurrent(0)
+              setUrl(urlObject[0].rpcUrl)
+            }}
+          >
+            Set to Default
+          </ButtonRound>
```
KadenZipfel, line 229: As a user, when selecting 'Set to default', I expect the same effect as selecting an endpoint and clicking save, i.e. going to a new page and reloading

app/src/components/settings/settings_view/index.tsx
```diff
@@ -0,0 +1,254 @@
...
+              if (!(await checkRpcStatus())) return
+
+              sessionStorage.setItem(
+                'rpcAddress',
+                JSON.stringify({
+                  url: url,
+                  network: context.networkId,
+                }),
+              )
+              history.replace('/')
```
KadenZipfel, line 242: Would also be nice here if the user was taken back to their last page instead of going to the landing page

app/src/util/networks.ts
```diff
@@ -201,6 +223,13 @@ const networks: { [K in NetworkId]: Network } = {
   [networkIds.XDAI]: {
     label: 'xDai',
     url: 'https://rpc.xdaichain.com/',
+    alternativeUrls: [
+      {
+        rpcUrl: 'https://rpc.xdaichain.com/',
+        name: 'xDai',
+      },
+      { rpcUrl: 'https://xdai-archive.blockscout.com', name: 'xDai' },
+    ],
```
KadenZipfel, line 232: These should have different names to be less confusing. Maybe 'xDai' and 'xDai - Blockscout' or something like that

KadenZipfel (changes requested):
Some other notes:
- The `app/omen-exchange` directory also snuck into this PR somehow
- If I switch my endpoint, then go back to the settings page, I expect my current endpoint to be selected
- The settings cog icon is quite dark compared to other elements, which draws my eye towards it quite a bit. Personally I think it should be a little less noticeable, maybe with a lighter color (cc @pimato)
- Would be nice to have icons for the other endpoints if possible:
image

app/src/components/settings/assets/images/DXdao.svg
```diff
@@ -0,0 +1,25 @@
+<svg width="25" height="24" viewBox="0 0 25 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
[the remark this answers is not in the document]
Mi-Lan: hmm yeah that is leftover from kboomr will delete it!

Mi-Lan:
> Some other notes:
> 
> * The `app/omen-exchange` directory also snuck into this PR somehow
> * If I switch my endpoint, then go back to the settings page, I expect my current endpoint to be selected
> * The settings cog icon is quite dark compared to other elements, which draws my eye towards it quite a bit. Personally I think it should be a little less noticeable, maybe with a lighter color (cc @pimato)
> * Would be nice to have icons for the other endpoints if possible:
>   image

- Hmm so that should work the current rpc thats being used is selected by default...Do you think it should be different?

https://user-images.githubusercontent.com/33226956/108552645-e46d8680-72f1-11eb-90d4-a434c1a3bbd8.mov

pimato:
> Some other notes:
> * The settings cog icon is quite dark compared to other elements, which draws my eye towards it quite a bit. Personally I think it should be a little less noticeable, maybe with a lighter color (cc @pimato)

Thanks for the great feedback @KadenZipfel !

@Mi-Lan could we please change the color of the settings icon to: #86909E

pimato (changes requested):
see comment

pimato:
Hey @Mi-Lan ,

just created a cloudfare icon:
[cloudflare.svg.zip](https://github.com/protofire/omen-exchange/files/6012876/cloudflare.svg.zip)

app/src/components/settings/settings_view/index.tsx
```diff
@@ -0,0 +1,275 @@
...
+      secondaryText: '',
+      onClick: item.onClick,
+    }
+  })
+  const checkRpcStatus = async () => {
+    try {
+      await axios.post(url, {
+        id: +new Date(),
+        jsonrpc: '2.0',
+        method: 'eth_getBlockByNumber',
```
hexyls, line 164: looks like more providers might support `eth_blockNumber` over `eth_getBlockByNumber`. Check out my comment below about the blockscout endpoint, it seems to only support `eth_blockNumber`

app/src/util/networks.ts
```diff
@@ -88,6 +89,15 @@ const networks: { [K in NetworkId]: Network } = {
   [networkIds.MAINNET]: {
     label: 'Mainnet',
     url: `https://mainnet.infura.io/v3/${INFURA_PROJECT_ID}`,
+    alternativeUrls: [
+      {
+        rpcUrl: `https://mainnet.infura.io/v3/${INFURA_PROJECT_ID}`,
+        name: 'Infura',
+      },
+      { rpcUrl: 'https://cloudflare-eth.com/', name: 'Cloudflare' },
+      { rpcUrl: 'https://blockscout.com', name: 'BlockScout' },
```
hexyls, line 98: blockscout mainnet endpoint should be `https://blockscout.com/poa/xdai/api/eth-rpc`

app/src/components/settings/settings_view/index.tsx
```diff
@@ -0,0 +1,275 @@
...
+  dropdownItem.push({
+    title: 'Custom',
+    image: undefined,
+    onClick: () => {
+      setCurrent(dropdownItem.length - 1)
+    },
+  })
+
+  const filterItems = dropdownItem.map(item => {
+    // console.log(index)
```
hexyls, line 147: logggyy

hexyls (changes requested):
new settings page looking sick @Mi-Lan nice work 🙌

app/src/components/settings/settings_view/index.tsx
Mi-Lan: @pimato yeah we should probably agree on this since geronimo gave feedback to not change page but to stay on the current page....

app/src/components/settings/settings_view/index.tsx
Mi-Lan: Ohh great will change that! Wasn't aware of it...

app/src/util/networks.ts
Mi-Lan: awesome seems that i was targeting wrong url! Now it works!

pimato:
Hey @Mi-Lan I´ve added a blockscout svg icon:
[Cloudflare.svg.zip](https://github.com/protofire/omen-exchange/files/6029126/Cloudflare.svg.zip)

pimato:
Also if an Item has no circle-icon, the height should still stay at 38px.
Currently:
Bildschirmfoto 2021-02-23 um 14 41 57

Should:
Bildschirmfoto 2021-02-23 um 14 43 16

pimato:
Bildschirmfoto 2021-02-23 um 14 44 53

If I click on the settings icon again, it will redirect me to the main page. It should not do anything, not even a reload!

pimato:
Bildschirmaufnahme 2021-02-23 um 15 45 38

Mi-Lan:
Current progress is that i'm trying to implement feature that will recognize if current rpc is broken and redirect to a settings page so that user can choose some other rpc...Because currently the way it is setup is that once app encounters bad rpc it will just present us with loading circle of death. Which will display when the infura node is down so kinda defeats the purpose of settings page if you can't access it once the default url is down...

When rpc is invalid we don't have access to which chain we are on so i don't know which list of alternative urls to show. Is it okay to get networkId based on url?
@KadenZipfel @hexyls @pimato

hexyls:
@Mi-Lan yeah I had a similar problem of finding the right networkId when I had to disable Authereum on xDai - [see code here](https://github.com/protofire/omen-exchange/blob/78615a991edbe928cc2f04d57daea7ccd1f9bdec/app/src/components/modal/modal_connect_wallet/index.tsx#L217). Like you said you can use the subdomain in the url and also `windowObj.ethereum.chainId` if it is available.

app/src/hooks/connectedWeb3.tsx
```diff
@@ -31,12 +31,15 @@ export const useConnectedWeb3Context = () => {
...
+interface Props {
+  children?: React.ReactNode
+  setStatus?: any
+}
 /**
  * Component used to render components that depend on Web3 being available. These components can then
  * `useConnectedWeb3Context` safely to get web3 stuff without having to null check it.
  */
-export const ConnectedWeb3: React.FC = props => {
+export const ConnectedWeb3: React.FC = (props: Props) => {
```
hexyls, line 42: to fix that prop error you were getting you can do `React.FC<Props>`

Mi-Lan: ohh man you are saving me again :D Thanks man you'r awesome!

Mi-Lan:
The problem i mentioned on the call was with Header component(app/src/components/common/layout/header/index.tsx)
namely i import header in app.tsx(in order to show settings page when user is not logged in) and in Main component where we show it as usual.
Screenshot 2021-03-01 at 19 33 48

- Issue i had was that in the MAIN component i have access to both web3 context and react router but in App.tsx since its higher order component web3 and router haven't been initialised
Screenshot 2021-03-01 at 19 34 48
- To solve that i exported one header with router and one without...
Screenshot 2021-03-01 at 19 35 57
- There the problem arises since one version of header has history properties and one has not....ReactRouterComponents don't work since by default they can't be exported i tried with Partial as well...if i just try to put history?,router? it still doesn't work properly...
Screenshot 2021-03-01 at 19 36 33

pimato (changes requested):
Currently, if the user sets a custom rpc node, the UI does not reflect the usage of it after user clicked save:
custom-1

Should stay like this, after user saved the custom rpc:
custom

Mi-Lan:
> Currently, if the user sets a custom rpc node, the UI does not reflect the usage of it after user clicked save:
> custom-1
> 
> Should stay like this, after user saved the custom rpc:
> custom

That was already implemented its just that along the way different source for network id was used so it got lost...Nice catch!
Screenshot 2021-03-01 at 23 50 27
Is it okay to be implemented so that placeholder is the current rpc or?

hexyls (approved):

app/src/hooks/connectedWeb3.tsx
```diff
@@ -31,16 +31,20 @@ export const useConnectedWeb3Context = () => {
...
  * Component used to render components that depend on Web3 being available. These components can then
  * `useConnectedWeb3Context` safely to get web3 stuff without having to null check it.
  */
-export const ConnectedWeb3: React.FC = props => {
+export const ConnectedWeb3: React.FC<Props> = (props: Props) => {
   const [networkId, setNetworkId] = useState<number | null>(null)
   const safeAppInfo = useSafeApp()
   const context = useWeb3Context()
   const { account, active, error, library } = context
+  const rpcAddress: string | null = sessionStorage.getItem('rpcAddress')
```
pRoy24, line 47: Did you consider the usage of sessionStorage vs localStorage? 
Do you want the user settings to carry over if the user closes the tab and opens again?
If so better to use localStorage.

Mi-Lan: yeah i tried localStorage but in my opinion its too persistent...whats your opinion? I think its okay when user closes tab it goes away...

pimato:
> > Currently, if the user sets a custom rpc node, the UI does not reflect the usage of it after user clicked save:
> > custom-1
> > Should stay like this, after user saved the custom rpc:
> > custom
> 
> That was already implemented its just that along the way different source for network id was used so it got lost...Nice catch!
> Screenshot 2021-03-01 at 23 50 27
> Is it okay to be implemented so that placeholder is the current rpc or?

Would prefer to not have it as a hint in the Inputfield but actually as a text input.

pimato (approved):

pimato (approved):

app/src/hooks/connectedWeb3.tsx
pRoy24: I personally think that it should be localStorage as well, since changing the rpc URL is an infrequent action and I'd want my setting to be persisted if I close the tab.

app/src/App.tsx
```diff
@@ -1,36 +1,63 @@
...
 import connectors from './util/connectors'
+import { getInfuraUrl, networkIds } from './util/networks'
+import { checkRpcStatus, getNetworkFromChain } from './util/tools'
 
 const store = configureStore({ reducer: balanceReducer })
 
 const App: React.FC = () => {
+  const windowObj: any = window
+
+  const ethereum = windowObj.ethereum
```
pRoy24, line 28: This will again run into the same error that we fixed before for non web3 browsers.
image
You should check if etherum object exists before trying to reference ethereum.chainId

pRoy24 (changes requested):

pRoy24:
> @pimato @KadenZipfel @pRoy24 Would it make sense to disable save button if the url selected is not okay? Or offline? Also i think it would be useful to implement some error catcher if the default url is down and to redirect to settings page so that user can change it?

Yes I think disabling the save button if chosen custom RPC is down is the correct interaction.

However I am not sure if we should redirect to settings page vs. switch to default URL with notification that the custom web RPC is down at a future time. Perhaps @pimato can shed some light.

app/src/App.tsx
```diff
@@ -1,36 +1,63 @@
...
+  const networkId = ethereum.chainId
+  const [status, setStatus] = useState(true)
+  const network = getNetworkFromChain(networkId)
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
pRoy24: On second thought, I think window.ethereum.on('networkChanged', ) is deprecated now, so might be ok to listen to changes to window.ethereum object (although it may lead to unnecessary RPC calls) if other keys in the ethereum object change.

pRoy24:
Also if I add a custom RPC node and then switch network then the app becomes unusable and gets stuck in the loading page.
image

pRoy24:
I am still able to add different rpc network url from the network which I am connected to on Metamask.
Eg. I am connected to mainnet on metamask but able to add xdai rpc url or ropsten rpc url.
Wondering if this would lead to unexpected outcomes/values or cause errors.

image

pRoy24:
Also if I am connected to mainnet on metamask and give an xDai RPC url, how does the app know which network to resolve to?

Mi-Lan:
> Also if I am connected to mainnet on metamask and give an xDai RPC url, how does the app know which network to resolve to?

yeah sorry about that i just pushed it!

KadenZipfel:
When navigating to the settings page, Infura status seems to be defaulting to 'Unavailable' for some reason, even though it appears to work properly and clicking 'Set to Default' causes it to display 'Status: OK'.


https://user-images.githubusercontent.com/30579067/110163961-93f03180-7da5-11eb-96c0-52b1b56068a1.mov

Mi-Lan:
> When navigating to the settings page, Infura status seems to be defaulting to 'Unavailable' for some reason, even though it appears to work properly and clicking 'Set to Default' causes it to display 'Status: OK'.
> 
>  Screen.Recording.2021-03-05.at.11.22.21.AM.mov

@KadenZipfel I think this fixes it...It has that switch from unavailable to available for brief second because it needs time to fetch the api to check and i set it to be false by default...

https://user-images.githubusercontent.com/33226956/110165500-4ef2ec00-7df3-11eb-97fe-49ea37bcb6eb.mov

KadenZipfel:
> > When navigating to the settings page, Infura status seems to be defaulting to 'Unavailable' for some reason, even though it appears to work properly and clicking 'Set to Default' causes it to display 'Status: OK'.
> > Screen.Recording.2021-03-05.at.11.22.21.AM.mov
> 
> @KadenZipfel I think this fixes it...It has that switch from unavailable to available for brief second because it needs time to fetch the api to check and i set it to be false by default...
> 
>  Screen.Recording.2021-03-05.at.20.42.04.mov

Looks like that change caused a new bug:

https://user-images.githubusercontent.com/30579067/110186328-b98f3200-7dc9-11eb-8e57-53776c76b0bb.mov

Mi-Lan:
> > > When navigating to the settings page, Infura status seems to be defaulting to 'Unavailable' for some reason, even though it appears to work properly and clicking 'Set to Default' causes it to display 'Status: OK'.
> > > Screen.Recording.2021-03-05.at.11.22.21.AM.mov
> > 
> > 
> > @KadenZipfel I think this fixes it...It has that switch from unavailable to available for brief second because it needs time to fetch the api to check and i set it to be false by default...
> > Screen.Recording.2021-03-05.at.20.42.04.mov
> 
> Looks like that change caused a new bug:
> 
>  Screen.Recording.2021-03-05.at.3.43.59.PM.mov

Hmmm thats very weird...since it should reload page on save....I'm looking into it but couldn't quite figure out how to trigger it...

pRoy24 (approved):
LG

KadenZipfel:
> > > > When navigating to the settings page, Infura status seems to be defaulting to 'Unavailable' for some reason, even though it appears to work properly and clicking 'Set to Default' causes it to display 'Status: OK'.
> > > > Screen.Recording.2021-03-05.at.11.22.21.AM.mov
> > > 
> > > 
> > > @KadenZipfel I think this fixes it...It has that switch from unavailable to available for brief second because it needs time to fetch the api to check and i set it to be false by default...
> > > Screen.Recording.2021-03-05.at.20.42.04.mov
> > 
> > 
> > Looks like that change caused a new bug:
> > Screen.Recording.2021-03-05.at.3.43.59.PM.mov
> 
> Hmmm thats very weird...since it should reload page on save....I'm looking into it but couldn't quite figure out how to trigger it...
> 
>  Screen.Recording.2021-03-08.at.13.04.50.mov 
> This is how it is for me...

As an update, I'm still getting this bug. To reproduce: 
- Navigate to settings page from another page (Infura should already be selected)
- Click save
- Status is set as unavailable until you reselect Infura or another provider from the dropdown

A possible fix that I think is simple and intuitive would be to disable the 'save' button until a user selects a new provider (either from dropdown or selecting 'set to default') since it should already be saved anyway

Mi-Lan:
> > > > > When navigating to the settings page, Infura status seems to be defaulting to 'Unavailable' for some reason, even though it appears to work properly and clicking 'Set to Default' causes it to display 'Status: OK'.
> > > > > Screen.Recording.2021-03-05.at.11.22.21.AM.mov
> > > > 
> > > > 
> > > > @KadenZipfel I think this fixes it...It has that switch from unavailable to available for brief second because it needs time to fetch the api to check and i set it to be false by default...
> > > > Screen.Recording.2021-03-05.at.20.42.04.mov
> > > 
> > > 
> > > Looks like that change caused a new bug:
> > > Screen.Recording.2021-03-05.at.3.43.59.PM.mov
> > 
> > 
> > Hmmm thats very weird...since it should reload page on save....I'm looking into it but couldn't quite figure out how to trigger it...
> > Screen.Recording.2021-03-08.at.13.04.50.mov
> > This is how it is for me...
> 
> As an update, I'm still getting this bug. To reproduce:
> 
> * Navigate to settings page from another page (Infura should already be selected)
> * Click save
> * Status is set as unavailable until you reselect Infura or another provider from the dropdown
> 
> A possible fix that I think is simple and intuitive would be to disable the 'save' button until a user selects a new provider (either from dropdown or selecting 'set to default') since it should already be saved anyway

@KadenZipfel hmmm you might have the old code that was fixed...you can't click save for the item that is already selected

https://user-images.githubusercontent.com/33226956/110374205-ae9a0300-8050-11eb-9edc-3a1dac8e2f23.mov

pimato (approved):

Issue #1536, opened 2021-01-12
[BUG]When switching time frame for History chart error occurs


Mi-Lan:
Steps to reproduce
- Go to market 
- Go to history section
- Switch from 1m to 1y then to All and this error will occur
![Screenshot 2021-01-12 at 23 45 14](https://user-images.githubusercontent.com/33226956/104383512-3b807e80-5530-11eb-81fc-42c4e9431028.png)

Mi-Lan:
#1194

PR #1565, opened 2021-01-21
Fix/#1536

hexyls:
closes https://github.com/protofire/omen-exchange/issues/1536


hexyls:
oops this is being resolved in a different PR

Issue #1439, opened 2020-12-11
Make native blockchain asset primary compared to its wrapped alternative


pimato:
After going through @hexyls PR which will allow users to use native Ether to interact with Omen I think it is time to rethink the role of WETH on Omen. I believe we have now the chance to hide WETH almost entirely:

- Show market details page of markets with WETH as collateral always in Ether
<img width="686" alt="Bildschirmfoto 2020-12-11 um 12 09 03" src="https://user-images.githubusercontent.com/5337809/101896735-b4935b80-3ba9-11eb-9e66-34aed582bda6.png">

- In the `Buy View` we should show WETH values in ETH:
<img width="639" alt="Bildschirmfoto 2020-12-11 um 12 11 11" src="https://user-images.githubusercontent.com/5337809/101896973-1ce23d00-3baa-11eb-94b1-640bf68dc8ee.png">

- In the `Sell View` we should show WETH values in ETH:
<img width="723" alt="Bildschirmfoto 2020-12-11 um 12 20 36" src="https://user-images.githubusercontent.com/5337809/101897777-59faff00-3bab-11eb-8b7a-75db9f32aaec.png">

- In the `Deposit Pool` we should show WETH values in ETH:
<img width="837" alt="Bildschirmfoto 2020-12-11 um 12 23 17" src="https://user-images.githubusercontent.com/5337809/101897976-ab0af300-3bab-11eb-8c45-347243093d0f.png">

- In the `Withdraw Pool` we should show WETH values in ETH:
<img width="845" alt="Bildschirmfoto 2020-12-11 um 12 24 17" src="https://user-images.githubusercontent.com/5337809/101898111-dbeb2800-3bab-11eb-9097-a37b9ed1b350.png">

Issue #1538, opened 2021-01-13
When using an ETH pool with WETH as base token the Sell outcome/withdraw funding page should allow withdrawal in ETH and not just WETH


pRoy24:
Currently when creating an ETH pool, the create pool allows initial pool deposit in both ETH and WETH.
Same with Buy outcome token/fund pool for that pool (You can buy or fund using either ETH or WETH)
However when selling outcome tokens or withdrawing pool funding the withdrawal in allowed only in WETH.
The withdrawal pages should have a dropdown similar to funding page allowing user to choose which currency they want to withdraw in.

hexyls:
related https://github.com/protofire/omen-exchange/issues/1507

PR #1569, opened 2021-01-22
Fix/#1439

hexyls:
closes https://github.com/protofire/omen-exchange/issues/1439


Mi-Lan (changes requested):
When wanting to buy when i select weth i get ETH as indetifier? This is probably easy fix i see that you have implemented the same thing on deposit....
Screenshot 2021-01-25 at 12 11 14

KadenZipfel (approved):
LGTM!

Mi-Lan (approved):

app/src/hooks/useSymbol.tsx
```diff
@@ -0,0 +1,14 @@
+import { getNativeAsset, getWrapToken } from '../util/networks'
```
pRoy24, line 1: I think this is a candidate for a util function rather than a hook.
Hooks typically modify the state in some way, this is a stateless function that takes an input and returns an output.

app/src/components/market/common/market_top_details_open/index.tsx
```diff
@@ -110,25 +110,37 @@ const MarketTopDetailsOpen: React.FC<Props> = (props: Props) => {
...
+          element.collateralToken === wrapTokenAddress ? nativeAssetAddress : element.collateralToken
+        return collateralToken === currency.address.toLowerCase()
+      })
       if (selectedMarket && selectedMarket.collateralToken !== collateral.address) {
         history.replace(`/${selectedMarket.id}`)
       }
     }
   }
 
+  const filter = marketsRelatedQuestion.map(({ collateralToken }) =>
```
pRoy24, line 129: This logic is being reused in a bunch of places, could move this to a util function.

app/src/components/market/common/market_top_details_open/index.tsx
```diff
@@ -110,25 +110,37 @@ const MarketTopDetailsOpen: React.FC<Props> = (props: Props) => {
...
+  const nativeAssetAddress = getNativeAsset(context.networkId).address.toLowerCase()
+  const wrapTokenAddress = getWrapToken(context.networkId).address.toLowerCase()
+
   const onChangeMarketCurrency = (currency: Token | null) => {
     if (currency) {
-      const selectedMarket = marketsRelatedQuestion.find(e => e.collateralToken === currency.address.toLowerCase())
+      const selectedMarket = marketsRelatedQuestion.find(element => {
+        const collateralToken =
+          element.collateralToken === wrapTokenAddress ? nativeAssetAddress : element.collateralToken
+        return collateralToken === currency.address.toLowerCase()
```
pRoy24, line 121: Should lowercase both sides in address comparisions.

app/src/components/market/sections/market_sell/market_sell.tsx
```diff
@@ -5,7 +5,13 @@ import { RouteComponentProps, withRouter } from 'react-router-dom'
 import ReactTooltip from 'react-tooltip'
 import styled from 'styled-components'
 
-import { useAsyncDerivedValue, useConnectedCPKContext, useConnectedWeb3Context, useContracts } from '../../../../hooks'
+import {
```
pRoy24, line 8: Got error when trying to sell max shares in ETH 
image

app/src/components/market/sections/market_sell/market_sell.tsx
```diff
@@ -280,7 +287,7 @@ const MarketSellWrapper: React.FC<Props> = (props: Props) => {
               title={'Total'}
               value={`${
                 tradedCollateral ? formatNumber(formatBigNumber(tradedCollateral, collateral.decimals, 2)) : '0.00'
-              } ${collateral.symbol}`}
+              } ${symbol}`}
```
pRoy24, line 290: We could add a dropdown similar to https://github.com/protofire/omen-exchange/pull/1570/files#diff-b9e7bcd4798ff09ab957e9c99ff18b23f2ce18875844965cf021e5600cd13c94R175
Allowing user to withdraw either in native asset or wrapped native asset.

app/src/components/market/sections/market_sell/market_sell.tsx
pRoy24: Steps to reproduce.
Go to market /#/0xf64d13a5aea7b1d20cbae747898f6099e77a7370 on Rinkeby.
Buy outcomes using ETH.
Try to sell max amount of outcomes.

pRoy24: I have seen this error when developing the compound feature as well so might not be directly related to this particular PR but to one of the previous commits in this chain, so feel free to tackle this in a separate issue if you feel that this is not within the scope of this PR.

app/src/components/market/common/market_top_details_closed/index.tsx
```diff
@@ -49,7 +49,7 @@ const MarketTopDetailsClosed: React.FC<Props> = (props: Props) => {
     address,
     answerFinalizedTimestamp,
     arbitrator,
-    collateral: collateralToken,
+    collateral,
```
pRoy24, line 52: Thanks for fixing this :)
This keeps bugging me when I come to this file because in every other component collateral is the actual token but in here it is the address of the token.

pRoy24 (commented):
Left some comments regarding improvements that could be made in this PR such as allowing outcome tokens, pool tokens to be withdrawn in both native as well as wrapped native asset.

Got a blocking error when trying to withdraw max amount of outcome tokens in native asset.

Also generally I think we should use checksum address instead of lowercase address in all objects/comparisions, I find it confusing because in some places we set the lowercase value of addresses in objects/for comparisons and in other places we don't.

app/src/hooks/useSymbol.tsx
hexyls: for sure this would def make a great util function but it depends on another hook `useConnectedWeb3Context` so just keeping hookish things grouped together. It would be great if I could just pass networkId down as a param but this hook is used in places where networkId is not available.

app/src/components/market/common/market_top_details_open/index.tsx
hexyls: great point!

app/src/components/market/sections/market_sell/market_sell.tsx
hexyls: hmmmm I couldn't reproduce this 🤔 def make a new issue so we explore though. I think this has happened before and it turned out that the subgraph was out of sync but unsure if this is the same or not.

app/src/components/market/sections/market_sell/market_sell.tsx
hexyls: yeah definitely that will be awesome! will follow this up in https://github.com/protofire/omen-exchange/issues/1538

hexyls:
thanks @pRoy24, cleaned up some of the code by moving more stuff to utils 🤘

will explore withdrawal dropdown, blocking error, and replacing lowercasing with checksums in other issues 👍

pimato:
Amazing work!

pimato (approved):

Issue #1284, opened 2020-11-01
Deploy a tokenCuratedRegistry which stores a reality question id hash and not simply market contract address


pimato:
Currently, the DXdao omen curation is adding/removing market contract address which leads to multiple verification for a market which shares the same reality.eth question id. The best way of doing the curation would be to actually verify the question id hash on an on-chain registry so new markets with different currency are automatically valid as they share the verified question id.

pimato:
+ If we are going to create a new market curation registry, we should think about token address for multiple networks like we see here: https://github.com/protofire/omen-exchange/pull/1580

PR #1580, opened 2021-01-27
Add alternative token address map to get token images

hexyls:
So pretty much this is to make sure we load up the Stake token image on xDai. The address for Stake on xDai is `0xb7D311E2Eb55F2f68a9440da38e7989210b9A05e` but the [repo we use](https://github.com/trustwallet/assets) to get token images uses the Ethereum address `0x0Ae055097C6d159879521C384F1D2123D1f195e6`

Not totally sure this is the right approach though because we are loading Stake from the TCR so we really shouldn't have anything hard coded about it in the codebase... 🤔 🤷‍♂️ thoughts?


pRoy24 (approved):
Just a side note that, I have generally observed that we are not doing toChecksum address anywhere before doing these address string comparisons.
Depending on where the address is coming from, eg. 3rd party API or user input these string comparisons might fail because of casing of address strings, we should either toLowerCase() or do ethers.utils.getAddress to get checksum address.
In this case it is fine because the address is coming from our codebase itself.

Mi-Lan (approved):
Works good!

app/src/util/token.ts
```diff
@@ -1,7 +1,16 @@
...
+const alternativeAddresses: { [A in HasAlternativeAddress]: string } = {
+  // STAKE token
+  '0xb7D311E2Eb55F2f68a9440da38e7989210b9A05e': '0x0Ae055097C6d159879521C384F1D2123D1f195e6',
+}
+
 export function getImageUrl(tokenAddress?: string): string | undefined {
   if (!tokenAddress) return undefined
   tokenAddress = utils.getAddress(tokenAddress)
+  const alternativeAddress = alternativeAddresses[tokenAddress as HasAlternativeAddress]
+  tokenAddress = alternativeAddress ? alternativeAddress : tokenAddress
```
pRoy24, line 14: Good practice to transform to checksum address here as well since we are doing that in line 12.
utils.getAddress

KadenZipfel (approved):
I think this is fine as a quick fix, but of course not very scalable if we have to update this every time a token is added to the xDai TCR, so we should really make sure to come up with a longer-term solution at some point

pimato:
> I think this is fine as a quick fix, but of course not very scalable if we have to update this every time a token is added to the xDai TCR, so we should really make sure to come up with a longer-term solution at some point

I agree, this should be fixed on the market curation level: https://github.com/protofire/omen-exchange/issues/1284

app/src/util/token.ts
hexyls: hardcoded address is checksummed already 👍

Issue #1581, opened 2021-01-27
Value must be less than balance error behind dialog when trying to buy/sell outcome tokens, add/remove pool tokens


pRoy24:
![image](https://user-images.githubusercontent.com/4677914/105994714-955d7880-60da-11eb-865b-ce7d988cb26a.png)

When you have actually done a transaction, the value in the background still attempts to update.
This results in the outcome table flickering on the background while the transaction is processing and also shows an error in the background if the balance after the transaction would have processed is now less than the amount entered.

The simple way to solve this would be to reset the input fields immediately after the transaction has fired and not after the transaction has succeeded.

PR #1584, opened 2021-01-28
Fix for https://github.com/protofire/omen-exchange/issues/1581 Reset input values when transaction fires

pRoy24:
Verify that the issue referenced has been fixed.
Try to buy outcome tokens with almost all of of balance. Click buy.
You should not see an error in the background saying "Value must be less than or equal to remaining balance"
The outcome table values should be reset and should not be flickering in the background.
Repeat the same process for Sell outcome tokens and buy/sell pool tokens.


pimato (approved):
Great work!

Mi-Lan (approved):

KadenZipfel (approved):

PR #1586, opened 2021-01-29
fix dropdown control issues

hexyls:
At the moment the dropdown component's item index is controlled both by external props and internal state at the same time. This can cause issues where we have to wait a whole render cycle for the item index to be updated which can cause a brief flash of incorrect content. PR makes the dropdown either externally or internally controlled but never both.


pimato (approved):

Mi-Lan (approved):

KadenZipfel (approved):
Beautiful

Issue #1591, opened 2021-02-01
Show all relevant information until a tx is complete or cancelled


hexyls:
When buying, selling, etc we are clearing info from the UI before a transaction is complete/cancelled. As a user you prob want to see this info before confirming a transaction. Should clear info after tx confirmation/cancellation/error.

edit: think it is best to copy what happens with scalar markets - clear only when successful, leave info if the tx is cancelled so the user can retry if they want to.

<img width="569" alt="Screen Shot 2021-02-01 at 2 16 35 pm" src="https://user-images.githubusercontent.com/72715177/106411104-6a877180-6498-11eb-8f52-c9469efb6cd9.png">

PR #1592, opened 2021-02-01
fix/#1591

hexyls:
closes https://github.com/protofire/omen-exchange/issues/1591


pimato (approved):
perfect!

Mi-Lan (approved):
GJ!

pRoy24 (approved):
lgtm. I made this change as a fix to https://github.com/protofire/omen-exchange/issues/1581
But you are right we should not reset tx values until txn has finished. Better fix would be to not fire calcBuyAmount, calcSellAmount etc. once the txn has fired

PR #1597, opened 2021-02-01
Fix for 1581

pRoy24:
I had initially attempted to fix this by reseting the input fields once the transaction is fired.
While that fixes the issue, the resetting of input fields before the user has finished the transaction is a weird UX.

This is a better fix to not show amount error while the transaction is processing.


KadenZipfel (approved):
LGTM!

hexyls (approved):
🙌

Mi-Lan (approved):

Mi-Lan (dismissed):
looks good!

pimato (approved):

Issue #1512, opened 2021-01-07
Include pool data in scalar markets


KadenZipfel:
Categorical markets have a component which lists various values related to liquidity pooling, e.g. your liquidity, total liquidity, etc.. This should also be included in scalar markets.

KadenZipfel:
@pimato should we reuse the same pool data view used for categorical markets?

pimato:
> @pimato should we reuse the same pool data view used for categorical markets?

Yeah I agree.

I´ve also added infos regarding user receiving additional shares:
<img width="1796" alt="Fund-scalar" src="https://user-images.githubusercontent.com/5337809/106470210-85f67a80-64a0-11eb-940f-81927de391c0.png">

Figma:
[Design Specification](https://www.figma.com/file/aTivxuHmp2l6Sn8nho1Ngu/Playground?node-id=1588%3A47556)
[Playground](https://www.figma.com/proto/aTivxuHmp2l6Sn8nho1Ngu/Playground?node-id=1588%3A47556&viewport=206%2C260%2C0.5617057681083679&scaling=min-zoom)

KadenZipfel:
I've completed the pool data section (in pr above) and exported the additional shares section into this issue: https://github.com/protofire/omen-exchange/issues/1601 so I can refactor the value boxes as well.

PR #1600, opened 2021-02-02
Include pool data in scalar markets view

KadenZipfel:
Closes: https://github.com/protofire/omen-exchange/issues/1512


Mi-Lan (commented):
Screenshot 2021-02-02 at 21 56 52
Hmmm is the calculation correct after depositing 13 dai it shows only 5 dai as deposited...i know additional shares are bought in the background but is it that much?

pimato (approved):

hexyls (approved):
Same issue as @Mi-Lan but it looks like total user liquidity is being calculated the same way as it is in categorical markets.

Other than that looks great!

Mi-Lan (commented):
Maybe i'm overthinking it but follow me here, i emptied all deposited and bought scalar markets pt and shares then:
- I deposited 15 dai
Screenshot 2021-02-04 at 17 48 57
- Deposited amount 8.18DAI
Screenshot 2021-02-04 at 18 00 01
- Short shares 1.54DAI
Screenshot 2021-02-04 at 17 59 33
- Long shares 4.18DAI
Screenshot 2021-02-04 at 17 59 45

Which adds up to 13.9 DAI? Hmm where did that 1.1 DAI go? Am i missing something?

Mi-Lan (approved):
Looks good in consultation with Kaden it seems that between buying of shares market odds change which causes different prices...

pRoy24 (approved):
lgtm

PR #1629, opened 2021-02-05
Remove safeTxGas for safe apps

hexyls:
Removed some custom CPK stuff now that gas estimations are a lot better within the gnosis safe interface. Gotta keep sdk.txs.getBySafeTxHash and some other stuff for a bit longer tho.


pimato (approved):

pRoy24 (approved):
We are only setting the gas values for cases where the user is not signed into a gnosis safe and the collateral address is the native token address.

So I am assuming that the gas values for cases where the user is not signed into a gnosis safe and the collateral address is not native token address is handled automatically?

KadenZipfel (approved):

Mi-Lan (approved):

hexyls:
@pRoy24 yep that is exactly right! The CPK lib does a pretty good job at gas estimations, but on the rare occasion that it underestimates the gas for a native token transaction the ETH can get stuck in the CPK address.

If it underestimates the gas for a normal token transaction it doesn't really matter, the internal transactions will just fail and the user can try again with a higher manual gas setting 👍

PR #1636, opened 2021-02-09
remove authereum login option on xdai

hexyls:
as mentioned on keybase authereum does not support xDai so gotta hide it on the xdai subdomain.


pRoy24 (commented):
It still shows up as an option when on localhost when connected to xdai network and will also show up as an option on https://xdai.dxtest.eth.link/ Wondering if there is a better way to detect the current provider RPC URL in window context and disable based on that.

Mi-Lan (changes requested):
I agree with roy..but maybe we can just add additional locations with "OR" statment and cover all the cases ....

Mi-Lan (approved):
Screenshot 2021-02-11 at 13 05 43
works good!

KadenZipfel (approved):
👌

pimato (approved):

PR #1638, opened 2021-02-09
removed unit for scalar markets

Mi-Lan:
Screenshot 2021-02-10 at 00 36 51
Is this how its supposed to be just without unit in the market view with the one on top?
#1635


pimato (approved):

hexyls (approved):

app/src/components/market/common/list_item/index.tsx
```diff
@@ -138,8 +138,11 @@ export const ListItem: React.FC<Props> = (props: Props) => {
 
   let currentPrediction
   let unit
+  let scalarTitle
+
   if (isScalar) {
     unit = title.split('[')[1].split(']')[0]
+    scalarTitle = title.split(' [')[0]
```
KadenZipfel, line 145: We noticed in a previous PR that retrieving/removing the unit like that can cause a bug if there are square brackets used in the title at all as it will grab the wrong thing. To fix this we have `getUnit()`, which I see is actually unused here. Can you use `getUnit` for the unit here and also create and use a similar function to `getUnit` but for removing the unit from the title?

KadenZipfel (changes requested):

app/src/components/market/common/list_item/index.tsx
Mi-Lan: Good suggestion! Will use it

app/src/util/tools.ts
```diff
@@ -492,6 +492,11 @@ export const getUnit = (title: string): string => {
   const unit = splitTitle[splitTitle.length - 1].split(']')[0]
   return unit
 }
+export const getScalarTitle = (title: string): string => {
+  const unit = getUnit(title)
+  const scalarTitle = title.substring(0, title.length - (unit.length + 3))
+  return scalarTitle
+}
```
KadenZipfel, line 499: Looks good, just need a test for this function now

KadenZipfel (changes requested):

pRoy24 (approved):
lgtm

Issue #1640, opened 2021-02-10
App does not load on mobile phone browser/browser without web3 wallet


pRoy24:
The default chainid  for view is not defined.
The fix is to default chainid by hostname.
i.e. 100 for xdai.omen.eth.link and 1 for every other host in case chainid is not defined.

PR #1641, opened 2021-02-10
add default chainid for mobile browsers 

pRoy24:
Fixes the mobile display for UI

### DIFF
--- a/app/src/components/main/index.tsx
+++ b/app/src/components/main/index.tsx
@@ -29,23 +29,28 @@ import { MarketWizardCreatorContainer } from '../market/sections/market_create/m
 import { MarketHomeContainer } from '../market/sections/market_list/market_home_container'
 
 const RedirectToHome = () => <Redirect to="/" />
-
 export const Main: React.FC = () => {
   const context = useWeb3Context()
   const windowObj: any = window
-
-  const [networkId, setNetworkId] = useState(windowObj.ethereum.chainId)
+  const host = window.location.hostname
+  let defaultChainID = 1
+  if (host === XDAI_LOCATION) {
+    defaultChainID = 100
+  }
+  const [networkId, setNetworkId] = useState(windowObj.ethereum ? windowObj.ethereum.chainId : defaultChainID)
   const [wrongNetwork, setWrongNetwork] = useState(false)
 
-  windowObj.ethereum.on('chainChanged', (chainId: string) => {
-    setNetworkId(chainId)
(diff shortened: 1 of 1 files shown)


app/src/components/main/index.tsx
```diff
@@ -29,23 +29,28 @@ import { MarketWizardCreatorContainer } from '../market/sections/market_create/m
...
-  windowObj.ethereum.on('chainChanged', (chainId: string) => {
-    setNetworkId(chainId)
-    if (location.host === MAINNET_LOCATION && XDAI_NETWORKS.includes(chainId)) {
-      location.assign(`http://${XDAI_LOCATION}`)
-    }
-    if (location.host === XDAI_LOCATION && MAIN_NETWORKS.includes(chainId)) {
-      location.assign(`http://${MAINNET_LOCATION}`)
-    }
-  })
+  if (windowObj.ethereum) {
```
pRoy24, line 43: The default is only for the init case when there is no chain id.
Once the user logs in via a wallet, then windowObj.ethereum will be defined and this onChange handler would work as usual.

pRoy24:
Fixes https://github.com/protofire/omen-exchange/issues/1640

app/src/components/main/index.tsx
```diff
@@ -29,23 +29,28 @@ import { MarketWizardCreatorContainer } from '../market/sections/market_create/m
...
   const context = useWeb3Context()
   const windowObj: any = window
-
-  const [networkId, setNetworkId] = useState(windowObj.ethereum.chainId)
+  const host = window.location.hostname
+  let defaultChainID = 1
+  if (host === XDAI_LOCATION) {
+    defaultChainID = 100
+  }
+  const [networkId, setNetworkId] = useState(windowObj.ethereum ? windowObj.ethereum.chainId : defaultChainID)
```
Mi-Lan, line 40: One question why are we using defaultChainId variable and then networkId useState hook...why not just consolidate it to use hook and use setNetworkId to set default ones?

KadenZipfel (approved):

pimato:
IMG_FD00C7D2BF7B-1

tried it on my mobile and it does not seem to work.

pimato (approved):

app/src/components/main/index.tsx
pRoy24: NetworkID is not the same as chainid, they have the same values but they are not the same entity.
https://ethereum.stackexchange.com/questions/37533/what-is-a-chainid-in-ethereum-how-is-it-different-than-networkid-and-how-is-it

hexyls (approved):

Issue #1601, opened 2021-02-02
Display additional shares during scalar LPing and refactor value boxes


KadenZipfel:
In addition to displaying additional shares, I will refactor the value boxes implementation since its usage has expanded beyond its initial implementation.

![image](https://user-images.githubusercontent.com/30579067/106647966-77619f00-6544-11eb-8511-0752b5a07dbd.png)

Figma:
[Design Specification](https://www.figma.com/file/aTivxuHmp2l6Sn8nho1Ngu/Playground?node-id=1588%3A47556)
[Playground](https://www.figma.com/proto/aTivxuHmp2l6Sn8nho1Ngu/Playground?node-id=1588%3A47556&viewport=206%2C260%2C0.5617057681083679&scaling=min-zoom)

PR #1645, opened 2021-02-11
Display additional shares during scalar LPing

KadenZipfel:
Closes: https://github.com/protofire/omen-exchange/issues/1601


hexyls (approved):

pimato (approved):

app/src/components/market/common/market_scale/value_box.tsx
```diff
@@ -0,0 +1,126 @@
...
+        ? `left: ${
+            props.xValue <= 0.115
+              ? `0`
+              : props.xValue <= 0.885
+              ? `${props.xValue * 100}%; transform: translateX(-50%);`
+              : ``
+          }`
+        : `right: 0;`
+      : ''}
+  background: white;
```
Mi-Lan, line 51: should we maybe use value from theme here? and also closing column above it for the xValue...
Screenshot 2021-02-12 at 12 36 09

app/src/components/market/common/market_scale/value_box.tsx
```diff
@@ -0,0 +1,126 @@
...
+  position: absolute;
+  top: calc(${SCALE_HEIGHT} + ${VALUE_BOXES_MARGIN});
+  border-radius: 4px;
+`
+
+const ValueBoxTitle = styled.p<{ positive?: boolean | undefined }>`
+  font-size: 14px;
```
Mi-Lan, line 58: this value is also default font-size from theme but i would suggest adding just one font-size in wrapper valueBoxSegment so we don't have to repeat it for all the children...

app/src/components/market/common/market_scale/value_box.tsx
```diff
@@ -0,0 +1,126 @@
...
+    border-bottom-right-radius: 4px;
+    border-top-left-radius: 0px;
+    border-bottom-left-radius: 0px;
+    border-left: none;
+  }
+`
+
+const ValueBoxSingle = styled.div<{ xValue?: number }>`
+  padding: 12px;
+  border: 1px solid ${props => props.theme.scale.box};
```
Mi-Lan, line 35: also here i think we have borderLineDisabled in theme which has 1px solid with the desired color...

Mi-Lan (changes requested):
Great job! Works like charm 🥇  Just added couple of small css things i found

Mi-Lan (approved):

Issue #1527, opened 2021-01-10
Remove arbitrary 15 refetches of market maker data


hexyls:
Market maker data is being refetched [at least 15 times](https://github.com/protofire/omen-exchange/blob/master/app/src/hooks/useGraphMarketMakerData.tsx#L266). If anything awaits this function is takes ages to resolve making the user wait longer than they need to.

KadenZipfel:
I know that the intended effect is to refetch a _maximum_ of 15 times, and using the `needRefetch` check to see if a refetch is necessary. The reasoning for setting it up this way is to wait for the subgraph to update, and to stop refetching once it is updated. I think the caveat though is that if there is no subgraph update it goes through 15 fetches, so there's probably a way to make this more efficient without sacrificing the ability to wait for the subgraph to update as it's a pretty messy solution anyways.

PR #1646, opened 2021-02-11
Fix/scalar over fetch of user tx data

hexyls:
When we refetch data in `useGraphMarketUserTxData` it currently [expects both trade data and liquidity data](https://github.com/protofire/omen-exchange/blob/master/app/src/hooks/useGraphMarketUserTxData.tsx#L187) to change before it stops refetching. Buys/sells don't change user liquidity data, so it keeps refetching until the max 15 retries is up.

PR removes refetch logic, waitForBlockToSync in the CPK service makes sure data is up to date by the time we try to refetch data.

closes https://github.com/protofire/omen-exchange/issues/1527


pimato (approved):

KadenZipfel (approved):
🔥 🔥 🔥  
`waitForBlockToSync` is a game changer. Great call on this! I think there are still some similar hooks using refetches which this can be applied to as well, right?

hexyls:
Yeah good point @KadenZipfel , `useGraphMarketMakerData` does a similar thing so just updated that too  👍

pimato (approved):

Mi-Lan (approved):

KadenZipfel (approved):

Issue #1652, opened 2021-02-15
Move Compound Service to a hook so as to not have to pass it down as props to each of the categorical and scalar market actions.


pRoy24:
While developing the compound integration for Scalar markets noticed that the Scalar markets have separate components for each of the buy/sell outcome tokens and pool tokens functionalities.
To prevent having to pass down the compoundService object to each of the new components as well, better to move that functionality to a hook which each of those components can subscribe to.

PR #1656, opened 2021-02-15
Fix/1652

pRoy24:
Resolved https://github.com/protofire/omen-exchange/issues/1652
Replaced the compound service which was being passed down as a prop with a hook call.
This is because the Scalar market components are in different files and will need to duplicate and pass it down multiple times to support scalar market.
This would be a cleaner approach.


pimato (approved):

app/src/hooks/useCompoundService.tsx
```diff
@@ -0,0 +1,45 @@
...
+  }, [collateral.address, account, collateral.symbol, provider])
+
+  useEffect(() => {
+    if (collateral.symbol.toLowerCase() in CompoundTokenType) {
+      const compoundService = new CompoundService(collateral.address, collateral.symbol, provider, account)
+      setCompoundService(compoundService)
+      fetchCompoundService()
+    }
+    // eslint-disable-next-line
+  }, [])
```
KadenZipfel, line 42: Do we need the `useEffect` here? Doesn't `useMemo` run on mount?

Mi-Lan (approved):
Works good!

app/src/hooks/useCompoundService.tsx
pRoy24: yeah useMemo does indeed run on mount, however I tested without it and the app doesn't get the initial values of compoundExchangeRate.
I assume it is because it uses a memoized value of collateral and since it called in the entry level component to display the top section value market_status_open the memoized value of collateral is still null,
That component (market_status_open and market_status_closed) uses the on mount effect to get the values.
Everywhere else the value is fetched using the momized value returned by useMemo.

KadenZipfel (approved):

Issue #1659, opened 2021-02-17
Discrepancy of "24h volume"


KeenanLukeOM:
Originally pointed out by Kobello on Keybase, there is a discrepancy between the filter and entries for volume. Filter shows "24h volume", while entries show "24hr Volume".

"24hr Volume" likely makes sense for consistency.

![image](https://user-images.githubusercontent.com/76129769/108138578-1d4bf800-7073-11eb-88ba-ca9fbfd9e7b9.png)

PR #1663, opened 2021-02-17
Replace instances of '24h volume' with '24hr volume'

KadenZipfel:
Closes: https://github.com/protofire/omen-exchange/issues/1659

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "app",
-  "version": "1.2.2",
+  "version": "1.2.3",
   "private": true,
   "engines": {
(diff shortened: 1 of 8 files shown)


hexyls (approved):

Mi-Lan (approved):

pimato:
I would actually go with `24h volume` and not `24hr volume` it is just more common, for example coinbase pro:
Bildschirmfoto 2021-02-18 um 15 57 34

pimato (changes requested):
see comment

pRoy24 (approved):
LG

pimato (approved):

Issue #1594, opened 2021-02-01
New Sell Position View


pimato:
![New-Sell-Position](https://user-images.githubusercontent.com/5337809/106466200-52652180-649b-11eb-9c16-ba88604a9b86.png)

Changes:
- Remove horizontal long/short button-picker
- Remove Balance component
- Add Radio picker list with Short/Long as option
- If user user has no position in a position-type, radio button should be disabled
- Replace Radio-like slider with new slider-dot design

Figma:
- [Design Specification](https://www.figma.com/file/R056tSYMjvfwaFidEBItYy/Widgets-V2?node-id=3270%3A44)

PR #1670, opened 2021-02-19
New scalar sell position view

KadenZipfel:
Closes: https://github.com/protofire/omen-exchange/issues/1594

One thing that's worth considering for this feature is whether we should somehow indicate what the current prediction is, or how much slippage the trade causes.


pimato (changes requested):
can we get an no-touchable-icon on hover:
Bildschirmfoto 2021-02-20 um 11 25 44

app/src/theme/index.js
```diff
@@ -108,6 +108,7 @@ const theme = {
     verticalDivider: '#e8eaf6',
     clickable: '#7986CB',
     red: '#E57373',
+    black: '000',
```
pRoy24, line 111: This should be #000 right?

app/src/util/tools.ts
```diff
@@ -597,3 +597,11 @@ export const getInitialCollateral = (networkId: number, collateral: Token): Toke
     }
   }
 }
+
+export const reverseArray = (array: any[]): any[] => {
```
pRoy24, line 601: A more concise way to do this would be 
const newArray = array.map((e, i, a)=> a[(a.length -1) -i])

app/src/components/market/common/position_selection_box/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+const PositionSelectionBoxWrapper = styled.div`
+  height: 88px;
+  border: ${props => props.theme.borders.borderLineDisabled};
+  border-radius: ${props => props.theme.cards.borderRadius};
+  margin-bottom: 20px;
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+  padding: 16px 20px;
```
Mi-Lan, line 16: This values could be imported from theme theres paddingVertical and paddingHorizontal value...

app/src/theme/index.js
KadenZipfel: Good eyes!

app/src/components/market/common/position_selection_box/index.tsx
KadenZipfel: I actually think we shouldn't import these values because they're not consistent between vertical and horizontal. For example, we have `cards.paddingHorizontal: '20px'` and `cards.paddingVertical: '12px'`, which is inconsistent with ours at `16px 20px` meaning it probably shouldn't be classified as a card. If we were to create another element with similar dimensions, I would say it'd be good to create a whole `theme.positionBox` object. Overall, the theme file doesn't have a consistent system atm which should really be fixed

hexyls (approved):
holy moly dude you are so damn fast! this looks great 🏋️‍♂️

Mi-Lan (approved):

pimato:
Still see:
Bildschirmfoto 2021-02-23 um 13 15 07

pimato (changes requested):
see comment

KadenZipfel:
> Still see:
> Bildschirmfoto 2021-02-23 um 13 15 07

Think my `git push` didn't quite work... Should be fixed now

pimato (approved):

Issue #1669, opened 2021-02-19
Scalar market creation UI bug


KadenZipfel:
The 'starting point' box should be under the scale knob.
![image](https://user-images.githubusercontent.com/30579067/108564645-4c9f7680-72b8-11eb-864e-d128fc562c9d.png)

PR #1671, opened 2021-02-19
Fix scalar market creation UI bug

KadenZipfel:
Closes: https://github.com/protofire/omen-exchange/issues/1669


app/src/components/market/common/market_scale/index.tsx
```diff
@@ -489,7 +489,7 @@ export const MarketScale: React.FC<Props> = (props: Props) => {
       subtitle: startingPointTitle,
       xValue: currentPrediction
         ? Number(currentPrediction)
-        : calcXValue(startingPoint || new BigNumber(0), lowerBound, upperBound, 18),
+        : calcXValue(startingPoint || new BigNumber(0), lowerBound, upperBound, 18) / 100,
```
pRoy24, line 492: Does this value 18 represent token decimals? If so...dangerous to hardcode to 18

hexyls (approved):

Mi-Lan (approved):
Screenshot 2021-02-22 at 17 39 17
Looks good!

app/src/components/market/common/market_scale/index.tsx
KadenZipfel: Thanks for pointing that out. The decimal value is always 18 since it just represents values unrelated to tokens. I've removed the decimals parameter to reflect this

pimato (approved):

Issue #1637, opened 2021-02-09
Allow user to create market using cToken collateral for scalar markets


pRoy24:
The current compound integration targets categorical markets.
Once resolved, we should allow the same integration for scalar markets as well.

Issue #1675, opened 2021-02-23
User should receive native xDai and not wrapped xDai after selling a position/outcome tokens


pimato:
Currently, if a user sells a position or outcome token he/she receives wrapped xDai. With the tighter xDai integration coming, we should try our best to always use the native assets. Therefor, we should change the receive asset type from wrapped xdai to native xdai on sell outcome/sell position widget.

Issue #1507, opened 2021-01-06
Add option to sell or remove liquidity as native asset


KadenZipfel:
Currently, it's possible to buy and add funding to a market in the native asset, but not so when it comes to selling or withdrawing.

hexyls:
This might be related to https://github.com/protofire/omen-exchange/issues/1439

Do we want to abstract away wrapped assets in general? Users fund with ETH and withdraw to ETH and wrapping is just handled in the background?

The only downside is losing convenience if a user already has a wrapped asset in their wallet.

Or do we want to keep both?

@Mi-Lan @pRoy24 thoughts?

pimato:
https://github.com/protofire/omen-exchange/pull/1783 will fix this

Issue #1654, opened 2021-02-15
Allow base or wrapped asset withdrawal for Scalar markets


pRoy24:
When using pseudo native asset for scalar markets, currently withdrawal is only possible as the wrapped asset. We need to allow withdrawal in both base and wrapped asset.
This feature is already live for categorical markets, needs to be put in place for scalar markets as well.

PR #1677, opened 2021-02-24
Add compound integration for scalar markets

pRoy24:
Resolves
 https://github.com/protofire/omen-exchange/issues/1637
https://github.com/protofire/omen-exchange/issues/1654


app/src/components/market/sections/market_sell/scalar_market_sell.tsx
```diff
@@ -187,7 +252,7 @@ export const ScalarMarketSell = (props: Props) => {
       setAmountShares(null)
       setAmountSharesToDisplay('')
```
hexyls, line 253: I think we might have to zero out `displaySellShares` here after a sell.

app/src/components/market/sections/market_buy/scalar_market_buy.tsx
```diff
@@ -240,17 +281,25 @@ export const ScalarMarketBuy = (props: Props) => {
...
+        useBaseToken = true
+        inputAmount = compoundService.calculateCTokenToBaseExchange(baseCollateral, amount)
+      }
+      const sharesAmount = formatBigNumber(displayTradedShares, baseCollateral.decimals)
       setStatus(Status.Loading)
       setMessage(`Buying ${sharesAmount} shares ...`)
 
       await cpk.buyOutcomes({
-        amount,
+        amount: inputAmount,
```
hexyls, line 297: One thing I found was that if I am buying off a cDai market with cDai as my selected currency, it will convert my Dai to cDai instead of using my existing cDai.

hexyls (changes requested):

app/src/components/market/sections/market_sell/scalar_market_sell.tsx
pRoy24: Yes good catch. Fixed

app/src/components/market/sections/market_buy/scalar_market_buy.tsx
pRoy24: Fixed

pimato:
fixes: https://github.com/protofire/omen-exchange/issues/1675

pimato:
also fixed: https://github.com/protofire/omen-exchange/issues/1507

pimato (changes requested):
Looks like we keep always some dust even after selling all of our Shares:
Bildschirmaufnahme 2021-03-01 um 22 16 32

Mi-Lan (commented):
@pRoy24 Good job tested all the features mentioned and it works! I was just wondering why am i getting eth and cEth instead of weth for wrapped asset?
Screenshot 2021-03-03 at 12 51 38
- Also i don't know if this issue is related to this pr but while i have eth selected im prompted to setAllowance for cEth we had similar issue  in the past if i remember?
Screenshot 2021-03-03 at 12 53 15

pRoy24:
> Looks like we keep always some dust even after selling all of our Shares:
> Bildschirmaufnahme 2021-03-01 um 22 16 32

A bit tricky to get rid of all of the dust since we there is difference in precision in the token and cToken. I'll give it another shot later today.

pRoy24:
> @pRoy24 Good job tested all the features mentioned and it works! I was just wondering so when market has compound there is only option to withdraw in eth or cEth... no weth?
> Screenshot 2021-03-03 at 12 51 38
> 
> * Also i don't know if this issue is related to this pr but while i have eth selected im prompted to setAllowance for cEth we had similar issue  in the past if i remember?
>   Screenshot 2021-03-03 at 12 53 15

yes for ctoken pools only base token/ctoken. Since the ctoken to basetoken conversion is an atomic swap, allowing wrapping the native asset would lead to lose conditional logic on cpk service and component side.
For base token/wrapped token pools deposit/withdrawal in base token or in ctoken.

CToken -> Wrapped token or Wrapped token to cToken conversions are currently not supported.

***
eth selected im prompted to setAllowance for cEth we...
This has been fixed now.

KadenZipfel:
> > Looks like we keep always some dust even after selling all of our Shares:
> > Bildschirmaufnahme 2021-03-01 um 22 16 32
> 
> A bit tricky to get rid of all of the dust since we there is difference in precision in the token and cToken. I'll give it another shot later today.

You can also look at possibly modifying `isDust()` to be suitable for this kind of thing if greater precision isn't possible, and that way we don't have the position appearing in the market details view.

Mi-Lan (changes requested):
@pRoy24 Seems that i'm still getting the cEth allowance while on eth....
Screenshot 2021-03-09 at 14 04 04
Also same for dai and cDai
Screenshot 2021-03-09 at 13 59 49

pRoy24:
> @pRoy24 Seems that i'm still getting the cEth allowance while on eth....
> Screenshot 2021-03-09 at 14 04 04
> Also same for dai and cDai
> Screenshot 2021-03-09 at 13 59 49

Aah yes, I thought I fixed it but seems it got reverted when I reset to origin while testing fix for the maximum value sell.

pimato (changes requested):
just created a scalar market with cDai as collateral and it looks like I can`t sell max amount it always gives me an error.
if I can`t use the sell max how am I gonna test the fix for not having dust?

app/src/components/market/sections/market_buy/scalar_market_buy.tsx
```diff
@@ -276,11 +328,39 @@ export const ScalarMarketBuy = (props: Props) => {
...
   }
 
-  const currencyFilters =
+  let currencyFilters =
     collateral.address === wrapToken.address || collateral.address === pseudoNativeAssetAddress
       ? [wrapToken.address.toLowerCase(), pseudoNativeAssetAddress.toLowerCase()]
       : []
 
+  if (collateralSymbol in CompoundTokenType) {
+    if (baseCollateral.symbol.toLowerCase() === 'eth') {
```
KadenZipfel, line 337: We could use `networks[networkId].nativeAsset` here as well as a few other places where this is used

KadenZipfel (changes requested):
Great work so far on a big PR 💪  everything seems to work pretty well for me in testing, just made some reusability/maintenance suggestions.

Also I noticed that the payout/profit/loss values on the slider remain denominated in the base token even when the cToken is selected, causing a mismatch in values with the transaction details view. But I think this may be fine to move into a separate issue
image

app/src/components/market/sections/market_buy/scalar_market_buy.tsx
```diff
@@ -230,7 +268,11 @@ export const ScalarMarketBuy = (props: Props) => {
   const isBuyDisabled =
     (status !== Status.Ready && status !== Status.Error) ||
     amount.isZero() ||
-    (!cpk?.isSafeApp && collateral.address !== pseudoNativeAssetAddress && hasEnoughAllowance !== Ternary.True) ||
+    Number(sharesTotal) == 0 ||
```
KadenZipfel, line 271: I noticed this line causes this issue https://github.com/protofire/omen-exchange/issues/1722 it should be fine without it since we already have `amount.isZero()`. Same thing on `scalar_market_sell`, I think

KadenZipfel (changes requested):

pRoy24:
> Great work so far on a big PR 💪 everything seems to work pretty well for me in testing, just made some reusability/maintenance suggestions.
> 
> Also I noticed that the payout/profit/loss values on the slider remain denominated in the base token even when the cToken is selected, causing a mismatch in values with the transaction details view. But I think this may be fine to move into a separate issue
> image

Resolved all other comments, except for "Postion values are displayed in base collateral even if cToken is selected"
This is in accordance with guidelines provided by @pimato who stated that when creating compound market all display values should be in base token regardless of which token is chosen to transact in.

KadenZipfel:
I'm trying to create a USDC scalar market with the compound integration and getting, `batch transaction execution expected to fail`. I suspect it's some kind of decimal issue since I can create a regular USDC scalar market and a DAI market with the compound integration, but not USDC with compound

KadenZipfel:
Maybe an issue with rebasing, but I'm getting this error when I checkout the branch:
image

Mi-Lan:
Hey @pRoy24 wondering if this is the way its supposed to work. So it seems that it does couple of conversions from cDai to Dai then again to cDai then to Dai. When selling on compound market and receiving DAI. Here is transaction hash for that https://rinkeby.etherscan.io/tx/0x7329cac289cd77dff5be810bb24fa2bf9914f6e7b877640c02068997863b411f
Screenshot 2021-04-12 at 11 47 52

Mi-Lan:
Would be good to fix these compilation warnings before push...
Screenshot 2021-04-12 at 12 35 25

Mi-Lan (approved):
@pRoy24 Good job haven't had any issues with it! Heres what i tested:
- Scalar market creation in cDai
- Buying shares 
- Selling shares and getting them in cDai and Dai
- Depositing and Withdrawing
- Redeeming scalar market

hexyls (approved):

pimato (approved):
Works great!
The only issue I have found is the user will get cDai back and not they contributed Dai. Should be fixed in a separate issue here: https://github.com/protofire/omen-exchange/issues/1850

PR #1674, opened 2021-02-23
Fix failing tests

KadenZipfel:
Closes: https://github.com/protofire/omen-exchange/issues/1672

Notes: 
- A lot of tests were failing. We need to stay on top of them to make sure they're up to date and things aren't breaking.
- The pre-push hook isn't running for me, so this is partly incomplete, but I wanted to see if it runs for others to help troubleshoot.
- The outcome related tests no longer apply so they've been removed.

PR #1433, opened 2020-12-10
Feature/#1358

Mi-Lan:
- To test claim states uncomment code on components/common/claim_dai_modal and change number by hand to see how it will react....
- This will need to be deleted before the realease

PR #1678, opened 2021-02-24
Update pre-push test script

KadenZipfel:
Continuation of https://github.com/protofire/omen-exchange/pull/1674 to properly run the test script


Mi-Lan:
@KadenZipfel i included it in the latest Bridge pr since i couldn't push otherwise didn't know how to get over that screen https://github.com/protofire/omen-exchange/pull/1433

Issue #1687, opened 2021-03-01
Scalar market page crashes if all liquidity is pulled out and no further action is possible subsequently


pRoy24:
Steps to reproduce-
Create a scalar market with some initial liquidity.
Go to pool page and withdraw all of the liquidity.
The market page will crash on that and subsequent page loads and you will not be able to do any further actions on that market.

Error
×
TypeError: Reduce of empty array with no initial value

Log

> 223 | const totalDepositedTokens = totalUserShareAmounts.reduce((min: BigNumber, amount: BigNumber) =>
      | ^  224 |   amount.lt(min) ? amount : min,
  225 | )
  226 | 
  
  04 |   setMarketMakerData(newMarketMakerData)
      | ^  205 |   setStatus(Status.Ready)
      
 On refreshing the page the outcome table shows undefined/undefined

If you go to pool page, the view will crash again.

KadenZipfel:
Can confirm, just reproduced

pimato:
great catch! Just replicated!

PR #1689, opened 2021-03-01
Fix scalar market crash on no liquidity

KadenZipfel:
Closes: https://github.com/protofire/omen-exchange/issues/1687


pimato (approved):
works!

hexyls (approved):

pRoy24 (approved):
LGTM.

1 small issue I see is that upon removing all liquidity the prediction shows as unknown (Expected) but if I add liquidity to a market with no liquidity the prediction doesn't immediately switch back to normal value without me having to hit refresh on the browser.

pimato:
hm... might be some delay with the subgraph/rpc node but I did not need a refresh:
Bildschirmaufnahme 2021-03-02 um 10 43 17

PR #1693, opened 2021-03-02
created market meta data button which stores info that would overflow

### LINKED ISSUE #1642: Market Meta-information design V2
With our new compound integration the meta-information component needs a better way to show all relevant meta information for a specific market. If the horizontal line exceeds the width space, we need to add an horizontal more icon which will allow the user to still get the information he wants. Currently:

Bildschirmfoto 2021-02-10 um 15 48 47

Should:
meta-info

Figma:
[Playground](https://www.figma.com/proto/aTivxuHmp2l6Sn8nho1Ngu/Playground?node-id=1659%3A42273&viewport=179%2C630%2C0.273661732673645&scaling=min-zoom)
[Design Specification](https://www.figma.com/file/aTivxuHmp2l6Sn8nho1Ngu/Playground?node-id=1659%3A41927)

gem1no:
fixes #1642 
 if either compound token is detected, or available width becomes too small, a "。 。 。" button is generated which will hold meta data for items which don't fit. mobile view changes the modal to a vertically stacked list of further condensed items.

https://i.imgur.com/L0MCJMA.gif


app/src/components/market/common/additional_market_data/index.tsx
```diff
@@ -155,84 +196,222 @@ export const AdditionalMarketData: React.FC<Props> = props => {
...
-            isError={!curatedByDxDaoOrKleros}
-          >
-            {curatedByDxDaoOrKleros ? <IconVerified size={'24'} /> : <IconAlert size={'24'} />}
-            <AdditionalMarketDataSectionTitle isError={!curatedByDxDaoOrKleros}>
-              {curatedByDxDaoOrKleros ? 'Verified' : 'Not Verified'}
-            </AdditionalMarketDataSectionTitle>
-          </AdditionalMarketDataSectionWrapper>
-        )}
+
+  if (window.screen.width > 552) {
```
hexyls, line 200: Is there a way make these changes without duplicated code, using media queries maybe?

@KadenZipfel @pRoy24 @Mi-Lan thoughts?

hexyls (commented):
Hey man sorry for the slow review!

This is looking great 🙌 see comment on maybe attempting this change without duplicating code.

app/src/components/market/common/additional_market_data/index.tsx
Mi-Lan: Yeah i agree !

Mi-Lan (commented):
Yeah there is a probably a better to handle this. And keeping in mind very important principle in programming "Don't repeat yourself"(DRY). Duplicate code causes lot of issues when changing something down the line.

Mi-Lan (commented):
This is interesting issue. Since its not guaranteed that items will overflow just below certain width or when compound is present. I think challenge here is to figure out when items overflow and then which items are overflowing so we know what to place in dropdown and not show in additionalMarketDataSection...Thats how i would go about it...
I think this might help https://stackoverflow.com/questions/42012130/how-to-detect-overflow-of-react-component-without-reactdom or maybe even better https://www.npmjs.com/package/react-overflow? @KadenZipfel @pRoy24 @hexyls

app/src/components/market/common/additional_market_data/index.tsx
pRoy24: You can use media queries with styled components.
https://jsramblings.com/how-to-use-media-queries-with-styled-components/

pimato:
I am gonna close this for now until we have the bandwith to get back on this.

PR #1702, opened 2021-03-04
Fix scalar redeem issue

KadenZipfel:
Fix of a bug that prevented earned amount on scalar markets from being calculated in certain cases


Mi-Lan (approved):
I didn't know exactly how to test it. But it seems reasonable and i believe that you know what you are doing  :D

KadenZipfel:
> I didn't know exactly how to test it. But it seems reasonable and i believe that you know what you are doing :D

If it helps, the expected effect is that it always returns null if there are no balances, but should never return null if there are any balances.

hexyls (approved):

pRoy24 (approved):
lg

Issue #1703, opened 2021-03-04
Implement getBlockExplorerURL


KadenZipfel:
In https://github.com/protofire/omen-exchange/pull/1698 I've implemented a `getBlockExplorerURL` function. Since https://github.com/protofire/omen-exchange/pull/1698 is already a very large PR, I will implement the function everywhere else needed in a separate PR for simpler testing.

PR #1705, opened 2021-03-04
Implement getBlockExplorerURL

KadenZipfel:
Closes: https://github.com/protofire/omen-exchange/issues/1703
Depends on: https://github.com/protofire/omen-exchange/pull/1741


app/src/util/tools.ts
```diff
@@ -758,7 +758,7 @@ export const reverseArray = (array: any[]): any[] => {
   return newArray
 }
 
-export const getBlockExplorerURL = (networkId: number, txHash: string): string => {
+export const getTxHashBlockExplorerURL = (networkId: number, txHash: string): string => {
```
Mi-Lan, line 761: Don't know if it's scope of this issue but i would add a suggestion that all etherscan links are stored in networks and that we make a function that fetches it based on network like we did for bunch of stuff....like for example getChainSpecifiAlternativeUrls? @KadenZipfel @hexyls what do you think?

Mi-Lan (commented):
Will give it another round of review after it's merged with master...Theres bunch of left out code

app/src/util/tools.ts
KadenZipfel: Good call! I've moved it to the networks file

Mi-Lan (commented):
Should we remove EtherscanLink enum from types.ts?

Mi-Lan (approved):
Other than that small thing should be GTG !!!!!

KadenZipfel:
> Should we remove EtherscanLink enum from types.ts? Seems that it's unused now

Ahhh, yes of course! Removed

PR #1709, opened 2021-03-08
fixed decimal places showing when not needed scalar market

### LINKED ISSUE #1404: Scalar market creation numerical input display issue
When creating a scalar market and going back from the `funding_and_fee` step to the `ask_question_step`, the numerical inputs behave imperfectly. Inputs with the value "0" are entirely empty, and other numerical inputs have a ".0" appended to the end of whatever value exists. Fixing this up would make things a lot cleaner.

Mi-Lan:
closes #1404 
- i banged my head on this issue for some time now...i managed to get it to not show those decimals when present but still got to resolve zero showing placeholder when going back...
- something has to be changed in big_number_input with updateValue function or useEffect hook i couldn't figure it out tried lots of stuff...


KadenZipfel (approved):
Yeah, I tried messing around here a bit to try to get 0's to display, but couldn't get it. Removing the decimals here is a good improvement still. 👍

hexyls (approved):

Mi-Lan:
> Yeah, I tried messing around here a bit to try to get 0's to display, but couldn't get it. Removing the decimals here is a good improvement still. 👍

Dunno if we should leave this pr open until someone resolves it or merge it like this with partial solution?

KadenZipfel:
> > Yeah, I tried messing around here a bit to try to get 0's to display, but couldn't get it. Removing the decimals here is a good improvement still. 👍
> 
> Dunno if we should leave this pr open until someone resolves it or merge it like this with partial solution?

imo we should merge this and create a new issue for just the 0 issue.

pRoy24 (approved):
LG

Mi-Lan:
As Kaden suggested i created issue for the one thing left to be done https://github.com/protofire/omen-exchange/issues/1716 so someone could take a stab at it...

PR #1713, opened 2021-03-09
form validation for Violets review

kazuolol:
image

image

### DIFF
--- a/app/yarn.lock
+++ b/app/yarn.lock
@@ -22207,9 +22207,9 @@ sshpk@^1.7.0:
     tweetnacl "~0.14.0"
 
 ssri@^6.0.1:
-  version "6.0.1"
-  resolved "https://registry.yarnpkg.com/ssri/-/ssri-6.0.1.tgz#2a3c41b28dd45b62b63676ecb74001265ae9edd8"
-  integrity sha512-3Wge10hNcT1Kur4PDFwEieXSCMCJs/7WvSACcrMYrNp+b8kDL1/0wJch5Ni2WrtwEa2IO8OsVfeKIciKCDx/QA==
+  version "6.0.2"
+  resolved "https://registry.yarnpkg.com/ssri/-/ssri-6.0.2.tgz#157939134f20464e7301ddba3e90ffa8f7728ac5"
+  integrity sha512-cepbSq/neFK7xB6A50KHN0xHDotYzq58wWCa5LeWqnPrHG8GzfEjO/4O8kpmcGW+oaxkvhEJCWgbgNk4/ZV93Q==
   dependencies:
     figgy-pudding "^3.5.1"
 


app/src/components/market/sections/market_create/steps/items/create_scalar_market.tsx
```diff
@@ -125,6 +137,35 @@ export const CreateScalarMarket = (props: Props) => {
     upperBound,
   } = props
 
+  const lowerBoundNumber = lowerBound?.abs() !== undefined && Number(lowerBound.abs())
```
pRoy24, line 140: I think just lowerBound !== undefined should work, also casting BigNumber to Number directly is dangerous and can often lead to precision errors.
Use formatBigNumber instead with a precision value of 2 and then cast result to Number type.

pRoy24 (changes requested):

app/src/components/market/sections/market_create/steps/items/create_scalar_market.tsx
```diff
@@ -76,6 +77,14 @@ const NumericalInput = styled(BigNumberInput)`
...
 `
 
+const NumericalInputError = styled.div<{ error?: string }>`
+  border-color: ${props => (props.error ? props.theme.colors.error : 'display: none')};
+  border-style: ${props => (props.error ? props.theme.textfield.borderStyle : 'display: none')};
+  border-width: ${props => (props.error ? props.theme.textfield.borderWidth : 'display: none')};
+  border-radius: ${props => (props.error ? props.theme.textfield.borderRadius : 'display: none')};
+  border: ${props => (props.error ? 'display: flow-root' : 'display: none')};
+`
+
```
Mi-Lan, line 87: You don't need this styled component its just creating you issue as you see in image below it has some weird border. You need to change just border color property conditionally in Numerical input.
Screenshot 2021-03-16 at 14 21 53

Mi-Lan (commented):
Screenshot 2021-03-16 at 14 35 56
Also i think it shouldn't be displaying error below lower bound in this case. @pimato What do you think should we maybe disable negative numbers for these input fields?

Mi-Lan:
Screenshot 2021-03-16 at 14 28 52
Also you shouldn't use display:none; when you don't want property to show...it does work but in a way that its unsupported value...
Instead if this   border-color: ${props => (props.error ? props.theme.colors.error : 'display: none')};
you can do this ${props => (props.error && `border-color:${props.theme.colors.error}` )}; with the help of backtick

app/src/components/common/form/form_error/index.tsx
```diff
@@ -3,11 +3,14 @@ import styled from 'styled-components'
...
-  margin: 0 0 10px;
-  text-align: left;
+  font-family: Roboto;
+  font-style: normal;
+  font-weight: normal;
+  font-size: 14px;
+  line-height: 16px;
+  display: flex;
+  align-items: center;
+  letter-spacing: 0.2px;
```
Mi-Lan, line 13: Nicely done changing properties here! Also you can omit display flex and import font-size from theme...

app/src/theme/index.js
```diff
@@ -88,7 +88,7 @@ const theme = {
     alertHover: '#EF5350',
     borderColorDark: '#9FA8DA',
     darkGray: '#acacac',
-    error: '#fa0000',
+    error: '#e57373',
```
Mi-Lan, line 91: also there is this color in the theme already so instead of changing color itself import different one ....as you can see above there is 'alert' color which is the same as one you need

app/src/components/market/sections/market_create/steps/items/create_scalar_market.tsx
gem1no: good catch. i tend to overcomplicate things, this is a much more elegant approach.

gem1no:
prettier errors happening because i edited the merge  conflict in github. i will submit a new PR soon

gem1no:
> Screenshot 2021-03-16 at 14 35 56
> Also i think it shouldn't be displaying error below lower bound in this case. @pimato What do you think should we maybe disable negative numbers for these input fields?

requiring Number(startingPointNumber) > 0 for the logic that checks if error should display fixes this "must be less than 0" errors

app/src/components/market/sections/market_create/steps/items/create_scalar_market.tsx
```diff
@@ -125,6 +120,62 @@ export const CreateScalarMarket = (props: Props) => {
...
 
+  let lowerBoundError
+  let startingPointError
+  let upperBoundError
+  let startingPointNumber
+  let lowerBoundNumber
+  let upperBoundNumber
+  // if the input values aren't null the number is cast to BigNumber with 18 decimals and precision value of 2
+  if (startingPoint !== null) {
+    startingPointNumber = formatBigNumber(startingPoint, 18, 2)
```
KadenZipfel, line 131: You can simplify this to, `startingPointNumber = startingPoint && formatBigNumber(startingPoint, 18, 2)`, same for the other values including the `!== undefined` checks

app/src/components/market/sections/market_create/steps/items/create_scalar_market.tsx
```diff
@@ -125,6 +120,62 @@ export const CreateScalarMarket = (props: Props) => {
...
+
+  if (lowerBound !== null) {
+    lowerBoundNumber = formatBigNumber(lowerBound, 18, 2)
+  }
+
+  if (upperBound !== null) {
+    upperBoundNumber = formatBigNumber(upperBound, 18, 2)
+  }
+  //check for values to be non negative integers
+  if (startingPointNumber !== undefined && startingPointNumber < '0') {
```
KadenZipfel, line 142: Should be `<= 0` otherwise it will allow a 0 value. Also `'0'` shouldn't be in quotes; probably works fine, but best practice to use a number type

app/src/components/market/sections/market_create/steps/items/create_scalar_market.tsx
```diff
@@ -125,6 +120,62 @@ export const CreateScalarMarket = (props: Props) => {
...
+    startingPointError = 'Value must be greater than 0'
+  }
+
+  if (lowerBoundNumber !== undefined && lowerBoundNumber < '0') {
+    lowerBoundError = 'Value must be greater than 0'
```
KadenZipfel, line 147: Lower bound is ok to be 0 so the `< 0` check here is good but the error should be 'Value must be greater than or equal to 0'

app/src/components/market/sections/market_create/steps/items/create_scalar_market.tsx
```diff
@@ -125,6 +120,62 @@ export const CreateScalarMarket = (props: Props) => {
...
+  }
+
+  if (upperBoundNumber !== undefined && upperBoundNumber < '0') {
+    upperBoundError = 'Value must be greater than 0'
+  }
+  //checks if error is appropriate or not. if true, displays error and removes decimals from numbers in error message.
+  //checks for starting point to not be undefined and not equal 0. this prevents unwanted error messages like "lower bound must be less than 0."
+  if (
+    Number(startingPointNumber) > Number(upperBoundNumber) &&
```
KadenZipfel, line 155: I would convert the values to numbers right away upon assigning them as it would better reflect the variable names and work better with comparison checks

app/src/components/market/sections/market_create/steps/items/create_scalar_market.tsx
```diff
@@ -125,6 +120,75 @@ export const CreateScalarMarket = (props: Props) => {
...
+  if (lowerBoundNumber !== undefined && lowerBoundNumber < '0') {
+    lowerBoundError = 'Value must be greater than 0'
+  }
+
+  if (upperBoundNumber !== undefined && upperBoundNumber < '0') {
+    upperBoundError = 'Value must be greater than 0'
+  }
+  //checks for numbers which don't follow the input rules: lowerBound < startingPoint < upperBound. also errors for numbers equal to each other.
+  if (
+    Number(startingPointNumber) > Number(upperBoundNumber) &&
```
KadenZipfel, line 155: This could be `Number(startingPointNumber) >= Number(upperBoundNumber)` and then you can omit the else statement

app/src/components/market/sections/market_create/steps/items/create_scalar_market.tsx
```diff
@@ -125,6 +120,75 @@ export const CreateScalarMarket = (props: Props) => {
...
+    upperBoundNumber !== undefined &&
+    Number(upperBoundNumber) > 0
```
KadenZipfel, line 157: Do we need the > 0 check?

KadenZipfel (changes requested):
Pretty good so far! One thought from a UX perspective, ideally errors should display under the current input to avoid confusion, see below

Uploading Screen Recording 2021-03-17 at 9.34.38 PM.mov…

app/src/components/market/sections/market_create/steps/items/create_scalar_market.tsx
gem1no: i had it like this originally but roy told me that casting big number type to Number was dangerous

app/src/components/market/sections/market_create/steps/items/create_scalar_market.tsx
gem1no: good call

pimato:
The warning message has 10px margin-top but it looks like it is not working correctly:
Bildschirmfoto 2021-03-22 um 10 28 40

Should:
Bildschirmfoto 2021-03-22 um 10 30 13

pimato (changes requested):
see comments

app/src/components/common/form/form_row/index.tsx
```diff
@@ -24,6 +24,7 @@ const TitleWrapper = styled.div`
 
 const Error = styled(FormError)`
   margin-top: 10px;
+  margin-bottom: 900px;
```
pimato, line 27: why do we need a margin-bottom of 900px here?

pimato (changes requested):

app/src/components/market/sections/market_create/steps/items/create_scalar_market.tsx
```diff
@@ -125,6 +120,52 @@ export const CreateScalarMarket = (props: Props) => {
...
+    startingPointNumber = formatBigNumber(startingPoint, 18, 2)
+  }
+
+  if (lowerBound !== null) {
+    lowerBoundNumber = formatBigNumber(lowerBound, 18, 2)
+  }
+
+  if (upperBound !== null) {
+    upperBoundNumber = formatBigNumber(upperBound, 18, 2)
+  }
```
KadenZipfel, line 137: Since we now have access to the `BigNumber` return value of the `NumericalInput` component, we're actually better off to keep all the values as `BigNumber`s, so we can ditch all this

app/src/components/market/sections/market_create/steps/items/create_scalar_market.tsx
```diff
@@ -125,6 +120,52 @@ export const CreateScalarMarket = (props: Props) => {
...
+  const [lowerBoundError, setLowerBoundError] = useState('')
+  const [startingPointError, setStartingPointError] = useState('')
+  const [upperBoundError, setUpperBoundError] = useState('')
+
+  const handleLowerBoundError = (value: BigNumberInputReturn) => {
+    Number(value.formattedValue) <= 0
+      ? setLowerBoundError('Value must be greater than 0')
+      : Number(value.formattedValue) > Number(upperBoundNumber) && Number(upperBoundNumber) !== 0
+      ? setLowerBoundError(`Value must be less than ${upperBoundNumber}`)
+      : setLowerBoundError('')
```
KadenZipfel, line 148: Now if we're just doing `BigNumber` comparisons here, we can just do e.g. `value.lt(Zero) ? setLowerBoundError`.

For reference: https://docs.ethers.io/v5/api/utils/bignumber/

app/src/components/market/sections/market_create/steps/items/create_scalar_market.tsx
```diff
@@ -125,6 +120,52 @@ export const CreateScalarMarket = (props: Props) => {
...
+    upperBoundNumber = formatBigNumber(upperBound, 18, 2)
+  }
+
+  const [lowerBoundError, setLowerBoundError] = useState('')
+  const [startingPointError, setStartingPointError] = useState('')
+  const [upperBoundError, setUpperBoundError] = useState('')
+
+  const handleLowerBoundError = (value: BigNumberInputReturn) => {
+    Number(value.formattedValue) <= 0
+      ? setLowerBoundError('Value must be greater than 0')
```
KadenZipfel, line 145: Lower bound can be equal to 0

app/src/components/market/sections/market_create/steps/items/create_scalar_market.tsx
```diff
@@ -125,6 +120,52 @@ export const CreateScalarMarket = (props: Props) => {
...
+      : Number(value.formattedValue) > Number(upperBoundNumber) && Number(upperBoundNumber) !== 0
+      ? setLowerBoundError(`Value must be less than ${upperBoundNumber}`)
+      : setLowerBoundError('')
+  }
+
+  const handleStartingPointError = (value: BigNumberInputReturn) => {
+    Number(value.formattedValue) <= 0
+      ? setStartingPointError('Value must be greater than 0')
```
KadenZipfel, line 153: For `startingPoint` and `upperBound`, we can remove the > 0 check because they have to be greater than `lowerBound` anyway

KadenZipfel (dismissed):
Architecture looking great now, just want to reduce complexity as much as possible now

app/src/components/market/sections/market_create/steps/items/create_scalar_market.tsx
gem1no: `value` here is of type `BigNumberInputReturn` which `lt()` method doesn't exist on. I think this is the ideal way of comparing values?

gem1no: those docs suggest you can typecast  the value using `BigNumber.from()` but it only gave me errors? i tried the syntax they use as well `ethers.BigNumber.from(n)` is it possible?

KadenZipfel: > `value` here is of type `BigNumberInputReturn` which `lt()` method doesn't exist on. I think this is the ideal way of comparing values?

Yeah, should actually be `value.value` here to get the actual `BigNumber` value

> those docs suggest you can typecast the value using `BigNumber.from()` but it only gave me errors? i tried the syntax they use as well `ethers.BigNumber.from(n)` is it possible?

My bad, I shared the v5 docs when we're actually on v4: https://docs.ethers.io/v4/api-utils.html#big-numbers

pimato (changes requested):
Looks like I can´t check it out:

```git checkout 1444
fatal: Cannot switch branch to a non-commit '1444'```

**Comment by KadenZipfel** [comment]
> Looks like I can´t check it out:
> 
> ```
> fatal: Cannot switch branch to a non-commit '1444'```
> ```

I think that's because git thinks it's a commit hash. Work around would be to do:

```
git fetch origin pull/1713/head:pr-1713
git checkout pr-1713
```

File: app/src/components/market/sections/market_create/steps/items/create_scalar_market.tsx [hunk 16]
```diff
@@ -25,16 +26,15 @@ const Row = styled.div`
...
   justify-content: space-between;
-
   & > * {
     width: calc(50% - 10px);
     margin-bottom: 0;
   }
 `
-
-const NumericalInput = styled(BigNumberInput)`
+//dynamically change props
```
Mi-Lan, line 34: This comment can be removed

Mi-Lan (commented):
Hey men great progress so far! We are nearly there!

Mi-Lan:
So because of the way you structured it to evaluate error message on input once i change input the other input is showing incorrect error message since its input didn't update.
Screenshot 2021-04-07 at 15 15 16
What i would suggest is to tie them all up in one useEffect and include all three input values as trigger for it (so use [input1,input2,inpiut3] at the end of useEffect so it triggers). And have it evaluate every single value on each input change. I think that will solve most of problems.

pimato (approved):
@gem1no great work so far. I tested it out and it works!

app/src/components/market/sections/market_create/steps/items/create_scalar_market.tsx
```diff
@@ -125,6 +120,30 @@ export const CreateScalarMarket = (props: Props) => {
...
   } = props
 
+  const [lowerBoundError, setLowerBoundError] = useState('')
+  const [startingPointError, setStartingPointError] = useState('')
+  const [upperBoundError, setUpperBoundError] = useState('')
+
+  useEffect(() => {
+    Number(lowerBound) < 0
+      ? setLowerBoundError('Value cannot be negative')
+      : upperBound && Number(upperBound) !== 0 && Number(lowerBound) > Number(upperBound)
```
KadenZipfel, line 130: It works to convert to `Number`s then do the arithmetic, but in general we want to avoid type conversions wherever practicable. So for example, on this line we should do as follows:
`upperBound && !upperBound.eq(0) && lowerBound.gt(upperBound)`.

It's super nitpicky, and not completely necessary but it's best practice and super useful to get used to using the `BigNumber` type. 

For reference: https://docs.ethers.io/v4/api-utils.html#big-numbers

app/src/components/market/sections/market_create/steps/items/create_scalar_market.tsx
```diff
@@ -139,28 +158,42 @@ export const CreateScalarMarket = (props: Props) => {
...
             formField={
               <NumericalInput
                 decimals={STANDARD_DECIMALS}
+                error={upperBoundError}
+                min={0}
                 name="upperBound"
-                onChange={handleChange}
+                onChange={value => {
+                  handleChange(value)
+                  // handleUpperBoundError(value2)
```
KadenZipfel, line 195: Should remove this comment

app/src/components/market/sections/market_create/steps/items/create_scalar_market.tsx
```diff
@@ -171,11 +204,17 @@ export const CreateScalarMarket = (props: Props) => {
...
             formField={
               <NumericalInput
                 decimals={STANDARD_DECIMALS}
+                error={startingPointError}
+                min={0}
                 name="startingPoint"
-                onChange={handleChange}
+                onChange={value => {
+                  handleChange(value)
+                  // handleStartingPointError(value3)
```
KadenZipfel, line 216: And this one

KadenZipfel (changes requested):
Awesome work with the `useEffect` btw

app/src/components/market/sections/market_create/steps/items/create_scalar_market.tsx
gem1no: i see i see. i'll do this from now on, thanks.

app/src/components/market/sections/market_create/steps/items/create_scalar_market.tsx
```diff
@@ -125,6 +120,30 @@ export const CreateScalarMarket = (props: Props) => {
     upperBound,
   } = props
 
+  const [lowerBoundError, setLowerBoundError] = useState('')
+  const [startingPointError, setStartingPointError] = useState('')
+  const [upperBoundError, setUpperBoundError] = useState('')
+
+  useEffect(() => {
+    Number(lowerBound) < 0
```
KadenZipfel, line 128: Should be `lowerBound.lt(Zero)` here with `Zero` being imported from `ethers/constants` (could also use `new BigNumber(0)` as it's the same value, but `Zero` is a bit cleaner)

app/src/components/market/sections/market_create/steps/items/create_scalar_market.tsx
```diff
@@ -125,6 +120,30 @@ export const CreateScalarMarket = (props: Props) => {
...
+      ? setLowerBoundError('Value cannot be negative')
+      : upperBound && !upperBound.eq(0) && lowerBound?.gt(upperBound)
```
KadenZipfel, line 130: Should also make sure we're using `Zero` here instead of `0` as with other places

KadenZipfel: My bad btw, I think I suggested using `0` oops

KadenZipfel (changes requested):

Mi-Lan:
@gem1no Seems that there's still an issue with multiple fields displaying error when entering in one field...
Screenshot 2021-04-12 at 19 47 54
Screenshot 2021-04-12 at 19 49 07

app/src/components/market/sections/market_create/steps/items/create_scalar_market.tsx
gem1no: oops, good catch. missed that one on line 128. 

added these changes

gem1no:
> @gem1no Seems that there's still an issue with multiple fields displaying error when entering in one field...
> Screenshot 2021-04-12 at 19 47 54
> Screenshot 2021-04-12 at 19 49 07

this is fixed now

pimato (changes requested):
I think the new fix is awesome but I found a bug:
Bildschirmaufnahme 2021-04-13 um 11 38 33

Mi-Lan (changes requested):
Nice work @gem1no this is looking great! Just a little more needed. Seems that when you click on it after that you can enter whatever value and it will not display error. Similar to geronimos comment...

https://user-images.githubusercontent.com/33226956/114603606-3b5f5e80-9c98-11eb-8468-6c5658938ed2.mov

Mi-Lan (changes requested):
@gem1no Hmm seems that it was working better in previous iteration now it acts weird
Screenshot 2021-04-16 at 11 32 29

https://user-images.githubusercontent.com/33226956/115004877-92874e00-9ea7-11eb-9d02-00b677c9bbf2.mov
- I would also suggest testing it thoroughly in the future before pushing and requesting review you could have spotted this
- Take your time by all means

pimato:
Why are error messages disappearing on removing focus?
Bildschirmaufnahme 2021-04-28 um 10 44 35

pimato:
looks like clicking on the Lower Bound will remove the error message:
Bildschirmaufnahme 2021-04-28 um 17 49 33

pimato:
Also, if you set the same starting point to the upper bound no error is shown:
Bildschirmfoto 2021-04-28 um 17 55 45

pimato (changes requested):
see comments

pimato (changes requested):
I think there are a few things we have missed:

Bildschirmfoto 2021-05-05 um 15 51 01

Starting Point cannot be Lower Bound

Bildschirmfoto 2021-05-05 um 15 52 15

Lower Bound cannot be Upper Bound

pimato:
Even worse:

Lower Bound cannot be Upper Bound cannot be Starting Point:
Bildschirmfoto 2021-05-05 um 15 53 27

app/src/components/common/form/form_error/index.tsx
```diff
@@ -2,12 +2,15 @@ import React, { HTMLAttributes } from 'react'
...
-  margin: 0 0 10px;
-  text-align: left;
+  color: ${props => props.theme.colors.alert};
+  font-family: Roboto;
+  font-style: normal;
+  font-weight: normal;
+  line-height: 1.5px;
+  align-items: center;
+  letter-spacing: 0.2px;
+  margin-top: 15px;
```
pimato, line 12: margin-top: 10px;

pimato (changes requested):

pimato (approved):
looks good now!

Mi-Lan (approved):
Found some minor issues...but considering most people won't play around with it like we do for testing.... it should do its job just about right!

Issue #1722, opened 2021-03-10
Buy button disabled if less than 0.01 collateral amount


KadenZipfel:
![image](https://user-images.githubusercontent.com/30579067/110680880-01c5a000-818e-11eb-868e-528fcd521193.png)

I believe the issue is in the `isBuyDisabled` check, where it's checking the value of a formatted number limited to 2 decimal places instead of checking the full amount as a BigNumber. Likely the same or similar issue for selling and other actions, so should be checked for.

PR #1725, opened 2021-03-11
Fix disabled buy/sell buttons

KadenZipfel:
Closes: https://github.com/protofire/omen-exchange/issues/1722


hexyls (approved):

Mi-Lan (changes requested):
Should we maybe adjust rest of UI? This doesn't look good...
Screenshot 2021-03-11 at 12 38 48
Screenshot 2021-03-11 at 12 37 51

KadenZipfel:
> Should we maybe adjust rest of UI? This doesn't look good with zeros everywhere
> Screenshot 2021-03-11 at 12 38 48
> Screenshot 2021-03-11 at 12 37 51

Yeah good call, tough though because we don't wanna show a ton of decimals. Maybe we could display '<0.01'? @pimato any thoughts?

pimato:
> > Should we maybe adjust rest of UI? This doesn't look good with zeros everywhere
> > Screenshot 2021-03-11 at 12 38 48
> > Screenshot 2021-03-11 at 12 37 51
> 
> Yeah good call, tough though because we don't wanna show a ton of decimals. Maybe we could display '<0.01'? @pimato any thoughts?

yeah I think that would be the easiest solution right now. I think we will fix this issue once we have the new token curation contract where we store the decimals on-chain.

pimato (approved):
Fixes the underlying issue.

KadenZipfel:
I realized this '<0.01' issue is pretty time consuming as it has to be done in several files, so I exported to this issue: https://github.com/protofire/omen-exchange/issues/1738

Mi-Lan (approved):

Issue #1724, opened 2021-03-10
Scalar market list item displays current prediction as undefined if no liquidity


KadenZipfel:
![image](https://user-images.githubusercontent.com/30579067/110705246-f896fc00-81aa-11eb-9925-8969090ab91a.png)

PR #1726, opened 2021-03-11
Fix undefined list item current prediction

KadenZipfel:
Closes: https://github.com/protofire/omen-exchange/issues/1724

Maybe there's something better to put than 'Unknown'?


hexyls (approved):
hmmm couldn't think of anything better than Unknown, so looks good to me

pimato:
@KadenZipfel there is no way to check what the prediction was right before the liquidity was removed?

pRoy24 (approved):
LGTM

KadenZipfel:
> @KadenZipfel there is no way to check what the prediction was right before the liquidity was removed?

I think it could be done with a subgraph update but I think it might be confusing for users since re-adding liquidity would just set the probability to 50%. What do you think?

pimato (approved):

PR #1746, opened 2021-03-18
Fixed small position <0.01 displaying wrong value

### LINKED ISSUE #1734: Very small positions <0.01 cause position table to display 100% loss
Likely a rounding issue

image

Mi-Lan:
closes #1734 
- I just basically changed value used for calculation to have more decimals but it still uses same decimals in displaying...
- So now they can have 18 zeros and still won't break it !


gem1no (approved):

pimato (approved):
looks fixed:
Bildschirmfoto 2021-03-18 um 19 02 19

hexyls (approved):

KadenZipfel (approved):

Issue #1742, opened 2021-03-17
Winnings not being returned on market where user owns shares of winning outcome


KadenZipfel:
A user has reported that there is no acknowledgement of their winning shares on the front-end for [this market](https://xdai.omen.eth.link/#/0x17aad168c9dac86af6b05970cf131d67bda522dc). This is similar to an issue that was solved by: https://github.com/protofire/omen-exchange/pull/1702, but the user still reports the issue on the patched version.

From what I've been able to test so far, it seems as though the users shares must not be present in `marketMakerData.balances`.

DM me for the transaction if working on this

PR #1748, opened 2021-03-18
Fix failing getScalarBalances when no liquidity in closed market

KadenZipfel:
Closes: https://github.com/protofire/omen-exchange/issues/1742

The problem: We were checking the amount of outcomes in a scalar market based on a data point that becomes null if liquidity is withdrawn from the market. Instead, we should assert that scalar markets always have two outcomes.

To test: Check if redeem winnings functionality exists in scalar markets with and without liquidity

### DIFF
--- a/app/src/hooks/useBlockchainMarketMakerData.tsx
+++ b/app/src/hooks/useBlockchainMarketMakerData.tsx
@@ -46,15 +46,15 @@ const getBalances = (
 }
 
 const getScalarBalances = (
-  outcomePrices: string[],
+  outcomeAmounts: string[],
   marketMakerShares: BigNumber[],
   userShares: BigNumber[],
   payouts: Maybe<Big[]>,
 ): BalanceItem[] => {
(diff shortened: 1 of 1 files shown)


Mi-Lan (approved):
Tested on both empty and non empty markets and it works! GJ!
Empty market:
Master:
Screenshot 2021-03-18 at 19 48 32
This pr:
Screenshot 2021-03-18 at 19 48 28
Non empty market:
- Master 
Screenshot 2021-03-18 at 19 51 04
- this pr
Screenshot 2021-03-18 at 19 52 05

Mi-Lan:
Screenshot 2021-03-18 at 19 55 58
Only issue i found is that it sometimes just looks like this without any message what was won. This was for market where i had short position and was empty... And resolved to outcome 80 where i chose 22...

hexyls (approved):
👍

Screen Shot 2021-03-19 at 3 38 19 pm

pimato (approved):

Issue #1745, opened 2021-03-18
Reality link not working while on xDai relayer


Mi-Lan:
Steps to reproduce:
1)Make sure to be on branch where relayer pr is merged
2)Switch to xDai on the relayer button
3)Go to any market
4)Click on the reality.eth on additional market data 
5)Leads nowhere!

https://user-images.githubusercontent.com/33226956/111623052-2eab3000-87ea-11eb-8f0b-77d158a5ad6f.mov

Also for scalar markets on answer reality question as well! Heres where it can be seen http://localhost:3000/#/0x071bc23726c6363e2c017185ff431c4d2d9ddf8d

![Screenshot 2021-03-18 at 13 08 22](https://user-images.githubusercontent.com/33226956/111623737-0f60d280-87eb-11eb-9ec6-d9293308a90e.png)

KadenZipfel:
So the problem here is that we can incorporate a check if the relay is being used in `useRealityLink`, which will send it to the xDai reality URL, but we can't actually view the reality item unless we switch our network to xDai. I've incorporated this partial fix here: https://github.com/protofire/omen-exchange/pull/1758

PR #1758, opened 2021-03-19
Add relay check to useRealityLink

KadenZipfel:
Partially closes: https://github.com/protofire/omen-exchange/issues/1745

The problem with this solution is although we now have the correct reality link, we still have to switch our network to xDai to view the reality item.


hexyls (approved):
👍

Yeah it is annoying that users have to switch over to the xdai network once they are outside Omen. It looks like they even have to refresh the page after switching networks (not a lot we can do about that).

pimato (approved):
tested, works!

Mi-Lan (approved):
Better then before! But still sucks to have to switch network...Would be cool if they added feature to check markets out chain agnostic @pimato could they make this happen?

pimato:
> Better then before! But still sucks to have to switch network...Would be cool if they added feature to check markets out chain agnostic @pimato could they make this happen?

I think reality.eth needs to include the connect network feature by metamask at some point. So people can connect easier to xDai.

Issue #1757, opened 2021-03-19
Max sell failing on xDai


KadenZipfel:
It looks like there is a rounding issue when selling on xDai, whether through the relayer or not. Selling a little less than the max amount works. Same issue on scalar and categorical markets.

Possibly related to: https://github.com/protofire/omen-exchange/issues/1739 but I haven't been able to reproduce a failing redeem.

hexyls:
related: https://github.com/protofire/omen-exchange/issues/1676

Issue #1676, opened 2021-02-23
underflow when selling outcome tokens


hexyls:
not really sure about this one yet, but kinda related to https://github.com/protofire/omen-exchange/issues/1540 and https://github.com/protofire/omen-exchange/issues/1593

could be the inputs we pass from the frontend are wrong or the smart contract has a bug, but sometimes [calcSellAmount](https://github.com/gnosis/conditional-tokens-market-makers/blob/master/contracts/FixedProductMarketMaker.sol#L284) will incorrectly calculate `outcomeTokenSellAmount`. 

[example tx](https://dashboard.tenderly.co/tx/xdai/0xace4c12d7ec3d448fc82cdbbf27fd7817ce7c46bc56071e0bab7ebd4aff58a23/debugger) where `calcSellAmount` estimates `outcomeTokenSellAmount` to be `66304306214068927` when it should be `66304306214068926` (off by 1 gwei), which causes an underflow and prevents selling. This usually will occur when a user is trying to sell max amount of outcome tokens.

Issue #1739, opened 2021-03-16
incorrect earnedCollateral amount when redeeming


hexyls:
`earnedCollateral` can be too high when [transferring from the cpk to the user](https://github.com/protofire/omen-exchange/blob/master/app/src/services/cpk.ts#L1142), causing transactions to fail. 

example [tx](https://dashboard.tenderly.co/tx/xdai/0x8a6ed0d6877a774012de6a135ea196f1a171c7072ee2c47a424010235d3e40b3) where the amount redeemed from the conditional tokens contract is 4327888365033786**67371** but we try to transfer 4327888365033786**87044** from the cpk to the user.

pimato:
https://github.com/protofire/omen-exchange/pull/1786 potentially fixed this, closing for now.

PR #1761, opened 2021-03-22
fix max sell errors

hexyls:
closes https://github.com/protofire/omen-exchange/issues/1757 https://github.com/protofire/omen-exchange/issues/1676

There were clues in the code comments for how to resolve this, I tried to see if there was a better way but ultimately just kept the original solution mentioned.


KadenZipfel (approved):
🔥

pimato (approved):
works!

pimato:
lets connect the rounding fix with this issue: https://github.com/protofire/omen-exchange/issues/1739
redeem not working! Should round down the amount to redeem too.

pRoy24:
I think this may cause precision errors if the result has more decimals also would not work if precision error is more than a factor of 99999/100000 
Please checkout this alternative solution I propose and let me know if this works
https://github.com/protofire/omen-exchange/pull/1767

hexyls:
closing in favour of https://github.com/protofire/omen-exchange/pull/1761

PR #1767, opened 2021-03-22
Add fix for max sell and redeem by rounding down values

pRoy24:
Round down to 14 precision points .
This prevents overflow and ensures that there are no precision issues while selling.

Fixes https://github.com/protofire/omen-exchange/issues/1676
https://github.com/protofire/omen-exchange/issues/1757

PR #1769, opened 2021-03-23
move relay fee, default to unwrapping for relay

hexyls:
PR moves/hides the relay fee as part of transaction amounts, e.g. buy amount, sell amount where possible. So if you deposit 100 dai you can spend 100 dai rather than 99.999 dai. 

Also defaults sells/funding removals as unwraps for the relay (can't pay the relay fee in wxdai).

### DIFF
--- a/app/src/hooks/useCollateralBalance.tsx
+++ b/app/src/hooks/useCollateralBalance.tsx
@@ -1,7 +1,6 @@
 import { BigNumber } from 'ethers/utils'
 import { useEffect, useState } from 'react'
 
-import { RELAY_FEE } from '../common/constants'
 import { ERC20Service } from '../services'
 import { pseudoNativeAssetAddress } from '../util/networks'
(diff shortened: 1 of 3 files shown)


KadenZipfel (approved):
Great work! Had an issue testing max sell initially but worked as expected once I slotted in the code from https://github.com/protofire/omen-exchange/pull/1767/files

Mi-Lan (approved):
Nicely done man! 
Before:
Screenshot 2021-03-24 at 12 10 45
After:
Screenshot 2021-03-24 at 12 11 44

Mi-Lan:
Seems that there is small issue...when i do max sell right after i can still sell some dust but when i refresh page its unavailable...

https://user-images.githubusercontent.com/33226956/112304045-4fb5ca00-8c9d-11eb-9d4a-d0a413ba61d4.mov

pimato (approved):

PR #1770, opened 2021-03-23
Batch claim transaction

### LINKED ISSUE #1701: Batch multiple claims into one user actions
Currently, if a user transfers multiple xDai to Ethereum Mainnet the UI is only showing the latest transfer to be claimed. We should batch all claims into one transaction powered by the Gnosis Safe proxy kit.

Mi-Lan:
closes #1701 
- Note i haven't tested it myself because I wanted first to get feedback from someone before trying since it would be very expensive learning curve...


hexyls:
Nice work man! In general the approach in this PR should have worked just fine, however because the CPK is initialized on the xdai network you cannot claim through mainnet using CPK.

We could have initialized a second instance of the CPK to claim, however I think this is a good opportunity to make it as cheap as possible for relay users to claim their DAI. By skipping the CPK we can save users a lot on contract deployment costs. 

I have deployed a simple multi claim contract at [0xe0a392a61357f86f0da2d1586b91ca54e9097022](https://etherscan.io/address/0xe0a392a61357f86f0da2d1586b91ca54e9097022) and have updated this PR to use it. Here is an [example multi claim transaction](https://etherscan.io/tx/0xb4557bd961f3b0e94c71e569a912ef9d121437dd4e61b7d2741becb219471e17).

One problem I found though is that the [subgraph we are using](https://thegraph.com/explorer/subgraph/maxaleks/foreign-bridge-mainnet) does not appear to pick up multiple claims executions in a single transaction. I have [added a work around](https://github.com/protofire/omen-exchange/pull/1770/files#diff-68819a0aef2491738d5950d92e467e2eb3918455cadd15251a1146e8a564c0d7R234) that checks claim state directly on chain, but it would be great if you could reach out to the subgraph owners, share [this tx](https://etherscan.io/tx/0xb4557bd961f3b0e94c71e569a912ef9d121437dd4e61b7d2741becb219471e17), and see if they can update their subgraph 🙌

Mi-Lan:
> Nice work man! In general the approach in this PR should have worked just fine, however because the CPK is initialized on the xdai network you cannot claim through mainnet using CPK.
> 
> We could have initialized a second instance of the CPK to claim, however I think this is a good opportunity to make it as cheap as possible for relay users to claim their DAI. By skipping the CPK we can save users a lot on contract deployment costs.
> 
> I have deployed a simple multi claim contract at [0xe0a392a61357f86f0da2d1586b91ca54e9097022](https://etherscan.io/address/0xe0a392a61357f86f0da2d1586b91ca54e9097022) and have updated this PR to use it. Here is an [example multi claim transaction](https://etherscan.io/tx/0xb4557bd961f3b0e94c71e569a912ef9d121437dd4e61b7d2741becb219471e17).
> 
> One problem I found though is that the [subgraph we are using](https://thegraph.com/explorer/subgraph/maxaleks/foreign-bridge-mainnet) does not appear to pick up multiple claims executions in a single transaction. I have [added a work around](https://github.com/protofire/omen-exchange/pull/1770/files#diff-68819a0aef2491738d5950d92e467e2eb3918455cadd15251a1146e8a564c0d7R234) that checks claim state directly on chain, but it would be great if you could reach out to the subgraph owners, share [this tx](https://etherscan.io/tx/0xb4557bd961f3b0e94c71e569a912ef9d121437dd4e61b7d2741becb219471e17), and see if they can update their subgraph 🙌

Thank you so much for the detailed feedback! Means a lot 🙌  Yeah I overlooked fact that cpk is on xDai. I can ping xDai folks since its their subgraph although dunno how big of change it will require dunno if they are happy in first place that we are using their subgraph...

hexyls (approved):

pimato (changes requested):
@hexyls I was able to claim a batch but it looks like there was an error:
Bildschirmfoto 2021-03-25 um 21 21 44

Mi-Lan:
@hexyls GJ man you fixed it up nicely! here is tx hash of transaction https://etherscan.io/tx/0x0d646fba3f924f9205727085ef899c2819a1e1fe06d65f030b40a49ef344b6ae
Screenshot 2021-03-26 at 00 42 55

app/src/services/xdai.ts
```diff
@@ -104,19 +134,20 @@ class XdaiService {
     }
   }
 
+  claim = async (messages: string[], signatures: string[]) => {
+    const MULTI_CLAIM_ADDRESS = '0xE0A392a61357F86f0da2d1586b91cA54e9097022'
```
KadenZipfel, line 138: Would be best practice to move this to `constants.ts`

KadenZipfel (approved):
One small suggestion, other than that LGTM!

pimato (approved):

PR #1772, opened 2021-03-24
display <0.01 for values less than 0.01 to handle visual bug

### LINKED ISSUE #1738: Display <0.01 for small amounts
When inputting less than 0.01 for a transaction amount, the UI will display '0.00' in several places. Instead, it should display '<0.01'. Needs to be done in: buy, sell and pool views for both scalar and categorical markets.

Examples:
image
image

gem1no:
fixes #1738

some example: (old tx modal, hadnt pulled at this point yet)

(buying shares)

(successful buy)

(deposit)

(successful deposit)

- [x] categorial buy: yes + yes amounts <0.01
- [x] categorial sell: yes + yes for amounts <0.01
- [x] categorial pool deposit: yes and yes for amounts <0.01
- [x] categorial pool withdrawal: yes and yes for amounts < 0.01
- [x] scalar buy: yes and yes for amounts <0.01
- [x] scalar sell: yes and yes for amounts <0.01
- [x] scalar pool deposit: yes and yes for amounts <0.01 
- [x] scalar pool withdrawal: yes and yes for amounts <0.01


hexyls (commented):
I'm thinking maybe we should only show <0.01 when the number is less than 0.01 but greater than 0. If a value is actually 0 maybe we should show 0, thoughts?

Mi-Lan:
> I'm thinking maybe we should only show <0.01 when the number is less than 0.01 but greater than 0. If a value is actually 0 maybe we should show 0, thoughts?

I agree as well! I think I also mentioned that on previous pr...

Mi-Lan:
Screenshot 2021-03-24 at 14 11 10
I think minus sign in the max los should be after less then sign...

app/.vscode/settings.json
```diff
@@ -0,0 +1,2 @@
+{
+}
```
Mi-Lan, line 2: We should delete this or put it in gitignore

app/src/util/tools.ts
```diff
@@ -512,13 +512,104 @@ export const formatNumber = (number: string, decimals = 2): string => {
...
+
+  if (Number(sharesAmount) < 0.01 && depositOrWithdraw === 'hasWithdrawn') {
+    message(`Successfully withdrew <${0.01} ${displayCollateral.symbol}`)
+  }
+
+  if (Number(sharesAmount) > 0.01 && depositOrWithdraw === 'hasWithdrawn') {
+    message(`Successfully withdrew ${sharesAmount} ${displayCollateral.symbol}`)
+  }
+}
+
```
Mi-Lan, line 612: Shit man seems that you misunderstood my feedback or that i didn't communicate it quite well....So i didn't mean to put every message in the tools file. But to just take logic into one function like this:
`function smallAmoutChecker(value){
 if(value<0.01)return '<0.01' 
else return value
}
`
And then for example instead of 
setMessage(`Successfully sold ${sharesAmount} '${balances[outcomeIndex].outcomeName}' shares.`)
you would do
setMessage(`Successfully sold ${smallAmountChecker(sharesAmount)} '${balances[outcomeIndex].outcomeName}' shares.`)
Does that make sense?

gem1no: yes, sorry. misunderstanding. i thought you wanted all of the logic inside tools to reduce noise in other files...

gem1no:
> Screenshot 2021-03-24 at 14 11 10
> I think minus sign in the max los should be after less then sign...

doesn't this change the meaning of what's being displayed? putting the minus sign to the right of the less than sign like "Max loss: <-0.01", means, total possible loss = "less than negative 0.01" when it's not possible to have a negative loss?  while "-<0.008" means, omen subtracts less than 0.008 from your wallet if your prediction is wrong. please correct me if wrong! i appreciate your attention to detail

gem1no:
> I'm thinking maybe we should only show <0.01 when the number is less than 0.01 but greater than 0. If a value is actually 0 maybe we should show 0, thoughts?

Done!
thanks for helping me figure out that issue with precision
thank you @Mi-Lan for all your great suggestions here

gem1no:
> Screenshot 2021-03-24 at 14 11 10
> I think minus sign in the max los should be after less then sign...

Fixed in latest commit

hexyls (approved):
Great work @gem1no this looks good to me 🙌

Marking as approve, however just make sure you delete the empty `app/.vscode/settings.json` file and add it to .gitignore before you merge.

gem1no:
> Great work @gem1no this looks good to me 🙌
> 
> Marking as approve, however just make sure you delete the empty `app/.vscode/settings.json` file and add it to .gitignore before you merge.

damn idk how it got in there again! will do

app/src/components/market/sections/market_buy/market_buy.tsx
```diff
@@ -229,7 +230,7 @@ const MarketBuyWrapper: React.FC<Props> = (props: Props) => {
       const sharesAmount = formatBigNumber(displayTradedShares, baseCollateral.decimals)
       setTweet('')
       setStatus(Status.Loading)
-      setMessage(`Buying ${sharesAmount} shares...`)
+      setMessage(`Buying ${handleSmallShares(sharesAmount)} shares...`)
```
KadenZipfel, line 233: I think it would actually be better if we just use `formatNumber` in place of `handleSmallShares` since it already does what `handleSmallShares` does and also provides a consistent system for displaying numbers

KadenZipfel (changes requested):
Well done, just suggested some optimizations

app/src/components/market/common/position_selection_box/index.tsx
```diff
@@ -78,7 +79,7 @@ export const PositionSelectionBox = (props: Props) => {
           <PositionSelectionTitle>{balance.outcomeName}</PositionSelectionTitle>
         </PositionSelectionLeft>
         <PositionSelectionAmount>
-          {formatNumber(formatBigNumber(balance.shares, decimals))} Shares
+          {formatNumber(formatBigNumber(balance.shares, decimals, STANDARD_DECIMALS))} Shares
```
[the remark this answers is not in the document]
gem1no: without this `STANDARD_DECIMALS` as the third argument for some places the numbers don't display properly. 

for example on the `shortSharesFormatted` and `longSharesFormatted` variables 
inside of app/src/components/market/common/position_table/index.tsx

```const shortSharesFormatted = formatNumber(
    formatBigNumber(shortShares || new BigNumber(0), collateral.decimals, STANDARD_DECIMALS),
  )
  const longSharesFormatted = formatNumber(
    formatBigNumber(longShares || new BigNumber(0), collateral.decimals, STANDARD_DECIMALS),
  )```
  
 without the standard decimals here the short/long shares held  won't display properly for values under 0.01. i will try and find where it isn't necessary, though. the one you commented here was good. deleted it and seems like everything is still fine

File: app/src/components/market/sections/market_buy/market_buy.tsx (hunk 3 above)
Comment (gem1no) [reply app/src/components/market/sections/market_buy/market_buy.tsx:233]: when i try to wrap it in `formatNumber` it doesn't properly display number below 0.01 🤔

Comment (gem1no) [reply app/src/components/market/sections/market_buy/market_buy.tsx:233]: i messed with the logic of the `formatNumber` function too, literally copy pasted in what `handleSmallShares` was doing and it still wouldn't work properly.

Comment (KadenZipfel) [reply app/src/components/market/sections/market_buy/market_buy.tsx:233]: Seems the issue is that `formatBigNumber` is using the default of 2 for its precision parameter, so the value being used in `formatNumber` is just `0.00`

File: app/src/components/market/common/position_selection_box/index.tsx (hunk 4 above)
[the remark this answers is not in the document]
Comment (KadenZipfel) [reply app/src/components/market/common/position_selection_box/index.tsx]: Oh, yeah you should still include the third parameter, but it should be `decimals`, so it should be `formatBigNumber(balance.shares, decimals, decimals)`

File: app/src/components/market/common/position_table/index.tsx [hunk 5]
```diff
@@ -58,8 +59,12 @@ export const PositionTable = (props: Props) => {
 
   const shortShares = balances[0].shares
   const longShares = balances[1].shares
-  const shortSharesFormatted = formatNumber(formatBigNumber(shortShares || new BigNumber(0), collateral.decimals))
-  const longSharesFormatted = formatNumber(formatBigNumber(longShares || new BigNumber(0), collateral.decimals))
+  const shortSharesFormatted = formatNumber(
+    formatBigNumber(shortShares || new BigNumber(0), collateral.decimals, STANDARD_DECIMALS),
```
KadenZipfel, line 63: Have to make sure all the `formatBigNumber` instances are not using `STANDARD_DECIMALS` as the third param if the second param is a variable value like collateral.decimals

app/src/components/market/sections/market_buy/scalar_market_buy.tsx
```diff
@@ -121,7 +121,7 @@ export const ScalarMarketBuy = (props: Props) => {
     context,
   )
   const collateralBalance = maybeCollateralBalance || Zero
-  const walletBalance = formatNumber(formatBigNumber(collateralBalance, collateral.decimals, 5), 5)
+  const walletBalance = formatNumber(formatBigNumber(collateralBalance, collateral.decimals), 5)
```
KadenZipfel, line 124: Shouldn't be removing the third param from `formatBigNumber` here as it will return an improper value for `walletBalance`. Looks like this is done in other places too

KadenZipfel (changes requested):

Mi-Lan (changes requested):
Hey men good progress so far! 
Screenshot 2021-03-31 at 20 44 44
So for positive value it shouldn't be 0.00 but <0.01. Guess its good solution -0.00 for negative value i can't come up with anything better either.

Mi-Lan (changes requested):
Also in payout view it's not less then...
Screenshot 2021-03-31 at 20 53 05

app/src/components/market/common/position_table/index.tsx
gem1no: got it now, thanks, appreciate your patience w me

gem1no:
> Hey men good progress so far!
> Screenshot 2021-03-31 at 20 44 44
> So for positive value it shouldn't be 0.00 but <0.01. Guess its good solution -0.00 for negative value i can't come up with anything better either.

Thanks Violet, I appreciate this. I went back to optimize and in the process broke some other value displays. Good catch, should be fixed now.

app/src/components/market/sections/market_buy/scalar_market_buy.tsx
```diff
@@ -207,12 +207,18 @@ export const ScalarMarketBuy = (props: Props) => {
...
-  const baseCostFormatted = `${formatNumber(formatBigNumber(baseCost, collateral.decimals))} ${collateral.symbol}`
+  const feeFormatted = `${formatNumber(formatBigNumber(feePaid.mul(-1), collateral.decimals, collateral.decimals))} ${
+    collateral.symbol
+  }`
+  const baseCostFormatted = `${formatNumber(formatBigNumber(baseCost, collateral.decimals, collateral.decimals))} ${
+    collateral.symbol
+  }`
   const potentialProfitFormatted = potentialProfit.gt(Zero)
-    ? `${formatNumber(Number(formatBigNumber(potentialProfit, collateral.decimals)).toString())} ${collateral.symbol}`
+    ? `${formatNumber(Number(formatBigNumber(potentialProfit, collateral.decimals, collateral.decimals)).toString())} ${
```
KadenZipfel, line 217: I think this is from a past PR, but the `Number().toString()` part is redundant. Instead it should just be: 
`formatNumber(formatBigNumber(potentialProfit, collateral.decimals, collateral.decimals)))`

app/src/components/market/sections/market_buy/scalar_market_buy.tsx
```diff
@@ -371,7 +377,7 @@ export const ScalarMarketBuy = (props: Props) => {
             <TransactionDetailsLine />
             <TransactionDetailsRow
               title={'Max. Loss'}
-              value={`${!amount.isZero() ? '-' : ''}${formatNumber(formatBigNumber(amount, collateral.decimals))} ${
+              value={`${formatNumber(formatBigNumber(amount, collateral.decimals, collateral.decimals))} ${
```
KadenZipfel, line 380: We should keep this `${!amount.isZero() ? '-' : ''}` part to make sure we include the negative sign here

app/src/util/tools.ts
```diff
@@ -501,13 +501,16 @@ export const limitDecimalPlaces = (value: string, decimals: number) => {
...
 export const formatNumber = (number: string, decimals = 2): string => {
-  if (number.length < 1) {
-    return `0${decimals > 0 ? '.' + '0'.repeat(decimals) : ''}`
-  }
-
   const fixedInt = parseFloat(number.split(',').join('')).toFixed(decimals)
   const splitFixedInt = fixedInt.split('.')[0]
   const formattedSubstring = splitFixedInt.replace(/(\d)(?=(\d{3})+(?!\d))/g, '$1,')
+  if (number.length < 1) {
+    return `0${decimals > 0 ? '.' + '0'.repeat(decimals) : '0'}`
```
KadenZipfel, line 508: Not sure if there's a reason for setting `: '0'` instead of `: ''` here, but if I do `formatNumber('', 0)` (the only situation this should fire), then it returns '00' which is probably not what we want

KadenZipfel (changes requested):

app/src/components/market/sections/market_buy/scalar_market_buy.tsx
gem1no: so the reason i removed this was to solve the problem that @Mi-Lan brought up.

do you want max loss to say you're losing negative 7 dollars? wouldn't that mean you're gaining 7 dollars because a 7 dollar debt is being removed from you? two negatives = positive. personally i think semantically this makes more sense and solves the above mentioned problem but if you think i should try to solve this differently i will definitely try!

> > Screenshot 2021-03-24 at 14 11 10
> > I think minus sign in the max los should be after less then sign...
> 
> Fixed in latest commit

app/src/util/tools.ts
gem1no: thanks

app/src/components/market/sections/market_buy/scalar_market_buy.tsx
KadenZipfel: I agree it's a bit confusing to have a negative max. loss. Curious @pimato's take on this

pimato: I think it is fine with having no `-` in the Max. Loss value

pimato (approved):
works good!

Issue #1773, opened 2021-03-24
V1.2.5 testing checklist


KadenZipfel:
Focus should primarily be on testing xDai tight integration. Can do various generic testing via the relay, e.g. market buy/sell.
Copy/paste this checklist into a comment below and check off what you've tested. 

- [x] Relay
    - [x] Enable
    - [x] Deposit
    - [x] Withdraw
    - [x] Generic market interactions (see above checklist)
        - [x] Transaction modal works as expected
            - [x] Correct block explorer links
            - [x] Tweet functionality
- [x] History tab transaction overview works as expected
- [x] Correct reality links on xDai
- [x] List view does not display scalar market current probability as undefined if no liquidity
- [x] Market scale works as expected if value out of bounds
    - [x] > upper bound
    - [ ] < lower bound
- [x] Buy button not disabled if 0 < amount < 0.01
- [x] Scalar markets not displaying as invalid if final outcome exactly 50%
- [ ] Redeem on xDai
    - [ ] Proper amount of winnings displayed
    - [ ] Scalar market winnings displayed before condition resolution
    - [ ] Redeeming works as expected
- [x] Max sell works as expected
- [ ] Wallet connection modal works as expected
    - [ ] Metamask
    - [ ] Walletconnect

hexyls:
- [ ] Relay
    - [x] Enable
    - [x] Deposit
    - [ ] Withdraw
    - [ ] Generic market interactions (see above checklist)
        - [ ] Transaction modal works as expected
            - [ ] Correct block explorer links
            - [ ] Tweet functionality
- [ ] History tab transaction overview works as expected
- [ ] Correct reality links on xDai
- [ ] List view does not display scalar market current probability as undefined if no liquidity
- [ ] Market scale works as expected if value out of bounds
    - [ ] > upper bound
    - [ ] < lower bound
- [ ] Buy button not disabled if 0 < amount < 0.01
- [ ] Scalar markets not displaying as invalid if final outcome exactly 50%
- [ ] Redeem on xDai
    - [ ] Proper amount of winnings displayed
    - [ ] Scalar market winnings displayed before condition resolution
    - [ ] Redeeming works as expected
- [ ] Max sell works as expected
- [ ] Wallet connection modal works as expected
    - [ ] Metamask
    - [ ] Walletconnect

pimato:
- [ ] Relay
    - [ ] Enable
    - [ ] Deposit
    - [ ] Withdraw
[...]

PR #1778, opened 2021-03-25
Fix/#1773

hexyls:
related https://github.com/protofire/omen-exchange/issues/1773

contains small fixes:
- resets tweet messages
- fixes market history graph when using relay
- fixes relay max withdraw 
- only fetches bridge allowance when on mainnet


pimato (approved):

Mi-Lan (approved):
Gj finding these ones!

KadenZipfel (approved):
🔥

app/src/services/cpk.ts
```diff
@@ -1360,11 +1360,13 @@ class CPKService {
         // get mainnet relay signer
         const to = await this.cpk.ethLibAdapter.signer.signer.getAddress()
 
+        const withdrawAmount = amount.sub(RELAY_FEE)
```
KadenZipfel, line 1363: It looks like subtracting the relay fee causes problems with the minimum withdraw amount. If a user tries to withdraw the minimum 10 DAI, it's really only withdrawing 9.999, which causes the transaction to revert.

KadenZipfel (changes requested):

app/src/services/cpk.ts
KadenZipfel: As a note, the currently discussed solution is to only charge a relay fee if the amount to withdraw is not exactly 10 xDai, but seems like the relayer infrastructure doesn't allow omitting a fee: https://github.com/hexyls/omen-relay/blob/10429f7cbe24a52b3e77e42439dea540c5fb64c2/api/index.js#L35

KadenZipfel (approved):
🚀

Mi-Lan:
Works! Heres tx https://blockscout.com/poa/xdai/tx/0xb2daa1a3c55057596c59c0b494141e46cf72f55467cb5d2d3a94f6fb0fd0dcdf

PR #1782, opened 2021-03-25
Added User address in history chart

### LINKED ISSUE #1688: History table is showing cpk(gnosis safe) address instead of users address
- Currently we are not getting users address but proxy address because of the JazzIcons which uses address to display unique image it can be clearly seen...
Screenshot 2021-03-01 at 19 57 05

Mi-Lan:
closes #1688


Mi-Lan:
@hexyls @KadenZipfel curious what you guys think...this is only way i found to be working by instantiating safe service for each address

hexyls:
Yep looks good @Mi-Lan! we can probably make this more efficient later on using a multicall contract but I think this is great for now.

pimato (approved):

PR #1784, opened 2021-03-26
Fixes proxy address vm error for history table

Mi-Lan:
- In certain cases like this market http://localhost:3000/#/0x3270f8f76da464e5f38c761b76bea6e7126b798c
- When you go on the history table and click next page it fails on the third
- Because creator is fpmmDeterministicFactory https://blockscout.com/poa/xdai/address/0x9083A2B699c0a4AD06F63580BDE2635d26a3eeF0/transactions
- And that messes up with getOwnerAddress function 
- This fix catches those cases and just displays proxy address

### DIFF
--- a/app/src/components/market/common/history_section/history_select.tsx
+++ b/app/src/components/market/common/history_section/history_select.tsx
@@ -149,7 +149,13 @@ export const History_select: React.FC<Props> = ({
         const response: any[] = await Promise.all(
           fpmmTransactions.map(async item => {
             const safe = new SafeService(item.user.id, context.library)
-            const owners = await safe.getOwners()
+            let owners: string
+            try {
+              const result = await safe.getOwners()
+              owners = result[0].toString()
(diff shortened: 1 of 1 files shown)


app/src/components/market/common/history_section/history_select.tsx
```diff
@@ -149,7 +149,13 @@ export const History_select: React.FC<Props> = ({
         const response: any[] = await Promise.all(
           fpmmTransactions.map(async item => {
             const safe = new SafeService(item.user.id, context.library)
-            const owners = await safe.getOwners()
+            let owners: string
```
KadenZipfel, line 152: This is good, but if there is only one owner then it's slightly less confusing if we change the variable name to `owner`

KadenZipfel (approved):

pimato (approved):
fixed the issue!

Issue #1776, opened 2021-03-24
Incorrect final outcome in invalid scalar market


KadenZipfel:
Currently we're using the `question.answers` array to check whether the final answer is invalid, which is apparently not reliable since it's returning every answer but the final answer, leading to the wrong result being displayed in a certain edge case. I will create an issue to get this fixed at the subgraph level, but will simply use `realitioAnswer` for this check instead.

`userWinnerShares` also appears to be incorrect.

![image](https://user-images.githubusercontent.com/30579067/112386364-1043a980-8cae-11eb-9ee4-22de33f1f667.png)

PR #1785, opened 2021-03-26
Clean up closed.tsx to carefully reference correct data related to user winnings

KadenZipfel:
Closes: https://github.com/protofire/omen-exchange/issues/1776

To test: create markets and carefully track your shares to make sure the correct data is returned. 
Kinda screwed up the commit history :/


app/src/services/cpk.ts
```diff
@@ -1360,11 +1360,13 @@ class CPKService {
         // get mainnet relay signer
         const to = await this.cpk.ethLibAdapter.signer.signer.getAddress()
 
+        const withdrawAmount = amount.sub(RELAY_FEE)
```
hexyls, line 1363: Just gotta remove the relay fee here, must have accidentally stuck around in the merge.

hexyls (approved):
Wasn't total sure how to test the changes but it does appear to work and I could redeem the correct amount:

Screen Shot 2021-03-29 at 3 36 56 pm

pimato (changes requested):
The app crashes on this market: http://localhost:3000/#/0x5e3aba6d741ed0d289b461508472ceacd64be339
Bildschirmfoto 2021-03-29 um 09 13 14

on master everything seems to load just fine:
Bildschirmfoto 2021-03-29 um 09 15 05

Mi-Lan (approved):
I've tested bunch of cases and could't find that error again. Forgot to note which market I was getting it on. Heres couple of cases where it works.

- Also tip for testing you can use history table to check if shares and amounts won align! Found that its pretty accurate! GJ

Screenshot 2021-03-30 at 17 29 44
Screenshot 2021-03-30 at 17 02 31

pimato (approved):

PR #1795, opened 2021-03-30
Bond balance error

### LINKED ISSUE #1777: Market bond should have a wallet balance check
Currently in the market bonding view, both for scalar and categorical markets, there is no check that the user has a sufficient balance causing a somewhat confusing error when the transaction fails. Instead, the button should be disabled and there should be an error similar to that given the same circumstances in the market buy view

Mi-Lan:
closes #1777 
How to test:
- Go to xDai
- For scalar go http://localhost:3000/#/0x0ab13e591d0ad9b2697d2c3a34d25a8155626d30
- For categorical go http://localhost:3000/#/0x2527151d67f784911bd434d629d3b67c6af36a9e
- Click on setOutcome
- Try bonding


pimato (approved):
works great!

hexyls (approved):
🙌

PR #1802, opened 2021-04-03
fix double border on categorial MarketDetailsTab

### LINKED ISSUE #1763: Remove double border of finalizing widget
Bildschirmfoto 2021-03-22 um 10 33 58

link to market: http://localhost:3000/#/0xfbe4eb1fa1d4d869c35b2c41695965a748399611
Please make sure to not remove a border somewhere else with it.

gem1no:
fixes #1763


Mi-Lan (approved):
LGTM! Tested it in couple use cases couldn't find issues

KadenZipfel (approved):
LGTM

hexyls (approved):
🙌

pimato (changes requested):
Looks like there is still a double border.
Market: http://localhost:3000/#/0x634b074e9db599da7f98829bed6547753beb1452
Network: Rinkeby
Bildschirmfoto 2021-04-06 um 15 23 28

pimato (changes requested):
@gem1no please test out all scenarios before you request a review. I found an issue with your fix:
Bildschirmfoto 2021-04-07 um 14 13 34

gem1no:
i tried many ways to do this in a more elegant way. the problem was the ternary operator had a check that was true 2x on the closed categorial market, and displayed 2x borders. i separated the logic out more to be more specific for each case, hopefully this is good enough. it's not completely DRY and feels a little hacky but i tried many different things.

Mi-Lan (changes requested):
@gem1no Seems that you covered all the cases now! But I think implementation is not optimal. I would suggest reading up on Styled Components Props https://styled-components.com/docs/basics#adapting-based-on-props. Since you can pass logic that will evaluate and show border based on condition. I don't think using class is best approach for this. Also there is added marginTop and borderTop property to buttonContainer component and when it evaluates to true it will display top border....
Screenshot 2021-04-19 at 11 57 50

gem1no:
> @gem1no Seems that you covered all the cases now! But I think implementation is not optimal. I would suggest reading up on Styled Components Props https://styled-components.com/docs/basics#adapting-based-on-props. Since you can pass logic that will evaluate and show border based on condition. I don't think using class is best approach for this. Also there is added marginTop and borderTop property to buttonContainer component and when it evaluates to true it will display top border....
> Screenshot 2021-04-19 at 11 57 50

thanks, check out new changes. i think it's pretty nice now :)

app/src/components/market/sections/market_profile/market_status/closed.tsx
```diff
@@ -475,28 +470,25 @@ const Wrapper = (props: Props) => {
...
-                      >
-                        Redeem
-                      </Button>
-                    </BorderedButtonContainer>
-                  )}
+                  <BorderedButtonContainer
+                    borderTop={
+                      (!isConditionResolved && hasWinningOutcomes) || (isConditionResolved && hasWinningOutcomes)
+                        ? true
+                        : false
```
Mi-Lan, line 477: We are almost getting there! So why do you need ternary operators here? Since if it evaluates to true it will return true if not it will return  false...So i would just omit them....heres link to learn more  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator .

gem1no: 👍

pimato (approved):
I couldn´t find any issues.

Mi-Lan (approved):
@gem1no Looks good now! GJ!

PR #1805, opened 2021-04-06
changed message from 0.005 DAI to 5 DAI

### LINKED ISSUE #1747: Increase the required deposit amount
Currently the Omen UI is showing:
Bildschirmfoto 2021-03-18 um 18 14 43

we should rename it to 
```You need to deposit at least 5 DAI.```

### PR DESCRIPTION
**Pull request by gem1no**
fixes #1747

### CODE REVIEW

**Comment by hexyls** [comment]
My bad, reviewed this in https://github.com/protofire/omen-exchange/pull/1806

**Review by hexyls — APPROVED** [review]

File: app/src/components/modal/modal_deposit_withdraw/index.tsx [hunk 1]
```diff
@@ -92,7 +92,7 @@ export const ModalDepositWithdraw = (props: Props) => {
   const DAI = getToken(1, 'dai')
 
   const wallet = exchangeType === ExchangeType.deposit ? daiBalance : xDaiBalance
-  const minDeposit = exchangeType === ExchangeType.deposit ? parseEther('0.005') : parseEther('10')
+  const minDeposit = exchangeType === ExchangeType.deposit ? parseEther('5') : parseEther('10')
```
Mi-Lan, line 95: I would rename this variable since its used both for deposit and withdraw amount. To maybe  minimumExchangeAmount or something similar....

Mi-Lan (changes requested):
Works good man! I added just one small comment. Also one more thing It would be good in the future to be consistent with branch naming like the rest of pr-s. So we usually use fix or feature followed by issue number. So for this it would be fix/#1747 and if it was some feature like the error messages pr you would use feature/#1767 or whatever the number of issue is. This is for future pr-s don't bother changing the current ones.

gem1no:
> Works good man! I added just one small comment. Also one more thing It would be good in the future to be consistent with branch naming like the rest of pr-s. So we usually use fix or feature followed by issue number. So for this it would be fix/#1747 and if it was some feature like the error messages pr you would use feature/#1767 or whatever the number of issue is. This is for future pr-s don't bother changing the current ones.

i noticed that. good to know, i will do this for future pr

KadenZipfel:
> Works good man! I added just one small comment. Also one more thing It would be good in the future to be consistent with branch naming like the rest of pr-s. So we usually use fix or feature followed by issue number. So for this it would be fix/#1747 and if it was some feature like the error messages pr you would use feature/#1767 or whatever the number of issue is. This is for future pr-s don't bother changing the current ones.

To add to this, including the `#` character messes with URLs a little bit so ideal to do without it, e.g. `fix/1747`

KadenZipfel (approved):
LGTM

Mi-Lan (approved):

Issue #1716, opened 2021-03-09
Zero not showing when switching back from scalar market creation


Mi-Lan:
How to replicate:
- Go to scalar market creation input 0 for starting amount
- Click next
- Then go back
- You will see that in place of zero there is just placeholder and not value

https://user-images.githubusercontent.com/33226956/110538613-811e8980-8124-11eb-8a4e-42130197d5b7.mov

PR #1807, opened 2021-04-06
Fix/#1716

hexyls:
closes https://github.com/protofire/omen-exchange/issues/1716

This was pretty confusing, mostly for inputs we display placeholder values when the default value is 0, so had to try and work around that.


pimato (approved):
fixed the issue for me! Works great!

Mi-Lan (approved):
Works!

KadenZipfel (approved):
Super creative, GJ!

Issue #1808, opened 2021-04-06
Creating categorical/scalar market on xDai/Rinkeby failed


pimato:
Network: Rinkeby
Market Type: Scalar market
Transaction: https://rinkeby.etherscan.io/tx/0x15646274546da7f32b63560960ba646bccecbe0db76e28624de289403a99272a

Error:
<img width="619" alt="Bildschirmfoto 2021-04-06 um 13 41 05" src="https://user-images.githubusercontent.com/5337809/113705579-e3e05200-96dd-11eb-8de6-d43c5c5dc2df.png">

Console:
<img width="646" alt="Bildschirmfoto 2021-04-06 um 13 42 21" src="https://user-images.githubusercontent.com/5337809/113705626-f2c70480-96dd-11eb-8c77-e869670103eb.png">

Mi-Lan:
I tried solving it but of no success.

PR #1809, opened 2021-04-06
Check if on mainnet before fetching unclaimed xdai => dai

KadenZipfel:
Closes: https://github.com/protofire/omen-exchange/issues/1808


Mi-Lan (approved):
GJ! I tried the same exact thing when i was trying to solve it today...and yet it didnt work. You a magician man

pimato (approved):

PR #1810, opened 2021-04-07
Fix unnecessary market list reload on xdai relay

hexyls:
Fixes this sort of annoying double loading of the market list we have on the xdai relay atm:

https://user-images.githubusercontent.com/72715177/113811474-99c7a100-97af-11eb-843b-78c52204ea32.mov

PR also sneaks in two fixes for small errors, one to handle undefined `transactions` response from `fetchXdaiTransactionData` and the other to catch that annoying `refetch of undefined` error we get from hot reloading.


pimato (approved):

Mi-Lan (approved):
🚀

PR #1825, opened 2021-04-13
Fixed incorrect account object being passed to IconJazz component

### LINKED ISSUE #1760: Account Jazz-icon different then account button
Bildschirmfoto 2021-03-20 um 19 34 05
Bildschirmfoto 2021-03-20 um 19 34 10

gem1no:
fixes #1760 

The issue appears to have been how the `account` object was being generated from `useConnectedWeb3Context()`

it confused the connected wallet's pub key with the xDai proxy account. i replaced the reference to `account` with `context.rawWeb3Context.account` so we get access to the originating address always

### DIFF
--- a/app/src/components/modal/modal_your_connection/index.tsx
+++ b/app/src/components/modal/modal_your_connection/index.tsx
@@ -219,6 +219,7 @@ export const ModalYourConnection = (props: Props) => {
 
   const fetchAllowance = async () => {
     const owner = context.rawWeb3Context.account
+
     if (relay && owner && context.rawWeb3Context.networkId === networkIds.MAINNET) {
       const collateralService = new ERC20Service(context.rawWeb3Context.library, owner, DAI.address)
       const allowance = await collateralService.allowance(owner, DAI_TO_XDAI_TOKEN_BRIDGE_ADDRESS)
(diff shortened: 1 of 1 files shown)


app/src/components/modal/modal_your_connection/index.tsx
```diff
@@ -288,7 +289,7 @@ export const ModalYourConnection = (props: Props) => {
               <TopCardHeaderLeft>
                 <ConnectionIconWrapper>
                   <ConnectorCircle>{connectorIcon}</ConnectorCircle>
-                  <IconJazz account={account || ''} size={28} />
+                  <IconJazz account={context.rawWeb3Context.account || ''} size={28} />
```
pimato, line 292: can`t we just use the owner variable which has already be assigned to `context.rawWeb3Context.account` ?

pimato (changes requested):

KadenZipfel (approved):
Great work!

app/src/components/modal/modal_your_connection/index.tsx
KadenZipfel: Looks like the owner variable is used in the function scope, but would be good to assign it at the component level once and use it everywhere that way

Mi-Lan (approved):
@gem1no Good job it works well! I would also suggest what kaden and geronimo mentioned about including in the component scope and retrieving value from there...

app/src/components/modal/modal_your_connection/index.tsx
gem1no: i tried using that var but got error. i will check again

hexyls (approved):

pimato (approved):

Issue #1796, opened 2021-03-30
If you input too many decimals on any form, an unhandled error occurs and entire dapp crashes.


gem1no:
![](https://i.imgur.com/oUEZGiG.gif)

tested for market buy/sell/deposit and 
scalar buy/sell/deposit

PR #1839, opened 2021-04-19
fix/1796

hexyls:
closes https://github.com/protofire/omen-exchange/issues/1796 

Also handles an issue where the `BigNumberInput` component was not getting updated when the decimal prop changes.


Mi-Lan (approved):
Works!

pimato (approved):

PR #1844, opened 2021-04-20
wait for subgraph to sync after submitAnswer

hexyls:
Fixes an issue where the UI would not always update after bonding an answer on any non-relay networks.

### DIFF
--- a/app/src/services/cpk.ts
+++ b/app/src/services/cpk.ts
@@ -1284,8 +1284,12 @@ class CPKService {
         txOptions.gas = defaultGas
         return this.execTransactions(transactions, txOptions, setTxHash, setTxState)
       }
-      const txObject = await realitio.submitAnswer(question.id, answer, amount, setTxHash, setTxState)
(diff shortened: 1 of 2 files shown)


Mi-Lan (changes requested):
Great job! Its already much better but seems that it still takes some time to update it...

https://user-images.githubusercontent.com/33226956/115382345-dbefda00-a1d4-11eb-977e-2e67298fcdd4.mov
(Dunno why my whole screen is recorded...

hexyls:
Good point and good catch @Mi-Lan!

I moved things around so the tx confirmation doesn't happen until after subgraph sync 👍

pimato (approved):
works perfect!

Mi-Lan (approved):
This is it man! Great work! Heres one more Rinkeby market i have left if someone else want to test it http://localhost:3000/#/0x0aeeebe923434f7a6b874d59d1a2a66afc7a1655

Issue #1850, opened 2021-04-21
Scalar market with Compound support should  convert cDai back to Dai


pimato:
Currently, if a user wants to redeem winnings for a scalar market with cDai as a collateral the user will redeem cDai instead of Dai. We should include a convert transaction which will allow the user to redeem DAI instead of cDAI.

Issue #1737, opened 2021-03-15
add unwrapping for redeemPositions


hexyls:
We have unwrapping for buy/sells etc but not redeem at them moment.

PR #1856, opened 2021-04-23
fix/1850

hexyls:
closes https://github.com/protofire/omen-exchange/issues/1850
closes https://github.com/protofire/omen-exchange/issues/1737

### DIFF
--- a/app/src/util/tools.ts
+++ b/app/src/util/tools.ts
@@ -358,6 +358,14 @@ export const getCTokenForToken = (token: string): string => {
   }
 }
 
+export const isCToken = (symbol: string): boolean => {
+  const tokenSymbol = symbol.toLowerCase()
+  if (tokenSymbol.startsWith('c')) {
+    return true
+  }
+  return false
+}
+
 /**
  * Gets base token symbol for a given ctoken
  */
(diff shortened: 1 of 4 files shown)


pimato:
I just played around with it and created two scalar markets with compound enabled and redeemed my winnings.

It looks like the message while the market is not yet resolved looks good:
Bildschirmfoto 2021-04-23 um 13 41 26

Once resolved the winning message is always showing the DAI balance at 0.00 DAI: 
Bildschirmfoto 2021-04-23 um 13 42 21

pimato (changes requested):
see comment.

pimato (approved):
Works great now!

app/src/util/tools.ts
```diff
@@ -358,6 +358,14 @@ export const getCTokenForToken = (token: string): string => {
   }
 }
 
+export const isCToken = (symbol: string): boolean => {
+  const tokenSymbol = symbol.toLowerCase()
+  if (tokenSymbol.startsWith('c')) {
+    return true
+  }
```
KadenZipfel, line 365: This could potentially cause problems... I'm thinking we could probably call compound's API instead to do this check

KadenZipfel (changes requested):

app/src/util/tools.ts
hexyls: Oh yeah true! This has to be sync in some places, luckily I discovered we actually have a `CompoundTokenType` so I can check against that.

jpkcambridge (approved):

KadenZipfel (approved):

PR #1860, opened 2021-04-26
set max width of outcomes on list view

hexyls:
Outcomes can be quite long which can mess up the layout in the market list view:

PR is a change from:
Screen Shot 2021-04-26 at 4 17 10 pm

To:
Screen Shot 2021-04-26 at 4 17 21 pm

Users can click through to see the full outcome info if they are interested.


pimato (approved):

KadenZipfel (approved):
LGTM

Issue #1863, opened 2021-04-26
Set Permission transaction using 1,5 mill as gasLimit which results in 200$ transaction visuals


pimato:
Just going through @KadenZipfel checklist here: https://github.com/protofire/omen-exchange/issues/1862 and it seems like that we have a very high hardcoded gas limit for Ethereum Mainnet connection for setting approval/permission:
<img width="1001" alt="Bildschirmfoto 2021-04-26 um 22 26 09" src="https://user-images.githubusercontent.com/5337809/116146287-9f454680-a6de-11eb-8c22-59824c8195da.png">

Are we actually creating a gnosis safe while setting permission? I decreased the gasLimit but it through out of gas: https://etherscan.io/tx/0x2c4371546d6d5f279ad1ba1482236dfc02665239048805fd190b6fc28d534fe4

we are calling `createProxyAndExecTransaction` which indicates we are creating a gnosis safe on set Permission?

PR #1864, opened 2021-04-27
update gas limits

hexyls:
closes https://github.com/protofire/omen-exchange/issues/1863

As discussed in Keybase it would obviously be better if we don't have to hardcode gas for mainnet transactions at all. The main reason we are hardcoding gas is because Gnosis Safe does not revert transactions if internal transactions fail, which makes it impossible for wallets to estimate gas correctly (among other issues). 

I've deployed a new master copy of Gnosis Safe on [Rinkeby](https://rinkeby.etherscan.io/address/0xcb05c7d28766e4ffb71ccbdaf6ae1cec555d61f8#code) and [Mainnet](https://etherscan.io/address/0xcb2e9fa32603cdc2740b82a9a67ed3e977c33416#code) that will always revert if internal transactions fail.

If you want to test if reverts are working you can try the following:
1. Open Omen in two seperate windows
2. Go to a native ETH market on Rinkeby and buy like 0.01 ETH worth of shares but don't confirm the tx
3. In the other window, open the same market and buy 1 ETH of shares, confirm the tx, and wait for confirmation.
4. Now go back and confirm the first tx
5. Slippage caused by the 1 ETH buy should cause the 0.01 ETH buy to fail and the whole tx should revert, refunding the ETH


pimato:
Hey @hexyls do we show this:

> PR also adds a function for claiming stuck eth from a proxy contract.

In the UI somewhere?

pimato (approved):

jpkcambridge:
👍

jpkcambridge (approved):

gem1no (approved):
My gas limit went down from 1.5m to 1m on a tx with compound approval. Really nice work, I'm a big fan of reducing gas costs 🥇. Code looks good too, couldn't see anything wrong with it.

app/src/services/cpk.ts
```diff
@@ -1448,8 +1485,6 @@ class CPKService {
 
   claimDaiTokens = async () => {
     try {
-      const txOptions: TxOptions = {}
-      txOptions.gas = defaultGas
```
KadenZipfel, line 1452: Do we not need to set the gas for this function?

pimato:
I have been able to revert a purchase successfully:
https://rinkeby.etherscan.io/tx/0x655653cf33dd307c50ac2b69bf2d5686d0ec59f2915dfd72564595cfa70ca4de

pimato:
I just tested the new gas estimations on Mainnet with:

- Approving a token
- Buying an outcome token

pimato (approved):

Mi-Lan (approved):
GJ it reverts like you said it would! I tested it on both categorical and scalar!

KadenZipfel (approved):
LGTM! Great work!

Issue #1841, opened 2021-04-19
Add OMN Button in Navigation


pimato:
We need to add a direct visualisation of how much OMN tokens an address has. With that we need to add a Button (no action for now, later a modal needs to open up onClick) with:
- Total OMN user balance
- OMN Icon

Which needs to look like:
<img width="750" alt="OMN-Button" src="https://user-images.githubusercontent.com/5337809/115237358-c4edb100-a11c-11eb-924b-acc5c2d57d49.png">

**Figma:**
[Design Specification](https://www.figma.com/file/R056tSYMjvfwaFidEBItYy/Widgets-V2?node-id=1209%3A141)

OMN Token address: 
https://rinkeby.etherscan.io/token/0x0A08ECa47C56C305F4FeB4fa062AEcd5807BeBb8

PR #1881, opened 2021-04-30
Feature/1841 omn button

gem1no:
#1841 
merged PR 1848 into 1819


app/src/components/common/layout/header/index.tsx
```diff
@@ -284,6 +288,15 @@ const HeaderContainer: React.FC = (props: any) => {
             />
           )}
 
+          <HeaderButton>
+            {relay
+              ? `${formattedxOmenBalance.slice(0, formattedOmenBalance.length - 4)}`
+              : `${formattedOmenBalance.slice(0, formattedOmenBalance.length - 4)}`}
```
Mi-Lan, line 294: I would suggest that it's maybe better to change formatting of Omen token in useConnectedBalance since for xOmen we have 2 decimals and for Omen 3 decimals it causes this on xDai...Maybe we could change it in useConnectedBalance to always have zero decimals so we don't have to format it anywhere else @pimato ?
Screenshot 2021-05-03 at 12 34 11

app/src/components/common/layout/header/index.tsx
```diff
@@ -284,6 +288,15 @@ const HeaderContainer: React.FC = (props: any) => {
             />
           )}
 
+          <HeaderButton>
```
Mi-Lan, line 291: Color of balance should be #86909E instead of black....

app/src/components/common/layout/header/index.tsx
```diff
@@ -284,6 +288,15 @@ const HeaderContainer: React.FC = (props: any) => {
...
+            {relay
```
Mi-Lan, line 292: I just noticed this...After switching from xDai to mainnet....Xdai balance is not showing. This is not related to your pr but I will solve it on omn token bridge....

https://user-images.githubusercontent.com/33226956/116867966-385bec00-ac0e-11eb-9040-6ec4ac30deef.mov

Mi-Lan (changes requested):
Good job! Looks good....I just found couple of small things and also one issue on my bridge pr which needs to be solved....

Mi-Lan (approved):
This is looking great! GJ!

pimato (approved):

PR #1886, opened 2021-05-03
Fix incorrected decimal value shown on scalar market sell and display correct collateral asset on modal success message

### LINKED ISSUE #1866: Reduce transaction modal outcome shares sell amount to two decimals
Looks like we are showing the amount of outcome shares the user wants to sell with all decimals:
Bildschirmfoto 2021-04-27 um 14 41 25

 This is not needed. We should reduce the amount to two decimals.

### LINKED ISSUE #1867: Transaction modal success mesage still showing cDai asset description instead of Dai
just successfully withdrew Dai from pool tokens and only the success message is still showing cDai as asset description even if the amount is already in DAI.

Sign:
Bildschirmfoto 2021-04-27 um 15 04 46

Broadcasted:
Bildschirmfoto 2021-04-27 um 15 05 18

Success:
Bildschirmfoto 2021-04-27 um 15 05 53

cDai description should be changed to DAI

gem1no:
fixes #1866 
fixes #1867


hexyls (approved):
Good job man, works 🙌

app/src/components/market/sections/market_pooling/scalar_market_pool_liquidity.tsx
```diff
@@ -309,7 +309,7 @@ export const ScalarMarketPoolLiquidity = (props: Props) => {
 
       setTxState(TransactionStep.waitingConfirmation)
       setIsTransactionModalOpen(true)
-      let fundsAmount = formatBigNumber(depositedTokensTotal, collateral.decimals)
+      let fundsAmount = formatBigNumber(depositedTokensTotal, collateral.decimals, collateral.decimals)
```
Mi-Lan, line 312: Third parameter for the formatBigNumber is how many decimals it should have. So you are putting 18 decimals.....I would suggest leaving it like before since if you don't provide third parameter it will use the default of 2 decimals which is needed in this case. Also then you don't need to use formatNumber in setMessage below.

Mi-Lan (changes requested):
Hey men GJ works well just added one small comment!

app/src/components/market/sections/market_pooling/scalar_market_pool_liquidity.tsx
gem1no: thanks, will remove the third param in formatBigNumber. I believe `formatNumber` is required for `setMessage` below because that's how it handles numbers < 0.01.

appreciate you reviewing this

pimato (approved):
looks good!

Mi-Lan (approved):
Looks good!

PR #1887, opened 2021-05-04
Migrate settings components to modal and fix "Component rendered outside the provider tree" error 

### LINKED ISSUE #1870: Configure RPC Endpoints Modal Design Specification
Now that the Guild is on the way, the header with the new OMN balance button and a new page navigation is getting crowded. I propose to move the Settings Button into our new connection modal:

RPC

Notes:
- Modal should have the same hight as the Connection Modal
- The complete flow of the settings page should work exactly as is on the `Configure RPC Endpoints - Modal`
- After user set a custom RPC node and clicked on the button "Save" the frontend stores the new RPC node and refreshes the page with the modal showing the new configuration as saved.
- If something goes wrong with the RPC, the UI needs to show the modal so the user can change the RPC. He should not be able to remove the modal only once the RPC connection is working positive.

Figma:
[Design Specification](https://www.figma.com/file/R056tSYMjvfwaFidEBItYy/Widgets-V2?node-id=3832%3A241)
[Playground](https://www.figma.com/proto/aTivxuHmp2l6Sn8nho1Ngu/Playground?page-id=3624%3A91320&node-id=3624%3A91321&viewport=176%2C353%2C0.4668481647968292&scaling=min-zoom)

### LINKED ISSUE #1891: Error: Component rendered outside the provider tree
If [`status`](https://github.com/protofire/omen-exchange/blob/master/app/src/App.tsx#L30) is false in App.tsx we get a `Component rendered outside the provider tree` error. Likely because I added `useConnectedWeb3Context` and `useConnectedBalanceContext` to the [`HeaderContainer`](https://github.com/protofire/omen-exchange/blob/master/app/src/components/common/layout/header/index.tsx#L168) component which is needed for the relay/balances/etc.

Attempted to resolve in https://github.com/protofire/omen-exchange/pull/1890 but it was taking too long, will circle back later.

gem1no:
resolves #1870, resolves #1891 

newModal


pimato:
The buttons have more then 28px padding-bottom:
Bildschirmfoto 2021-05-10 um 15 31 07

app/src/components/settings/settings_view/index.tsx
```diff
@@ -37,28 +36,25 @@ const StatusSection = styled(Row as any)`
   margin-top: 6px;
```
pimato, line 36: can be reduced to 5px

pimato (changes requested):

pimato:
Inputfield has unecessary padding, currently:
Bildschirmfoto 2021-05-10 um 15 38 23

should:
Bildschirmfoto 2021-05-10 um 15 39 19

pimato:
Inputfield has wrong margin-top: 20px:
Bildschirmfoto 2021-05-10 um 15 39 55

should be 12px!

pimato:
What happend to my xDai svg:
Bildschirmfoto 2021-05-10 um 15 43 21

pimato:
padding not according my design specification, currently:
Bildschirmfoto 2021-05-10 um 15 44 13

should:
Bildschirmfoto 2021-05-10 um 15 46 14

app/src/theme/index.js
```diff
@@ -378,7 +378,7 @@ const theme = {
   fixedHeightModal: {
     content: {
       height: '510px',
```
Mi-Lan, line 380: Don't know how did this height get in there but it's causing issues with your connection modal...Because you have scroll just ommiting it or setting it to auto will fix it....
Screenshot 2021-05-12 at 12 00 03

app/src/components/common/icons/IconSettings.tsx
```diff
@@ -1,12 +1,43 @@
 import React from 'react'
+import styled from 'styled-components'
 
-import theme from '../../../theme'
+const Wrapper = styled.svg<{ hoverEffect: boolean | undefined }>`
```
Mi-Lan, line 4: you don't need or undefined since below when you put hoverEffect=false that means that false is default value if not present...

app/src/App.tsx
```diff
@@ -40,7 +39,6 @@ const App: React.FC = () => {
         {!status ? (
           <>
             <HeaderNoRouter />
-            <SettingsViewContainer networkId={networkId} />
```
Mi-Lan, line 43: The use of this was that when rpc node is down settings view would popup and you would have to select other node when default one or one selected is down....I don't see that handled anywhere? This is what i get if node is down. It should display the modal you made...
Screenshot 2021-05-12 at 12 30 52

app/src/components/modal/modal_your_connection/index.tsx
```diff
@@ -103,18 +104,30 @@ const AccountInfoAddress = styled.p`
   font-size: ${props => props.theme.fonts.defaultSize};
   color: ${props => props.theme.colors.textColorDark};
   margin: 0;
+  width: 78px;
+  line-height: ${props => props.theme.fonts.defaultLineHeight}
+  letter-spacing: 0.2px;
+  height: 16px;
```
Mi-Lan, line 110: @gem1no Imma kill you man! No heights or widths :D let the size of content determine that we just set the margins and paddings in most cases....

app/src/components/modal/modal_your_connection/index.tsx
```diff
@@ -103,18 +104,30 @@ const AccountInfoAddress = styled.p`
...
 `
 
 const AccountInfoWallet = styled.p`
   font-size: 12px;
   color: ${props => props.theme.colors.textColorLighter};
   margin: 0;
+  line-height: 14px;
+  height: 14px;
```
Mi-Lan, line 118: Same here...Next time try deleting height and see if its the same don't copy figma...

app/src/components/modal/modal_your_connection/index.tsx
```diff
@@ -103,18 +104,30 @@ const AccountInfoAddress = styled.p`
...
+  margin: 4px 0px;
 `
 
 const CardHeaderText = styled.p`
+  width: 40px;
+  height: 16px;
```
Mi-Lan, line 124: same here....

app/src/components/modal/modal_your_connection/index.tsx
```diff
@@ -164,6 +177,24 @@ const ConnectionModalNavigation = styled(ModalNavigation as any)`
   margin-bottom: 16px;
 `
 
+const IconSettingsWrapper = styled.div`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  margin-left: 65px;
+  width: 40px;
+  height: 40px;
```
Mi-Lan, line 186: same here no need for width and height

app/src/components/modal/modal_your_connection/index.tsx
```diff
@@ -164,6 +177,24 @@ const ConnectionModalNavigation = styled(ModalNavigation as any)`
   margin-bottom: 16px;
 `
 
+const IconSettingsWrapper = styled.div`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  margin-left: 65px;
```
Mi-Lan, line 184: not good idea to set margins statically like this....Instead you can se left margin to auto and set the right one how much is needed to be between it and the change button....

Mi-Lan (changes requested):
Looks good mostly! Left some comments below....

app/src/App.tsx
Mi-Lan: Screenshot 2021-05-12 at 12 57 45
Its the last things in specification.... geronimo wrote it more clearly.......

app/src/components/settings/settings_view/index.tsx
```diff
@@ -51,14 +30,23 @@ const TextLighter = styled.p`
...
+const ButtonContainer = styled.div`
   display: flex;
-  margin-left: auto;
-
-  button:first-child {
-    margin-right: 12px;
-  }
+  flex-direction: row;
+  justify-content: center;
+  width: 364px;
```
Mi-Lan, line 37: width here 🔪

app/src/theme/index.js
```diff
@@ -377,8 +377,8 @@ const theme = {
   },
   fixedHeightModal: {
     content: {
-      height: '510px',
-      width: '382px',
+      minHeight: '510px',
```
pimato, line 380: Lets move towards the modals adapting the height based on the inner components and not set min-heights.

pimato (changes requested):

app/src/components/settings/settings_view/index.tsx
```diff
@@ -212,57 +221,60 @@ const SettingsViewContainer = (props: Props) => {
...
+                ></Input>
+              </Column>
+            </Row>
+          </TopCardHeader>
         )}
+      </ModalCard>
 
-        <ButtonRow>
-          <ButtonRound
+      <ButtonContainer style={{ minHeight: isDropDownActive ? '186.2px' : '302px' }}>
```
Mi-Lan, line 266: Not good to use inline styles like this this is case for passing prop to component. Active dropdown should be handled differently without having to hardcode height...See in our other components how we did it...

app/src/components/settings/settings_view/index.tsx
```diff
@@ -107,24 +95,43 @@ const StatusBadge = styled.div<{ status: boolean }>`
...
-  margin-top: 20px;
-
+  margin: 12px 0px;
   ${TextfieldCSS};
-`
-const SettingsWrapper = styled(ListCard)`
-  min-height: initial;
+  padding: 12px 20px;
+  width: 320px;
+  height: ${props => props.theme.textfield.height};
```
Mi-Lan, line 102: same here....

app/src/components/settings/settings_view/index.tsx
```diff
@@ -51,14 +30,23 @@ const TextLighter = styled.p`
...
-    margin-right: 12px;
-  }
+  flex-direction: row;
+  justify-content: center;
+  width: 364px;
+  padding: 0px;
+  margin-top: 24px;
+  height: 100%;
+`
+const SetAndSaveButton = styled(ButtonRound)`
```
Mi-Lan, line 42: Can you see how I did deposit and withdraw buttons and base it on that....You are hardcoding widths again here...

Mi-Lan (changes requested):
Added some comments....

app/src/theme/index.js
gem1no: got it. FYI before i made this change it was set to `height: 510px`. should i change the name of this variable now that it's the opposite of 'fixed height'? confusing naming

app/src/components/settings/settings_view/index.tsx
gem1no: 🤯 
it worked
TIL `flex:1` 
lol

hexyls (changes requested):
Hey dude this is looking great!

atm I am getting quite a few `Cannot update a component from inside the function body of a different component.` errors. This is because of those `props.setStatus(true)` calls in `ConnectedWeb3`, I know these were there previously but it would be awesome to fix this in this PR. I think you can just move those calls out of the component function body and into their own useEffect.

gem1no:
> Hey dude this is looking great!
> 
> atm I am getting quite a few `Cannot update a component from inside the function body of a different component.` errors. This is because of those `props.setStatus(true)` calls in `ConnectedWeb3`, I know these were there previously but it would be awesome to fix this in this PR. I think you can just move those calls out of the component function body and into their own useEffect.

should be good now

hexyls (approved):

pimato (changes requested):
@gem1no it looks like we still have some padding-right on the Inputfield:
Bildschirmfoto 2021-05-31 um 11 44 37

Mi-Lan:
@pimato @gem1no Wondering do we really need these two arrow back and close icons for the modal when RPC is broken? Since they don't do anything and just display message ''Please set your RPC endpoint!'
Screenshot 2021-06-01 at 15 06 16

Mi-Lan:
Also good job on catching broken RPC I believe this is desired behaviour! Heres example how to trigger it. Not sure what you guys were referring in comments? 

https://user-images.githubusercontent.com/33226956/120328735-69057300-c2eb-11eb-889f-b7ecc1d2e89b.mov

Mi-Lan:
@gem1no Yeah there seems that more work needs to be done since if you fuck up the mainnet url xDai is displaying modal. Check out line 294 in networks.ts.

pimato:
> @pimato @gem1no Wondering do we really need these two arrow back and close icons for the modal when RPC is broken? Since they don't do anything and just display message ''Please set your RPC endpoint!'
> Screenshot 2021-06-01 at 15 06 16

Great point! Yeah it would be great to have the arrow-left button and the close modal removed.

pimato:
the pick-connection-modal is now decreased in height:
Bildschirmfoto 2021-06-03 um 15 44 40

we should add a min-height:510px to the `pick-connection-modal` ONLY.

pimato (changes requested):
see comments

app/src/App.tsx
```diff
@@ -1,33 +1,60 @@
...
   const windowObj: any = window
 
   const ethereum = windowObj.ethereum
   const networkId = ethereum && ethereum.chainId
   const [status, setStatus] = useState(true)
   const network = getNetworkFromChain(networkId)
+  const [settingsView, setSettingsView] = useState(false)
+
+  useInterval(() => {
+    console.log(status)
```
pimato, line 39: Please remove the console.log

pimato (changes requested):
console.log

app/src/App.tsx
gem1no: oops!

pimato (approved):

Mi-Lan:
> @gem1no Yeah there seems that more work needs to be done since if you fuck up the mainnet url xDai is displaying modal. Check out line 294 in networks.ts.

Seems that this issue is still present....So how to trigger is that you just mess up mainnet url in networks and go to xDai...

https://user-images.githubusercontent.com/33226956/121017562-b37c6900-c79d-11eb-9eb5-4852d1bfd9cf.mov

Mi-Lan:
steps to reproduce:
1)Clear localStorage rpc object
2)Go into networks mess up both url and line 101 and line 104 since they are the same default rpc
Screenshot 2021-06-07 at 19 57 07
2)switch the network to xDai
3)that above happens

Mi-Lan:
So its still not working for me...
1)In networks go into xDai fuckup both default urls in this case rpc.xdaichain.com
Screenshot 2021-06-09 at 22 18 00
2)go and switch network to mainnet 

So one indicator to check which rpc you are checking is to go into networks and see what you are calling in your case I noticed while on mainnet you are checking xDai url...
Screenshot 2021-06-09 at 22 19 22

app/src/App.tsx
```diff
@@ -62,6 +60,14 @@ const App: React.FC = (props: any) => {
...
   }, [ethereum])
 
+  useEffect(() => {
+    window.onerror = function(errorMsg) {
+      if (errorMsg) {
+        windowObj.location.reload(true)
+      }
+    }
+  }, [network, windowObj.location])
+
```
Mi-Lan, line 70: also I don't think this is good practice to catch error like this and reloading page....

app/src/App.tsx
```diff
@@ -1,47 +1,96 @@
...
+      localStorage.setItem(
+        'rpcAddress',
+        JSON.stringify({
+          url: networks[networkIds.XDAI].url || networkId.MAINNET,
+          network: networkIds.XDAI || networkIds.MAINNET,
+          index: 0,
+        }),
+      )
+    }
+  }, FETCH_RPC_INTERVAL)
```
Mi-Lan, line 55: looking at the code this part is little fishy to me possibly...are we checking if it's relay?

gem1no: > So its still not working for me...
> 1)In networks go into xDai fuckup both default urls in this case rpc.xdaichain.com
> Screenshot 2021-06-09 at 22 18 00
> 2)go and switch network to mainnet
> 
> So one indicator to check which rpc you are checking is to go into networks and see what you are calling in your case I noticed while on mainnet you are checking xDai url...
> Screenshot 2021-06-09 at 22 19 22

yeah its because of the way i have infuraURL set to both the URLs so if one fails it will fall back to the other but ive actually been trying to find a fix for this today, noticed some of the same issues you mentioned here. let's see how rest of today/tonight goes.

gem1no: also thanks for the networks tab tip. ill start using that too

Issue #1893, opened 2021-05-05
Replace Enable Dai View in Connection Modal with default Omen Account View


pimato:
Now that Omen is ERC-20 compatible and allows bridging of other assets besides native xDai token, we should replace the connection modal `Enable Dai - Component` with a default Omen Account view.

Currently:
<img width="355" alt="Bildschirmfoto 2021-05-05 um 15 31 28" src="https://user-images.githubusercontent.com/5337809/117149026-19e52480-adb7-11eb-848d-d3b0d02c9311.png">

Should show as default:
<img width="571" alt="Bildschirmfoto 2021-05-05 um 15 32 42" src="https://user-images.githubusercontent.com/5337809/117149069-25385000-adb7-11eb-99a5-abfc86f0f2f3.png">

Notes:
- If the users gnosis safe proxy is not holding any assets, the withdraw button should be disabled.

PR #1894, opened 2021-05-05
Feature/1884 deposit withdraw v2

### LINKED ISSUE #1884: Deposit/Withdraw Modal Version 2 - Design Specification
Withdraw:Depoist

Note:
- Replace Approval Component with new approval button
- Replace Min. Amount info component with a new deposit/withdraw-form structure
    - Show bridge fee 

**Figma**:
[Design Specification](https://www.figma.com/file/R056tSYMjvfwaFidEBItYy/Widgets-V2?node-id=3832%3A241)
[Playground](https://www.figma.com/proto/aTivxuHmp2l6Sn8nho1Ngu/Playground?page-id=1892%3A54309&node-id=1892%3A54420&viewport=-474%2C222%2C0.10631878674030304&scaling=min-zoom)

Mi-Lan:
closes #1884  #1893


Mi-Lan:
@KadenZipfel fucked up commit history here cause i started working on it off of Omen bridge pr before it was merged 😥  Should I do something?

app/src/components/modal/modal_deposit_withdraw/index.tsx
```diff
@@ -56,20 +63,32 @@ const WalletText = styled.div`
...
+const ExchangeDataItem = styled.div`
   display: flex;
-  margin-top: 20px;
-  padding: 16px 20px;
-  border-radius: 4px;
-  border: ${props => props.theme.borders.borderLineDisabled};
-  line-height: 20px;
+  justify-content: space-between;
+  width: 100%;
+  line-height: 16px;
```
gem1no, line 77: i learned from the very best, so this value should be passed through `props.theme.fonts.defaultLineHeight`

app/src/components/modal/modal_deposit_withdraw/index.tsx
```diff
@@ -374,41 +410,98 @@ export const ModalDepositWithdraw = (props: Props) => {
...
+              <ApproveButton
+                disabled={allowanceState !== ButtonStates.idle}
+                extraText
+                onClick={approve}
+                state={allowanceState}
+              >
+                {allowanceState === ButtonStates.idle &&
+                  `Approve ${currencySelected === ExchangeCurrency.Dai ? 'DAI' : 'OMN'}`}
+                {allowanceState === ButtonStates.working && 'Approving'}
+                {allowanceState === ButtonStates.finished && 'Approved'}
```
hexyls, line 483: after setting allowance the green Approved message shows up just fine but is quickly hidden again. I think we might need two seperate allowanceStates for omn and dai. Then for example we can update `daiWalletAllowance` and to be true if `daiAllowanceState` is `ButtonStates.finished` so that the Approved message can continue to be shown after setting allowance.

hexyls (changes requested):
Looking good man! We can add some extra allowance state then I think this is good to goooo

app/src/components/modal/modal_deposit_withdraw/index.tsx
pimato: great feedback and can confirm regarding the quickly hidden thing.

Mi-Lan:
@hexyls @pimato Fixed it! Heres video of how it is now....https://www.dropbox.com/s/dpp6uh1ulxuwr4i/Screen%20Recording%202021-05-13%20at%2018.11.51.mov?dl=0

pimato:
Active button has the wrong border, currently:
#E8EAF6
Bildschirmfoto 2021-05-13 um 21 22 22


Should be:
#3F51B5 

Bildschirmfoto 2021-05-13 um 21 22 52

pimato:
pending approval button has active border button state:
Bildschirmfoto 2021-05-13 um 22 11 01

should have disabled border color (like on the right):
Bildschirmfoto 2021-05-13 um 22 11 44

pimato:
I am just trying to withdraw some GEN, and it looks like the icon has no hover message:
Bildschirmfoto 2021-05-17 um 13 33 33

pimato:
It looks like after a successful withdraw on xDai, a mainnet claim triggered an error on the console:
```gnosis-conditional-exchange::Services::CPKService Error trying to claim Dai tokens from xDai bridge Cannot read property 'map' of null```

**Review by KadenZipfel — COMMENTED** [review]
My main concern with this is that when we go to add more tokens, we'll have to do some rewriting since many things are manually set when it comes to OMN/DAI. Ideally we have a xdaiRelay token list on the dxTokenRegistry, and we read from that token list to automatically read token properties to render the same things without having to add any code. Depending on priorities it wouldn't be too bad to ship with this to start with, but would save a lot of trouble in the future if we can get it right from the start

**Comment by Mi-Lan** [comment]
> My main concern with this is that when we go to add more tokens, we'll have to do some rewriting since many things are manually set when it comes to OMN/DAI. Ideally we have a xdaiRelay token list on the dxTokenRegistry, and we read from that token list to automatically read token properties to render the same things without having to add any code. Depending on priorities it wouldn't be too bad to ship with this to start with, but would save a lot of trouble in the future if we can get it right from the start

@KadenZipfel Yeah I agree if we have plans to add more tokens into it. I wasn't aware of that...I was thinking that this will be solely used for the xDai and Omen token. But it would then also require changes on the UI side?

File: app/src/components/button/button_stateful/index.tsx [hunk 3]
```diff
@@ -40,29 +40,37 @@ const IdleDisabledCSS = css`
...
 `
 
-const Text = styled.span<{ hide: boolean }>`
-  opacity: ${props => (props.hide ? '0' : '1')};
+const Text = styled.span<{ hide: boolean; state: ButtonStates }>`
+  display: ${props => (props.hide ? 'none' : 'flex')};
   position: relative;
   user-select: none;
   z-index: 1;
+  ${props => props.state === ButtonStates.idle && 'font-weight:500'};
```
gem1no, line 65: if you wanted to use the style library you could replace line 65 with 
`font-weight: ${props => props.state === ButtonStates.idle && props.theme.buttonSecondary.weight};`

app/src/components/button/button_stateful/index.tsx
```diff
@@ -40,29 +40,37 @@ const IdleDisabledCSS = css`
     opacity: 0.5;
   }
 `
+const SuccessCSS = css`
```
gem1no, line 43: I assume there's no way to get these to show without `!important`? generally speaking makes code harder to maintain

app/src/components/common/tooltip/img/IconInfo.tsx
```diff
@@ -1,8 +1,25 @@
...
-export const IconInfo = () => {
+interface Props {
+  hasCircle?: boolean
+  style?: any
+}
+
+const SvgStyle = styled.svg`
+  &:hover {
+    circle {
+      stroke: ${props => props.theme.dropdown.buttonBorderColorHover} !important;
```
gem1no, line 12: here

app/src/theme/overrides/react_tooltip_styles_override.tsx
```diff
@@ -73,7 +73,7 @@ export const ReactTooltipStylesOverride = css`
   }
 
   .scalarValueTooltip {
-    ${DefaultTooltip}
+    ${DefaultTooltip};
     padding: 5px 8px !important;
     font-weight: 500 !important;
```
gem1no, line 78: and these

gem1no (changes requested):
see comments! just small changes, maybe the `!important` css must stay as is, not sure.

app/src/components/button/button_stateful/index.tsx
Mi-Lan: Yeah in general it isn't good to use important but in this case I think it makes sense since this class is applied conditionally I don't see other way without repeating some code to increase specificity of that class. Do you have some suggestion?

app/src/theme/overrides/react_tooltip_styles_override.tsx
Mi-Lan: Only green code is written to this pr here I just put ";" cause it was screaming some small error at me . Dunno how else would this be written though?

app/src/components/common/tooltip/img/IconInfo.tsx
Mi-Lan: Good catch not needed here!

Mi-Lan:
@gem1no Included changes in the guild token lis pr #1921 and answered comments

PR #1908, opened 2021-05-11
Guild Token Lock

### LINKED ISSUE #1902: Guild Membership Modal - Design Specification
Now that we are done with the liquidity rewards feature we need to move ahead with the Omen Guild. The first step is to allow the user to lock their OMN tokens to participate as a governor on the Omen platform. This is the first version of the modal which will be extended with more features over the couple of weeks:

guild-membership

Note:
- This modal should be visible once the user clicks on the OMN balance button in the header.
- **Omen Account** is showing the currently available deposited Balance. If the user connects directly to xDai we need to show **Wallet Balance** instead.
- **Locked in Guild** is the amount of OMN tokens locked by the user.
- **Vote weight** is users voting power based on the amount of OMN tokens locked.
- Unlocked Date should be the date the user is able to unlock his/her locked OMN tokens.

**Figma**:
[Design Specification](https://www.figma.com/file/aTivxuHmp2l6Sn8nho1Ngu/Playground?node-id=3914%3A2410)
[Playground](https://www.figma.com/proto/aTivxuHmp2l6Sn8nho1Ngu/Playground?page-id=3914%3A0&node-id=3914%3A2410&viewport=598%2C445%2C0.42017415165901184&scaling=min-zoom)

Mi-Lan:
closes #1902


pimato (changes requested):
column-gap, should be increased from 10px to 20px

pimato:
Disabled button has not the right font color, currently:
Bildschirmfoto 2021-06-04 um 21 14 41

Has a opacity of 0.5 which should be removed!

Correct styling can be found here: https://www.figma.com/file/R056tSYMjvfwaFidEBItYy/Widgets-V2?node-id=3832%3A241

pimato:
margin-top should be increased from 28px to 32px:
Bildschirmfoto 2021-06-04 um 21 19 46

pimato:
Replace wrong arrow icon:
Bildschirmfoto 2021-06-04 um 21 20 40

With the already used one like this (can be found in buy/sell table):
Bildschirmfoto 2021-06-04 um 21 21 26

pimato:
Transaction status invisible after signing:
Bildschirmfoto 2021-06-04 um 21 23 00

and in general no default transaction process (Sign Transaction -> Transaction submitted -> Tranasaction confirmed)

pimato:
Bildschirmfoto 2021-06-04 um 21 26 31

add `remaining` like:
`61 Days remaining`

pimato:
if the user has no locked tokens in the Guild, we should still show that there is a Unlock Date like this:
Connection

pimato:
If a user already has tokens locked with an unlock date, we should tell the user the new unlock date in the `lock OMN token modal` if he/she wants to lock more tokens into OMN governance which will change the Unlock Date:

https://user-images.githubusercontent.com/5337809/120853817-856c0e80-c57c-11eb-9b41-c29496de430b.mov

pimato:
After the OMN lock process is successfull we should redirect the user to the default Modal which shows the new changes.

Currently:
https://user-images.githubusercontent.com/5337809/120853974-be0be800-c57c-11eb-91bc-9115d8a9cf6b.mov

pimato:
tooltip, should be at the top:
Bildschirmfoto 2021-06-22 um 17 40 14

pimato:
remove margin-bottom on the last row item here:
Bildschirmfoto 2021-06-22 um 17 40 24

also here:
Bildschirmfoto 2021-06-22 um 17 41 22

pimato:
tooltip should only be shown if the user actually hovers over the icon:
https://user-images.githubusercontent.com/5337809/122956485-54177f00-d381-11eb-8e52-743c642c5b9b.mov

pimato:
It looks like the `Lock OMN` button in the `Lock Omen - Modal` is active and gets disabled once the user sees the modal:
https://user-images.githubusercontent.com/5337809/122957703-7b228080-d382-11eb-8f49-29c9cdc36baa.mov

pimato:
Vote weight is still showing the transition on the default Omen Guild Membership modal after I did a new locking:
https://user-images.githubusercontent.com/5337809/122971438-584a9900-d38f-11eb-85ed-e339df541d18.mov

pimato:
could we add hover styling to the circle info icon:

https://user-images.githubusercontent.com/5337809/123606827-96b3de00-d7fd-11eb-9e8b-2e6bc43f65be.mov

pimato (changes requested):
see comments

pimato (approved):
Awesome work!

app/src/components/modal/modal_lock_yo_tokens/index.tsx
```diff
@@ -0,0 +1,425 @@
...
+`
+const PercentageText = styled.span<{ lightColor?: boolean }>`
+  ${props => props.lightColor && `color:${props.theme.colors.textColorLighter}`};
+`
+
+const ModalLockTokens = (props: Props) => {
+  const {
+    context,
+    fetchBalances,
+
```
KadenZipfel, line 104: white space 🐙

app/src/components/modal/modal_transaction/index.tsx
```diff
@@ -137,7 +137,7 @@ export const ModalTransaction = (props: Props) => {
             ? 'Transaction Confirmed'
             : txState === TransactionStep.error
             ? 'Transaction Failed'
-            : ''}
+            : 'WHAAAAT'}
```
KadenZipfel, line 140: I feel like we shouldn't keep this 😆

app/src/services/guild.ts
```diff
@@ -0,0 +1,115 @@
...
+  omenGuildAddress: string
+
+  constructor(provider: Web3Provider, network: number) {
+    const signer = provider.getSigner()
+    this.user = signer
+    this.network = network
+    this.provider = provider
+    this.omenGuildAddress = getContractAddress(network, 'omenGuildProxy')
+  }
+  get getOmenGuildContract(): Contract {
```
KadenZipfel, line 75: Just a stylistic preference, but can you clean up the spacing between functions in this file?

app/src/util/tools.ts
```diff
@@ -66,6 +66,17 @@ export const formatDate = (date: Date, utcAdd = true): string => {
...
+  return moment(date)
+    .tz('UTC')
+    .format(`MMMM Do, YYYY - HH:mm UTC`)
+}
+export const daysUntil = (timestamp: number): number => {
+  const current = moment()
+  const day = moment.unix(timestamp)
+
+  return Math.ceil(moment.duration(day.diff(current)).asDays())
+}
```
KadenZipfel, line 79: Do we have tests for these?

app/src/util/tools.ts
```diff
@@ -134,12 +145,16 @@ export const convertLocalToUTC = (date: Date): Date => {
...
+  try {
+    return (
+      a
+        .mul(scale)
+        .div(b)
+        .toNumber() / scale
+    )
+  } catch {
+    return 999999999999999
+  }
```
KadenZipfel, line 157: Why do we return this in the catch?

app/src/components/modal/modal_lock_yo_tokens/index.tsx
```diff
@@ -0,0 +1,425 @@
...
+    try {
+      if (cpk) {
+        let allowanceAddress
+        if (context.networkId === networkIds.MAINNET && !context.relay) {
+          allowanceAddress = await omen.tokenVault()
+        } else {
+          allowanceAddress = cpk.address
+        }
+
+        const omenAddress = await omen.omenTokenAddress()
```
KadenZipfel, line 180: Would be good to grab via `networks.getContractAddress()` for consistency

app/src/components/modal/modal_lock_yo_tokens/index.tsx
```diff
@@ -0,0 +1,425 @@
...
+    })()
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [networkId, account, cpk])
+
+  const getTokenLockInfo = async () => {
+    try {
+      if (cpk?.address) {
+        let address
+        if (context.networkId === networkIds.MAINNET && !context.relay) {
```
KadenZipfel, line 148: Why is this only on mainnet?

app/src/services/guild.ts
```diff
@@ -0,0 +1,115 @@
...
+    return new ethers.Contract(this.omenGuildAddress, GuildAbi, this.provider).connect(this.user)
```
KadenZipfel, line 76: We should do this in the constructor and assign it to a class property for a bit of an efficiency improvement

app/src/components/modal/modal_lock_yo_tokens/index.tsx
KadenZipfel: Same with other places

app/src/services/cpk.ts
```diff
@@ -1473,20 +1476,90 @@ class CPKService {
...
+  lockTokens = async (
+    amount: BigNumber,
+    setTxState: (step: TransactionStep) => void,
+    setTxHash: (arg0: string) => void,
+  ) => {
     try {
-      const txOptions: TxOptions = {}
-      txOptions.gas = defaultGas
+      // setTxHash: (arg0: string) => void
+      //     setTxState: (step: TransactionStep) => void
```
KadenZipfel, line 1487: Prob wanna remove these comments(?)

app/src/services/cpk.ts
```diff
@@ -1473,20 +1476,90 @@ class CPKService {
...
+        ]
+
+        const { transactionHash } = await this.execTransactions(transactions, txOptions)
+
+        return transactionHash
+      } else {
+        const transaction: Transaction = await OmenGuild.unlockTokens(amount)
+
+        return transaction
+      }
```
KadenZipfel, line 1560: Should there also be an if `!mainnet && !relay`, unlock tokens via cpk and transfer to user address?

KadenZipfel (changes requested):
I'm extra picky on this one, but it's looking really good!

app/src/components/modal/modal_transaction/index.tsx
Mi-Lan: ohh yeah sorry about that I was testing some shit...

app/src/util/tools.ts
Mi-Lan: So the issue with that is daysUntil uses current time to calculate and I couldn't create tests reliably with that? Do you have some suggestion maybe? But will create tests for formatLockDate!

app/src/util/tools.ts
Mi-Lan: Dunno how these things slipped up I will go through the code again...

app/src/components/modal/modal_lock_yo_tokens/index.tsx
Mi-Lan: aha okay! My thinking was to maybe simplify it like this since it will always fetch appropriate omenTokenAddress for the current guild we are on....And to handle network checking in the OmenGuildService

app/src/components/modal/modal_lock_yo_tokens/index.tsx
Mi-Lan: So that is because other then on mainnet we are checking cpk address for the votes user has and only on mainnet users address. This needs to be checked how it will work with cpk once we deploy it somewhere other then rinkeby...

app/src/components/modal/modal_lock_yo_tokens/index.tsx
Mi-Lan: I fixed it so that It's fetched from KnownTokens but it might be worth considering fetching them from directly from Guild

app/src/services/guild.ts
Mi-Lan: Ohh yeah great suggestion! I remember trying to implement it that way but I had some problems with some asynchronous stuff...But now its not there! Great call!

Mi-Lan:
@KadenZipfel Thanks for detailed feedback!

app/src/util/tools.ts
KadenZipfel: Maybe this might help: https://dev.to/jbranchaud/test-timing-based-js-functions-with-jest-5be

app/src/components/modal/modal_lock_yo_tokens/index.tsx
KadenZipfel: I just noticed that `getContractsAddress()` doesn't work in this case so I think your method is better

KadenZipfel (approved):
Looks good to go! @Mi-Lan can you make an issue to add a test for `daysUntil`? Possibly useful for reference: https://dev.to/jbranchaud/test-timing-based-js-functions-with-jest-5be

app/src/util/tools.ts
Mi-Lan: Thank you for the feedback on keybase! Fixed it!

Issue #1869, opened 2021-04-28
Pool Overview/Liquidity Rewards Design Specification


pimato:
With the new Liquidity Rewards feature we need to find a way to better structure the information and separate between overall pool statistic and rewards statistics:

Markets without an active liquidity rewards program:
![pool](https://user-images.githubusercontent.com/5337809/116375964-15959600-a810-11eb-973a-f9001d70b3e1.png)

Markets WITH active liquidity rewards program:
![pool-active](https://user-images.githubusercontent.com/5337809/116376087-33fb9180-a810-11eb-9a3f-0cfaefb47617.png)

Figma:
[Design Specification](https://www.figma.com/file/R056tSYMjvfwaFidEBItYy/Widgets-V2?node-id=2987%3A8760)

PR #1910, opened 2021-05-12
Liquidity Rewards UI

gem1no:
#1869

### DIFF
--- a/app/src/components/common/form/form_error/index.tsx
+++ b/app/src/components/common/form/form_error/index.tsx
@@ -2,12 +2,16 @@ import React, { HTMLAttributes } from 'react'
 import styled from 'styled-components'
 
 const FormErrorWrapper = styled.div`
-  color: ${props => props.theme.colors.error};
-  font-size: 13px;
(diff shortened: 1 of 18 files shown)


pimato (changes requested):
Great work!

After seeing my design changes life, I think it would be better if we put Current APY to the top of the liquidity rewards column:
Bildschirmfoto 2021-05-17 um 22 12 52

pimato:
the Header of the column has pure black as the color:
Bildschirmfoto 2021-05-17 um 22 14 37

Design specification says `#37474F`

pimato:
Why do we have the margin-right: 5px here?
Bildschirmfoto 2021-05-17 um 22 15 59

(looks like it is not new code?)

pimato (changes requested):
Now that we have the `<0.01` Indicator I think we should remove the `+` here:
Bildschirmfoto 2021-05-18 um 10 06 48

pimato (approved):

hexyls (approved):

Issue #1912, opened 2021-05-12
Add deployment of liquidity rewards contract to batch actions on market creation


pimato:
Now that we almost done with integrating the liquidity rewards feature into Omen it is important that we add the deployment of such liquidity rewards contract on market creation via gnosis-safe-proxy-kit batch transaction feature.

Default liquidity rewards contract for now:

- No cap
- No timelock
- duration = market duration - 1 day
- Rewards Asset: OMN

PR #1927, opened 2021-05-20
Add deployment of liquidity rewards contract to batch actions on market creation

KadenZipfel:
Closes: https://github.com/protofire/omen-exchange/issues/1912

To test: Create a categorical or scalar market as per usual, check the tx on etherscan, go to logs tab and check that the `DistributionCreated` event was executed.
- Currently only on Rinkeby.
- Currently set to staking factory disconnected from subgraph so won't see any liquidity rewards front end stuff here


hexyls (approved):
Looks good and `DistributionCreated` fires off as expected 👍

I think you mentioned this before but just asking so I can get a better understanding - so there are no security issues creating a distribution client side because the guild still has to fund the distribution right?

Mi-Lan (approved):
Works great event executes! 
Screenshot 2021-05-24 at 12 48 00
- Only thing I found and not sure if this is issue or my not understanding something. But when I create market with DAI as collalteral it doesn't fire off the event? Is that how it should be or?
- Heres one tx from market creation where it doesn't fire https://rinkeby.etherscan.io/tx/0x3720feed40ee0886e277f9c51d5f59b095fa3e73dbc9d9c4149856f4e278129f
I tested it more and couldn't replicate the error above so you can ignore it......

pimato (changes requested):
So it looks like several attempts to create a market lead to an error:
Bildschirmfoto 2021-05-24 um 15 10 25

pimato (approved):
Works!

Issue #1853, opened 2021-04-22
Fetch OMN price data


KadenZipfel:
Replaces: https://github.com/protofire/omen-subgraph/issues/155

To calculate liquidity mining APY we need price data for the OMN token. This can either be sourced via the swapr subgraph or from an API, e.g. Coingecko.

gem1no:
looks good. circling back on this and doing proper testing over weekend/monday

PR #1928, opened 2021-05-20
Fetch OMN price data

KadenZipfel:
Closes: https://github.com/protofire/omen-exchange/issues/1853

Prices are pretty wild because it's all rinkeby data

### DIFF
--- a/app/src/components/market/sections/market_pooling/scalar_market_pool_liquidity.tsx
+++ b/app/src/components/market/sections/market_pooling/scalar_market_pool_liquidity.tsx
@@ -601,7 +601,7 @@ export const ScalarMarketPoolLiquidity = (props: Props) => {
       omnToken,
       cpk.address,
       1, // Assume pool token value is 1 DAI
-      Number(formatBigNumber(tokenPrice, omnToken.decimals, omnToken.decimals)),
+      tokenPrice,
       Number(liquidityMiningCampaign.endsAt),
       liquidityMiningCampaign.rewardAmounts[0],
       Number(liquidityMiningCampaign.duration),
--- a/app/src/components/market/sections/market_pooling/scalar_market_pool_liquidity.tsx
+++ b/app/src/components/market/sections/market_pooling/scalar_market_pool_liquidity.tsx
@@ -20,6 +20,7 @@ import {
   useSymbol,
 } from '../../../../hooks'
 import { GraphResponseLiquidityMiningCampaign } from '../../../../hooks/useGraphLiquidityMiningCampaigns'
+import { useTokenPrice } from '../../../../hooks/useTokenPrice'
 import { StakingService } from '../../../../services/staking'
 import { getLogger } from '../../../../util/logger'
 import {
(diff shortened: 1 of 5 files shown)


gem1no:
code looks good. circling back on this and doing proper testing over weekend/monday

app/src/hooks/useTokenPrice.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const [tokenPrice, setTokenPrice] = useState(new BigNumber(0))
+
+  const { data: data, error: error, loading: loading } = useQuery<GraphResponse>(query, {
+    notifyOnNetworkStatusChange: true,
+    skip: false,
+    variables: { id: address },
+  })
+
+  useEffect(() => {
+    if (data && data.tokenPrice) {
```
hexyls, line 33: I think the path here might be `data.token.usdPerToken`, prob need to update the `GraphResponse` type to match graph data

app/src/components/market/sections/market_pooling/scalar_market_pool_liquidity.tsx
```diff
@@ -571,6 +572,8 @@ export const ScalarMarketPoolLiquidity = (props: Props) => {
     }
   }
 
+  const { tokenPrice } = useTokenPrice(getToken(networkId, 'omn').address)
```
hexyls, line 575: I think token address should be lowercase here like you did with the others, could maybe update `useTokenPrice` to make addresses lowercase automatically so we don't have to worry about it in the future

app/src/hooks/useTokenPrice.tsx
```diff
@@ -0,0 +1,42 @@
...
+      setTokenPrice(bigNumberify(data.tokenPrice))
```
hexyls, line 34: Once the path mentioned above is updated, bigNumberify is freaking out on [this market](http://localhost:3000/#/0xb4277c368b817db02bab8b684c9bfe10200461f7) with an `invalid BigNumber string value` error message, not really sure why 🤔

hexyls (changes requested):

pimato (approved):
works!

hexyls (approved):
🙌

app/src/hooks/useTokenPrice.tsx
```diff
@@ -0,0 +1,46 @@
...
+    if (data && data.token.usdPerToken) {
+      setTokenPrice(Number(data.token.usdPerToken))
+    }
+  }, [data])
+
+  console.log(tokenPrice)
+
+  return {
+    tokenPrice,
+    status: error ? Status.Error : loading ? Status.Loading : Status.Ready,
```
Mi-Lan, line 44: Small suggestion....When I was working on history table I found out that If its set like this it doesn't always display state correctly and when changed to this 
`status: error ? Status.Error : loading ? Status.Loading : data ? Status.Ready : Status.Loading,`
heres console logs with that implemented in this pr 
Screenshot 2021-06-01 at 14 29 31
And here is it like its now :
Screenshot 2021-06-01 at 14 29 48
Although dunno if it causes some other issues when done like this?

Mi-Lan (approved):
Looks good!

Issue #1926, opened 2021-05-20
Duplicat Set Allowance component by switching from ETH to WETH


pimato:
Network: Rinkeby

https://user-images.githubusercontent.com/5337809/118973560-197d8980-b972-11eb-94c1-170368164cb3.mov

Issue #1930, opened 2021-05-21
Buying with WETH does not work


hexyls:
Related https://github.com/protofire/omen-exchange/issues/1926 

It appears as if you can buy with WETH if ETH is also enabled, but in reality it just uses ETH not WETH.

PR #1931, opened 2021-05-21
fix/1926

hexyls:
closes https://github.com/protofire/omen-exchange/issues/1926

Fixes this issue in a couple of places.

However this needs to be followed up with https://github.com/protofire/omen-exchange/issues/1930


Mi-Lan (approved):
Works good!

gem1no:
nice work, i can't find anything wrong when I test! 🥇

gem1no (approved):

PR #1934, opened 2021-05-25
update web3 connection handling

hexyls:
At the moment we always load up an initial Infura connection even if a known user is revisiting the app, this can cause the UI to jump around a bit as we settle on a final connection.

PR attempts to improve how we manage web3 connections which should also make loading the market list and other parts of the app smoother and more user friendly.

Moved both `cpk` and `balances` into `connectedWeb3` so we have a single context across the app that contains all the user information we need.

Note: Low priority to merge this, can be merged at any time (I know we have some higher priority PRs open that this PR conflicts with). 

How to test:
- [ ] Balances are rendered on first load
- [ ] Mainnet balances are correct
- [ ] xDai tight integration balances are correct
- [ ] xDai balances are correct
- [ ] Can switch networks without error
- [ ] Works without a web3 connection (chrome incognito mode)


pimato:
Tested:
- [ ] Balances are rendered on first load
- [ ] Mainnet balances are correct
- [ ] xDai tight integration balances are correct
- [ ] xDai balances are correct
- [ ] Can switch networks without error
- [ ] Works without a web3 connection (chrome incognito mode)

pimato (changes requested):
It looks like on xDai direct connection the Dai balances has 3 decimals but should have two decimals:
Bildschirmfoto 2021-06-03 um 12 07 47

pimato (approved):
Works great now!

Mi-Lan (commented):
Great job man! This is big! Our ui is much more smoother now and I like how you bundled cpk into connectedWeb3! Tested shit ton today couldn't find any issues! Also there might be some conflicts with my token list pr regarding fetchBalances since I refactored code a little bit. But awesome job!

Mi-Lan (changes requested):
Works good in most cases! I found that when switching addresses it doesn't always update balance right away but you have to refresh the page heres example.

https://user-images.githubusercontent.com/33226956/122422640-97db4480-cf8d-11eb-9f7a-b18f4a159cfc.mov

KadenZipfel:
Testing: 
- [ ] Balances are rendered on first load
- [ ] Mainnet balances are correct
- [ ] xDai tight integration balances are correct
- [ ] xDai balances are correct
- [ ] Can switch networks without error
- [ ] Works without a web3 connection (chrome incognito mode)

app/src/hooks/useTokens.tsx
```diff
@@ -51,25 +53,25 @@ export const useTokens = (
...
-    skip: false,
-    variables: { listId: omenTCRListId.toString() },
-  })
+  const fetchTokenDetails = async () => {
+    if (context) {
+      const omenTCRListId = getOmenTCRListId(context.networkId)
+      const { httpUri } = getGraphUris(context.networkId)
+      const variables = { listId: omenTCRListId.toString() }
+      const response = await axios.post(httpUri, { query, variables })
+      const data: GraphResponse = response.data.data
```
KadenZipfel, line 71: What's the reasoning for switching from `useQuery` to `axios`?

KadenZipfel (approved):
Big improvement for both the UI and codebase! 🚀

app/src/hooks/useTokens.tsx
hexyls: This was an annoying dependency tree issue, `useQuery` requires [`ApolloProviderWrapper`](https://github.com/protofire/omen-exchange/blob/master/app/src/App.tsx#L47) which requires [`ConnectedWeb3`](https://github.com/protofire/omen-exchange/blob/master/app/src/contexts/Apollo.tsx#L8) and because we are now using `useTokens` inside `ConnectedWeb3` we don't have access to the `ApolloProviderWrapper` yet. I attempted to swap the providers around but just got the same issue in reverse lol

hexyls: Another issue was that hooks can't be used conditionally which was needed here if the connection context had not been defined yet.

hexyls:
@Mi-Lan I think balance updates now resolved by handling potential race conditions, let me know if you run into the issue again 👍

@KadenZipfel Changes I made today were larger than expected (added [multicall](https://github.com/makerdao/multicall.js) for fast initial load) so re-requested a review, cheers man

pimato:
Another round:
- [ ] Balances are rendered on first load
- [ ] Mainnet balances are correct
- [ ] xDai tight integration balances are correct
- [ ] xDai balances are correct
- [ ] Can switch networks without error
- [ ] Works without a web3 connection (chrome incognito mode)

Mi-Lan (approved):
GJ @hexyls Works good! 
- [x] Balances are rendered on first load
- [x] Mainnet balances are correct
- [ ] xDai tight integration balances are correct
- [x] xDai balances are correct
- [ ] Can switch networks without error
- [ ] Works without a web3 connection (chrome incognito mode)

pimato (approved):

PR #1938, opened 2021-05-26
I added/appended the /buy/sell/verify/pool/history on the links and made the routing function when a user copy the link.

### LINKED ISSUE #1415: Include buy/sell/pool/verify in the URL using history.replace()
Previously, I didn't think it was possible to include the market view in the URL while maintaining the proper history, but now I realize this is possible using `history.replace()`. When users visit one of the market tabs it should be appended in the URL, then when someone enters that URL it should automatically go to the proper tab.

Frankaus:
closes #1415


gem1no (changes requested):
see comment
also, normally we don't request review until a PR is 100% ready. 

thanks for giving me the opportunity to check out your work!

pimato (approved):
works great!

app/src/components/market/sections/market_buy/market_buy.tsx
```diff
@@ -56,6 +56,8 @@ import { TransactionDetailsLine } from '../../common/transaction_details_line'
 import { TransactionDetailsRow, ValueStates } from '../../common/transaction_details_row'
 import { WarningMessage } from '../../common/warning_message'
 
+/* eslint-disable no-console */
+
```
Mi-Lan, line 60: this should be deleted

app/src/components/market/sections/market_buy/market_buy_container.tsx
```diff
@@ -1,4 +1,6 @@
-import React from 'react'
+/* eslint-disable no-console */
```
Mi-Lan, line 1: this also should be deleted

app/src/components/market/sections/market_sell/market_sell.tsx
```diff
@@ -473,7 +475,13 @@ const MarketSellWrapper: React.FC<Props> = (props: Props) => {
         />
       )}
       <StyledButtonContainer borderTop={true} marginTop={isNegativeAmountShares}>
-        <Button buttonType={ButtonType.secondaryLine} onClick={() => switchMarketTab(MarketDetailsTab.swap)}>
+        <Button
+          buttonType={ButtonType.secondaryLine}
+          onClick={() => {
+            switchMarketTab(MarketDetailsTab.swap)
+            history.replace(`/${marketMakerAddress}`)
```
Mi-Lan, line 482: I don't think you need this history.replace on cancel buttons I tried it out and they seem to update accordingly without them probably because of those useEffect hooks with tab as trigger

app/src/components/market/sections/market_pooling/market_pool_liquidity_container.tsx
```diff
@@ -14,10 +15,19 @@ interface Props {
...
-  if (isScalar) return <ScalarMarketPoolLiquidity {...props} />
-  return <MarketPoolLiquidity {...props} />
+  const history = useHistory()
+
+  useEffect(() => {
+    history.replace(`/${address}/pool`)
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
+
+  return <>{isScalar ? <ScalarMarketPoolLiquidity {...props} /> : <MarketPoolLiquidity {...props} />}</>
```
Mi-Lan, line 30: I'm wondering if every instance of this could have been done more DRY. For example in the top component where we change currentTab and where we setCurrentTab so that we can maybe use useEffect on tab change and change url based on it? Have you tried that approach? That way we can also maybe avoid hardcoding everything and it would be easier to maintain?

Mi-Lan (changes requested):
Hey @Frankaus Good job man! This works as expected haven't found any issue with current implementation. This is really good for your first pr 💪  I just added some comments on alternative approach that this could have been done. Maybe I'm missing something but take a look.

app/src/components/market/sections/market_pooling/market_pool_liquidity_container.tsx
Frankaus: Nope I didn't, I will try! Thank you

app/src/components/market/sections/market_sell/market_sell.tsx
Mi-Lan: > I don't think you need this history.replace on cancel buttons I tried it out and they seem to update accordingly without them probably because of those useEffect hooks with tab as trigger

So I tested it again and this is not needed still...I explained in above comment why as well..

Mi-Lan (changes requested):
https://github.com/protofire/omen-exchange/pull/1938#discussion_r645063206
This comment was not addressed....So no need for history.replace on cancel buttons...

Frankaus:
Hey guys @Mi-Lan @KadenZipfel @gem1no can someone of you please delete the history.replace() on the Cancel buttons of both the market_buy and market_sell file pls? I deleted the repo from my computer and re-cloned it again but again that code is still not figuring on the version I have... I've tried everything, pulling and restoring but nothing, it always tells me 'up to date'. I can't move on further with that PR sorry.  As per that commit https://github.com/protofire/omen-exchange/pull/1938/commits/61c36a785df21b6b6d9cf104909501548314169f I did and pushed it.
Screenshot 2021-06-12 at 16 18 49

Mi-Lan (approved):
Nicely done!!

app/src/components/market/sections/market_buy/scalar_market_buy.tsx
```diff
@@ -510,7 +513,13 @@ export const ScalarMarketBuy = (props: Props) => {
         />
       )}
       <StyledButtonContainer>
-        <Button buttonType={ButtonType.secondaryLine} onClick={() => switchMarketTab(MarketDetailsTab.swap)}>
+        <Button
+          buttonType={ButtonType.secondaryLine}
+          onClick={() => {
+            switchMarketTab(MarketDetailsTab.swap)
+            history.replace(`/${marketMakerAddress}`)
```
Mi-Lan, line 520: this one slipped up....

Issue #1937, opened 2021-05-26
Increase default gas price from 1 gwei to 5 gwei for the xDai relayer


pimato:
After testing @Mi-Lan developed bridge and the Airdrop PR it seems that the relayer is relaying transaction with a fixed gas price of 1 gwei. the best would be to use a gasEstimation function to dynamically adjust the required gasPrice if this is not possible we should go with hardcoded 5gwei to make sure users are not waiting for a transaction.

PR #1941, opened 2021-05-27
fix/1937

hexyls:
closes https://github.com/protofire/omen-exchange/issues/1937

The [Omen Relay](https://github.com/hexyls/omen-relay) now includes dynamic gas pricing using the [xDai gas price oracle](https://www.xdaichain.com/for-developers/developer-resources/gas-price-oracle). It is set up to use the average gas price atm but we can easily switch to using the fastest option if need be.

To prevent issues in extreme circumstances where either the xDai network is in super high demand, the gas price oracle malfunctions, or the relay operator is malicious, I have set a `MAX_RELAY_FEE` of 0.1 xDai, anything above that will cause the relay tx to fail. Also imo if our users are paying above 0.1 xDai per tx then we should probably start looking for alternatives to xDai 🤷‍♂️ open to opinions on that though.

How to test: buy/create/fund txs should work using the xdai tight integration


gem1no:
looks good, will thoroughly test tomorrow

pimato (approved):
Works great!

Mi-Lan (approved):
Works!

KadenZipfel (approved):
LGTM!

Issue #1939, opened 2021-05-26
Deposit pool tokens to staking contract on market creation


KadenZipfel:
Staking contract is now being created along with market creation in https://github.com/protofire/omen-exchange/pull/1851, but we should also be depositing pool tokens into the staking contract in the `cpk.createMarket` and `cpk.createScalarMarket` transactions

PR #1907, opened 2021-05-11
feature/1861

hexyls:
closes https://github.com/protofire/omen-exchange/issues/1861

Keep in mind that the currently deployed airdrops on xdai and rinkeby are obviously test airdrops that award random test tokens, we will need to redeploy the airdrop contracts when the OMN token has been deployed.

If the address you are using to test did not receive an airdrop, you can grab an `airdropAddress` from the networks.ts file, e.g. on xdai the `airdropAddress` is `0x00aC027Def8227b60866b41e48Da30d23b656138`,  then find the associated airdrop .json file `services/airdrop/0x00aC027Def8227b60866b41e48Da30d23b656138.json` which will list all the addresses that are receiving an airdrop which you can use to test claiming functionality.

As mentioned on keybase we will integrate some of the airdrop ui into the guild modal in a later PR.

Note: If `airdrop` is set in local storage it will hide the airdrop modal, make sure to clear the `airdrop` local storage when testing 🙌

Steps to verify the airdrop json files:

1. Clone https://github.com/hexyls/omen-airdrop
2. `yarn install`
3. `yarn start`
4. The script can take a while to finish, maybe 5-10 mins
5. Verify that the generated mainnetProofs.json and xdaiProofs.json files match the airdrop json files in this PR

PR #1944, opened 2021-05-27
Deposit pool tokens to staking contract on market creation

KadenZipfel:
Closes: https://github.com/protofire/omen-exchange/issues/1939

Note: Please share your transaction hash when testing so I can verify that all looks good

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -55,6 +55,7 @@
     "contract-proxy-kit": "2.2.0-alpha.3",
     "disqus-react": "^1.0.8",
     "eslint-config-react-app": "^5.2.1",
+    "ethereumjs-util": "^7.0.10",
     "ethers": "^4.0.36",
     "git-revision-webpack-plugin": "^3.0.6",
     "graphql": "^14.6.0",
@@ -87,6 +88,7 @@
     "typescript": "^4.0.3",
     "use-debounce": "^3.1.0",
     "use-local-storage-state": "^2.0.0",
+    "web3": "^1.3.6",
     "web3-react": "^5.0.4"
   },
   "scripts": {
--- a/app/package.json
+++ b/app/package.json
@@ -88,7 +88,6 @@
     "typescript": "^4.0.3",
     "use-debounce": "^3.1.0",
     "use-local-storage-state": "^2.0.0",
-    "web3": "^1.3.6",
     "web3-react": "^5.0.4"
   },
   "scripts": {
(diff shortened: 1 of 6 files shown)


pimato:
@KadenZipfel looks like I am getting an error while trying to create a market. Error on console:
Bildschirmfoto 2021-05-31 um 23 25 48

market info:
Bildschirmfoto 2021-05-31 um 23 26 29

hexyls (approved):
Works! Here is my [tx](https://rinkeby.etherscan.io/tx/0x30ef582f600497ce4d692badaabfb1815eb872a5466c1ddb29d6f0dc5cd8a632)

idk if anyone else is going to experience this while testing, but for me both the in-built cpk gas estimator and the metamask gas estimator were both unable to estimate gas properly for this tx (I had to hardcode gas to get it to work). I'm not sure if it is due to complexity of the transaction or what but I will be interested to see if this is a problem for anyone else.

Also just one tip for testing - make sure you are creating markets that resolve at least 1 day from creation.

Mi-Lan:
Screenshot 2021-06-03 at 12 10 05
Im getting this error when trying to create market?

Mi-Lan:
But was able to create it on xDai here is transaction https://blockscout.com/xdai/mainnet/tx/0x6628a9ae384b2a920463b1ed817369c104c857ee92d6f573231caa2cf3d05029/token-transfers

KadenZipfel:
@pimato @Mi-Lan Can you guys try again after disabling cpk failsafe as described here: https://github.com/protofire/omen-exchange/pull/1907#issuecomment-843777240 then share your tx? (Rinkeby)

I think the cpk is falsely assuming that the transaction will fail and thus not even allowing you to proceed

hexyls:
I couldn't quite figure it out today but there could be something odd going on with timestamps causing this error. 

The error we are getting atm seems to be if the current params are executed immediately, but if you submit the same params in a later block it seems like there was never an issue to begin with.

If you make sure that `startingTimestamp` is less than the current block time then you will avoid any errors. I think this allows [`stake`](https://github.com/KadenZipfel/erc20-staking-rewards-distribution-contracts/blob/master/contracts/ERC20StakingRewardsDistribution.sol#L196) to work which uses the [`onlyRunning`](https://github.com/KadenZipfel/erc20-staking-rewards-distribution-contracts/blob/master/contracts/ERC20StakingRewardsDistribution.sol#L471) modifier which requires [`block.timestamp >= startingTimestamp`](https://github.com/KadenZipfel/erc20-staking-rewards-distribution-contracts/blob/master/contracts/ERC20StakingRewardsDistribution.sol#L471)

This is super confusing though because `startingTimestamp` must be [more than](https://github.com/KadenZipfel/erc20-staking-rewards-distribution-contracts/blob/master/contracts/ERC20StakingRewardsDistribution.sol#L106) the timestamp of the block it is included in to be valid) 🤷‍♂️ so really it should revert...

Issue #1961, opened 2021-06-04
Error on withdrawing STAKE token from a market


pimato:
It looks like it is not possible to withdraw Liquidity from this xDai market:
https://xdai.omen.eth.link/#/0x64fab0d2bd7309e381dd0e43ebe90de8edd2dc7f

I deposited 0.999 STAKE into the market and can´t withdraw it now and get the error message:
<img width="2220" alt="Bildschirmfoto 2021-06-04 um 13 21 21" src="https://user-images.githubusercontent.com/5337809/120793910-c857c300-c537-11eb-8eaa-0c99ba8c86b1.png">

```gnosis-conditional-exchange::Market::Fund Error trying to withdraw funds. - Couldn't find token with address '0xb7D311E2Eb55F2f68a9440da38e7989210b9A05e' in network '100'```

we had the same issue here for buying/selling. Did we forgot the fix to be added to deposit/withdraw?
https://github.com/protofire/omen-exchange/issues/1876

PR #1962, opened 2021-06-04
fix/1961

pimato:
I have added STAKE to the known list which fixes the withdraw issue described here: https://github.com/protofire/omen-exchange/issues/1961


hexyls (approved):
🙌

PR #1969, opened 2021-06-08
issue/1956bis

Frankaus:


pimato (approved):

Mi-Lan (approved):
Gj man! Works as it should! I would also add suggestion to add mainnet/xDai picker conditionally since in this case I'm on xDai and it shows when it shouldn't @pimato What do you think?
Screenshot 2021-06-11 at 15 08 54

KadenZipfel (approved):

pimato:
> Gj man! Works as it should! I would also add suggestion to add mainnet/xDai picker conditionally since in this case I'm on xDai and it shows when it shouldn't @pimato What do you think?
> Screenshot 2021-06-11 at 15 08 54

I think we should keep it for now because users might get a link to an xdai market without beeing connected and it is nice to get an indicator that the market they are loading up is actually on xDai.

PR #1990, opened 2021-06-14
fixed by adding fallback default token

### LINKED ISSUE #1981: [ERROR]Cannot destructure property 'address'
Screenshot 2021-06-11 at 17 45 26
Steps to reproduce:
1)Change network to xDai
2)Disconnect
3)Click on address/your connection
4)click on change
5)repeat  2-4 
6)Error will show
- Its a weird and inconsistent one and related to the way I handled currentToken on deposit withdraw modal...

https://user-images.githubusercontent.com/33226956/121715322-02e4d100-cadf-11eb-95f9-ce22f2420bac.mov

Mi-Lan:
closes #1981 
- So as I investigated issue was that when switching between networks so fast useTokens was refetching token list and component was not getting array that It needed to find all necessary token info
- It would be maybe even better to add some loader until data is fetched from useTokens? @pimato


pimato (approved):
fixed the bug which appeared after a user clicked on button `change` in the `Your Connection - Modal`

KadenZipfel (approved):

PR #1994, opened 2021-06-15
Test new fpmm contract

KadenZipfel:
Checkout this branch to test the new fpmm contract on Rinkeby. 

All newly created markets from this branch will use the new fpmm contracts - see https://github.com/KadenZipfel/conditional-tokens-market-makers/pull/3. This change introduces the new fee model, where instead of LPs receiving fees in the collateral token, the fees collect on the outcome tokens held by the AMM.

We should be testing:
- All actions work as expected, e.g. buy, sell, deposit, withdraw, etc.
- LPs receive about as much collateral when they withdraw, given identical circumstances as with old contracts

**Note:** This branch should never be merged. It's for testing purposes only. Reviews are not necessary.


hexyls:
Looks like LP stuff is all pretty much the same which is awesome 🏄‍♂️

For traders it looks like there may be some differences in fees, unsure if it is caused by something in the frontend or the smart contract itself. Can someone please confirm that I tested this correctly (repeat steps on both old and new versions of the fpmm):

Use two accounts, one for LP and one for trading:

1. With account 1, create a new categorical market (A, B) with 1 Dai liquidity 
2. With account 2, buy 10 Dai of A tokens
3. With account 2, sell all A tokens

Old - Account 2 will receive 9.60 Dai (3.97% total fees)
New - Account 2 will receive 9.02 Dai (9.78% total fees)

KadenZipfel:
> Looks like LP stuff is all pretty much the same which is awesome 🏄‍♂️
> 
> For traders it looks like there may be some differences in fees, unsure if it is caused by something in the frontend or the smart contract itself. Can someone please confirm that I tested this correctly (repeat steps on both old and new versions of the fpmm):
> 
> Use two accounts, one for LP and one for trading:
> 
> 1. With account 1, create a new categorical market (A, B) with 1 Dai liquidity
> 2. With account 2, buy 10 Dai of A tokens
> 3. With account 2, sell all A tokens
> 
> Old - Account 2 will receive 9.60 Dai (3.97% total fees)
> New - Account 2 will receive 9.02 Dai (9.78% total fees)

Yep, this is expected behaviour. Since the fees remain in the liquidity pool, the outcome probabilities don't diverge as much, meaning that outcome tokens are generally cheaper. This encourages buying and discourages selling, which imo is ideal. 

There are also some minor differences for LPs, as demonstrated by some initial testing: https://docs.google.com/spreadsheets/d/1YCFHJdgzNbiPXpPLEsGv9858pG3UG-RVTodxr2WE6dA/edit#gid=0
It seems the old model is slightly better for impermanent loss, though the newer model seems to profit more from volume (we can see that liquidity value is higher for the old model when we get closer to the probability extremes, yet higher for the new model when we come back towards the middle)

Mi-Lan:
I couldn't create markets both categorical and scalar? Is that supposed to happen?

KadenZipfel:
> @KadenZipfel I couldn't create markets both categorical and scalar? Is that supposed to happen?

No, market creation should work fine. I'm guessing it's working for me and hexyls because we probably have our modified cpk's to not throw if it expects an error. Not sure why it expects an error though.

hexyls:
> > @KadenZipfel I couldn't create markets both categorical and scalar? Is that supposed to happen?
> 
> No, market creation should work fine. I'm guessing it's working for me and hexyls because we probably have our modified cpk's to not throw if it expects an error. Not sure why it expects an error though.

@Mi-Lan make sure you set market closing dates at least one day into the future 👍

KadenZipfel:
Closing in favor of: https://github.com/protofire/omen-exchange/pull/2075

Issue #1975, opened 2021-06-11
Proxies needs to be upgraded generally, not just for native assets


hexyls:
Missed this when I took away all the hardcoded gas values, proxies need to be upgraded for everything now not just when dealing with native ETH. atm wallets cannot estimate gas properly for cpk transactions if the proxy is not upgraded, causing internal txs to fail 

To reproduce on Rinkeby -
1. Create a new wallet in metamask and fund it with some ETH
2. Go to https://app.compound.finance/ and claim some DAI from the faucet
3. Create a dai categorical market
4. UI will display "Market Not Found We couldn't find the market with the provided address."
5. Wait a bit, check the tx on etherscan, should show the error message: "Although one or more Error Occurred [execution reverted] Contract Execution Completed"
6. Check the tx on tenderly ([example tx](https://dashboard.tenderly.co/tx/rinkeby/0x9abf4c7d053ca772f33a684339a9d3f715e9553d00c686c629dc61f494dd95d8)), error will be `out of gas`

PR #2011, opened 2021-06-21
add proxy upgrade requirement for all markets

hexyls:
closes https://github.com/protofire/omen-exchange/issues/1975

Prompts users to upgrade their proxy regardless of the collateral a market uses (buy/fund/create). This allows correct gas estimations as detailed in the issue.

To test - Follow the steps in https://github.com/protofire/omen-exchange/issues/1975, a market should successfully be created.


pimato:
@hexyls so I just created a completely fresh new wallet on Rinkeby and on the final market creation view I now need to set allowance for Dai and Enable markets? Why do I need to enable those markets for a completely fresh new wallet?

Bildschirmfoto 2021-06-21 um 18 14 55

pimato (changes requested):
see comment

hexyls:
@pimato today I figured out a way to get around the double permission for _new_ users

But there is still one scenario where the double permission will show up - An existing user (proxy already deployed) needs to update their proxy and they are interacting with a market where they have not set allowance for the collateral token yet.

pimato (changes requested):
Set Allowance component has no bottom margin only for ERC-20 tokens:
Bildschirmfoto 2021-06-22 um 19 56 53

pimato:
Also looks like I found a bug:
https://user-images.githubusercontent.com/5337809/122976282-7bc41280-d394-11eb-9ab8-241a3d8c51d4.mov

pimato (approved):
works great!

app/src/components/market/sections/market_buy/market_buy.tsx
```diff
@@ -499,10 +498,11 @@ const MarketBuyWrapper: React.FC<Props> = (props: Props) => {
       )}
       {showUpgrade && (
         <SetAllowance
-          collateral={nativeAsset}
+          collateral={isNative ? displayCollateral : undefined}
```
KadenZipfel, line 501: Wouldn't we want this to always be undefined if this is just for upgrading the proxy and not setting allowance?

app/src/services/cpk.ts
```diff
@@ -1443,16 +1444,23 @@ class CPKService {
...
       if (implementation.toLowerCase() === getTargetSafeImplementation(network.chainId).toLowerCase()) {
         return true
       }
+      return false
     }
-    return false
+    if (isNative) {
+      return false
+    }
+    return true
```
KadenZipfel, line 1463: Why only false if native?

hexyls: edge case galore 😩 if the proxy has not been deployed but the user wants to interact with a native market we must deploy/upgrade first because both mainnet and rinkeby proxy factories don't support native transactions.

If the first market a user wants to interact with is a standard market then that is fine and we can flag it as upgraded.

app/src/components/market/sections/market_buy/market_buy.tsx
hexyls: True, I made this change to display our older messaging just for native markets but should remove to simplify.

Mi-Lan:
I don't know if this is expected behaviour but when I input amount Im able to click create market even though I don't have allowance set...
And also after first transaction it prompts me for another...

https://user-images.githubusercontent.com/33226956/123236950-876f2080-d4dd-11eb-8089-24f52916b1b1.mp4


https://user-images.githubusercontent.com/33226956/123236969-8a6a1100-d4dd-11eb-8cc9-b77e45fe52f8.mp4

KadenZipfel (approved):
LGTM

PR #2020, opened 2021-06-22
Replace ButtonStyles with new color variables and remove primaryAlternative styling 

pimato:


Mi-Lan (approved):
Awesome job! Figma and codebase UNITE!!! Also we need to cleanup those colors we use from buttons and delete them altogether from theme file as you mentioned...This is separate issue thought

Frankaus (approved):
Awesome! I'm moving on with the new theme colours where necessary

PR #2021, opened 2021-06-22
Deleted unused buttons 

Mi-Lan:


KadenZipfel (approved):
LGTM! 

One note for future reference is that as per our [git style guide](https://udacity.github.io/git-styleguide/), we should be including a type to our commit messages

hexyls (commented):
🤙

hexyls (approved):

PR #2022, opened 2021-06-22
Deleted unused components inside components folder

Mi-Lan:
- I just went through all components and found unused files and deleted them
- My thinking was that we are probably never going to reuse that stuff anyways and it would be better to simplify stuff as much as we can
- Theres really a lot of components that are used only once and bunch of components that are very unnecessarily complex sometimes less is more...
- @KadenZipfel @hexyls @pimato What do you guys think?


KadenZipfel:
Sounds like a good plan! We always have the components in our git history so nothing lost

hexyls (approved):
This is awesome, LGTM 🏌️

KadenZipfel (approved):
Awesome! Big clean up 🧼

PR #2024, opened 2021-06-22
[refactor] refactor importing theme with new useTheme hook

Frankaus:
I created a new hook called useTheme to pull the theme palette without having to import theme/index file

We can close and delete that PR. The new PR is #2049

### DIFF
--- a/app/src/components/market/sections/market_list/market_home_container.tsx
+++ b/app/src/components/market/sections/market_list/market_home_container.tsx
@@ -11,8 +11,9 @@ import useLocalStorageState from 'use-local-storage-state'
 import { MAX_MARKET_FEE } from '../../../../common/constants'
 import { useConnectedWeb3Context } from '../../../../hooks'
 import { useMarkets } from '../../../../hooks/useMarkets'
+import { useTheme } from '../../../../hooks/useTheme'
 import { queryCategories } from '../../../../queries/markets_home'
-import theme from '../../../../theme'
+// import theme from '../../../../theme'
 import { getArbitratorsByNetwork, getOutcomes, networkIds } from '../../../../util/networks'
 import { RemoteData } from '../../../../util/remote_data'
 import {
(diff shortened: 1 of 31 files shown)


app/src/components/market/sections/market_list/market_home_container.tsx
```diff
@@ -110,6 +111,7 @@ const wrangleResponse = (data: GraphMarketMakerDataItem[], networkId: number): M
 
 const MarketHomeContainer: React.FC = () => {
   const context = useConnectedWeb3Context()
+  const theme = useTheme()
```
Mi-Lan, line 114: Yeah this hook seems to do the job but I still think we should avoid it since we already have one theme context with styled components and we should use that as much as we can. I would suggest try to implement everything inside styled components for example in history chart you can target stroke,tick and those properties if you just make one styled component out of YAxis for example....Or some workaround if possible

Frankaus:
I still have 3 files left to refactor where we still import theme from the file (app/src/theme/index.tsx). These 3 files are not functional components where I can implement a wrapper or the useTheme hook... I searched for another logic but couldn't find any and so I am stuck. Would appreciate some help on Monday pls.

The remaining files are:
- app/src/components/common/icons/IconSettings.tsx
- app/src/theme/global_style.ts
- app/src/theme/utils.ts

Mi-Lan (changes requested):
Seems that this pr includes a lot of changes that are not related to it...I couldn't review it sufficiently here are few tips:
1) Make sure that pr always includes just changes that are related to issue that you are solving and nothing else I wrote in more detail on keybase here link keybase://chat/dx_dao.omen#general/12115
2) For the IconSettings there are two ways to go about it either make styled.svg and target path->fill property or just hardcode it into the fill because we hardcode  it for most of svgs anyways
3) So seems that only use of outcome/colors is with that getOutcomeColor function. Possible solution is to just copy all outcome/colors into the theme/utils file and delete them from the theme. Though not the best solution I think it will work for now @KadenZipfel what do you think?

hexyls:
Nice work @Frankaus 💪 

Potentially you could modify `getOutcomeColor` to accept the theme as an extra parameter to e.g. `getOutcomeColor(theme, index)`

Note: When I tested out this branch `useTheme` was not defined, likely you created a new file so make sure git is tracking it 👍

Frankaus:
> Nice work @Frankaus 💪
> 
> Potentially you could modify `getOutcomeColor` to accept the theme as an extra parameter to e.g. `getOutcomeColor(theme, index)`
> 
> Note: When I tested out this branch `useTheme` was not defined, likely you created a new file so make sure git is tracking it 👍

Thank you @hexyls. Good spot for useTheme, it got lost in the messy commits... I am fixing everything now

Frankaus:
We can close and delete that PR. The new PR is https://github.com/protofire/omen-exchange/pull/2049

Issue #2013, opened 2021-06-21
[REFACTOR ]Categorisation of tools file along with Tests for all of them


Mi-Lan:
- As discussed in Omen refactor meeting tools file is very unordered and should be categorised further and some tests are missing
1) Categorize tools file 
2) Make sure all functions have tests

PR #2038, opened 2021-06-24
Refactoring Tools Folder

### LINKED ISSUE #2014: [REFACTOR] Replace functions related to conditional tokens with Polymarket ones
As @KadenZipfel suggested in refactor meeting as well as in this doc https://hackmd.io/Ql5odKsvQCKscLQScG4A6w
- We should replace our functions related to conditional token framework with poly market ones from here https://github.com/Polymarket/amm-maths

Mi-Lan:
closes #2014 #2013
Summary: restructured current tools file to be more organised and not clumped up together and added new tests from the polymarket library

- I would love to get your guy's feedback @hexyls @KadenZipfel if I could have organised it better
- Heres things that I think could be looked more for improvement
1) Not sure If I could have categorized rest of categories like string_manipulation, web3, other,utils bettter? Also all tests for those functions are clumped together inside tools.spec.ts while the functions for the fpmm ones are inside the each appropriate fpmm folder
Screenshot 2021-06-25 at 17 56 18
2) One more consideration not sure if its optimal to have one file for each function in my opinion since it uses a lot of space and it has to be exported individually which can be forgotten I would suggest to group all those individual functions together and to export them together. 
fpmm/liquidity is handled each function separate file and I made fpmm/trade to be handled the other way where they are bundled into one file....What do you guys think?
Screenshot 2021-06-25 at 17 58 28


app/src/util/tools/fpmm/trading/tests/calcBuyAmountInShares.test.ts
```diff
@@ -0,0 +1,36 @@
...
+  //not a proper fix but couldn't figure out how to make them work either way this function is not used..
+  // describe('when no holdings', () => {
+  //   it('returns 0', () =>
+  //     expect(calcBuyAmountInShares(bigNumberify(10), 0, [0, 0].map(bigNumberify), 0.1)).toStrictEqual(Zero))
+  // })
+  //
+  // describe('when no funding', () => {
+  //   it('returns 0', () =>
+  //     expect(calcBuyAmountInShares(bigNumberify(0), 0, [100, 100].map(bigNumberify), 0.1)).toStrictEqual(Zero))
+  // })
```
KadenZipfel, line 35: Would be good to get this figured out as I think we should be using this function, but would be fine to move it into its own issue

KadenZipfel (commented):
Super refreshing to have a cleaner directory structure 🙏

1. I think this is good as is
2. Personally I think the approach taken for `/trading` is a bit better, but I could go either way

app/src/util/tools/other.ts
```diff
@@ -0,0 +1,223 @@
...
+  RINKEBY_NETWORKS,
+  SOKOL_NETWORKS,
+  XDAI_NETWORKS,
+} from '../../common/constants'
+import { MarketTokenPair } from '../../hooks/useGraphMarketsFromQuestion'
+import { CompoundService } from '../../services'
+import { getNativeAsset, getWrapToken, networkIds } from '../networks'
+import { BalanceItem, Token, TransactionStep } from '../types'
+
+export const checkRpcStatus = async (customUrl: string, setStatus: any, network: any) => {
```
hexyls, line 16: Just an idea that we don't have to include in this PR - The `utils/networks.ts` file was originally setup to contain functions for accessing information about what network the user is on, over time I added utility functions like `getGraphMeta`, `getBySafeTx`,  and `waitForBlockToSync` that don't really fit in there.

We could potentially move those functions into a new `utils/tools/network` or `utils/tools/http` file along with `checkRpcStatus` and `isValidHttpUrl`. Thoughts?

app/src/util/tools/utils.ts
```diff
@@ -0,0 +1,38 @@
+import { Zero } from 'ethers/constants'
```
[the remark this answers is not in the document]
Mi-Lan: good suggestion! Will change it!

app/src/util/tools/fpmm/trading/tests/calcBuyAmountInShares.test.ts
Mi-Lan: Created issue! https://github.com/protofire/omen-exchange/issues/2042

app/src/util/tools/other.ts
Mi-Lan: I like the idea! I also think the base util folder could be sorted more! Will implement it!

app/src/util/tools/index.ts
```diff
@@ -0,0 +1,20 @@
...
+  calcPoolTokens,
+  calcAddFundingSendAmounts,
+  calcDepositedTokens,
+  calcRemoveFundingSendAmounts,
+  calcDistributionHint,
+  calcInitialFundingSendAmounts,
+  calcAddFundingDepositedAmounts,
+  calcInitialFundingDepositedAmounts,
+  calcSharesBought,
+} from './fpmm/liquidity'
```
KadenZipfel, line 11: Can we switch these up to `export * from...` now to make it super clean?

app/src/util/tools/web3.ts
```diff
@@ -0,0 +1,129 @@
...
+import { BigNumber, getAddress, parseUnits } from 'ethers/utils'
+
+import { getLogger } from '../logger'
+import { getContractAddress, getNativeAsset, getToken, getWrapToken } from '../networks'
+import { CompoundEnabledTokenType, CompoundTokenType, Token } from '../types'
+
+import { strip0x } from './string_manipulation'
+const logger = getLogger('Tools')
+
+export const packSignatures = (array: any) => {
```
KadenZipfel, line 11: Can you clean up the spacing between functions in this file?

KadenZipfel (approved):
Some minor stylistic suggestions, but looks 👌

hexyls (approved):
LGTM! 🏄‍♂️

app/src/util/tools/index.ts
Mi-Lan: ohh yeah good catch dunoo how that slipped!

Issue #1997, opened 2021-06-16
[REFACTOR] Handling all text styling across codebase


Mi-Lan:
Current situation:
- When we have some text field we usually just create new styled component, add appropriate colors from theme, certain font-weight and font-size....
- This sometimes creates a lot of unnecessary styled components which just clutter up top portion of component

Solution:
- In my two weeks i worked with swapr I really liked this pattern and it made development so much easier and code much more readable. So uniswap has about 13 variations of text which in most cases have just 2 or 3 properties defined(font-size, font-weight and is some cases color but it can be ovewritten) and rest of text variations can be added with inline styles.... 
- how it looks like :
![Screenshot 2021-06-16 at 12 43 08](https://user-images.githubusercontent.com/33226956/122205326-69cc0680-cea0-11eb-8bcd-21290074e9f7.png)

Example of it in code:
![Screenshot 2021-06-16 at 12 45 49](https://user-images.githubusercontent.com/33226956/122205647-c92a1680-cea0-11eb-81df-5829627f5950.png)

This would require good alignment with figma so we keep it consistent. This could make development process much faster! If we have types for every text field and reinforce it with code reviews. What do you guys think @KadenZipfel @hexyls @pimato

KadenZipfel:
This is a good call, and I think the most important part is that we have a consistent strategy. Currently, we're doing this for some components and not for others. For example, `button_styling_types` is a well defined set of buttons that are easily reusable and we should use this as a reference for other similar reusable components.

Mi-Lan:
> This is a good call, and I think the most important part is that we have a consistent strategy. Currently, we're doing this for some components and not for others. For example, `button_styling_types` is a well defined set of buttons that are easily reusable and we should use this as a reference for other similar reusable components.

Couldn't agree more ! Part of this refactor can be just using reusing stuff that already in there and cleaning it up a bit so its simpler to reuse. But in my opinion we need to rework that buttons folder wayy to many options that I think none of us know how to use properly...

KadenZipfel:
> > This is a good call, and I think the most important part is that we have a consistent strategy. Currently, we're doing this for some components and not for others. For example, `button_styling_types` is a well defined set of buttons that are easily reusable and we should use this as a reference for other similar reusable components.
> 
[...]

PR #2041, opened 2021-06-28
Refactor: Feature/1997 - handling text styling

Frankaus:
closes https://github.com/protofire/omen-exchange/issues/1997

Refactor: We redefine 5 main text styles that will be implemented through a textWrapper

Figma design specification:
https://www.figma.com/file/zVniQrYgkj8AfQs5zNeBrS/Components-FINAL?node-id=39%3A22

MONDAY 05/07 UPDATE

This PR is deprecated because of the multiple wrong commits issue and has been replaced by https://github.com/protofire/omen-exchange/pull/2078

I will close this PR when the other has been reviewed


app/src/theme/index.tsx
```diff
@@ -524,3 +524,86 @@ export function ThemeProvider({ children }: { children: React.ReactNode }) {
...
+      <TextWrapper
+        align={'left'}
+        color={'text3'}
+        fontFamily={'Roboto'}
+        fontSize={'14px'}
+        fontWeight={400}
+        letterSpacing={'0.2px'}
+        lineHeight={'100% || 16.41px'}
+        verticalAlign={'top'}
+        {...props}
```
Mi-Lan, line 605: @Frankaus All in all this how it should be implemented. But one important thing to keep in mind which is crucial is that TextWrapper should be like base for all the shared styling that all text should include as you see you are repeating couple of properties on all of them like font-family, align,letter-spacing...Always keep DRY(dont repeat yourself) principle in mind one of the most important things in programming...

Mi-Lan:
Can you go over figma properties again...Seems that you have included some properties that are not mentioned  anywhere in figma like verticalAlign,align, paragraphSpacing...And would be good to check for example what are deafult properties so we can use as little properties as we can.
- I think looking at this part is proabaly best idea and not the top part for copying styles...Also for lineHeight I would just copy as it is in figma not sure why did you put 100%/16.41px?
Screenshot 2021-06-29 at 12 17 47

Frankaus:
> Can you go over figma properties again...Seems that you have included some properties that are not mentioned anywhere in figma like verticalAlign,align, paragraphSpacing...And would be good to check for example what are deafult properties so we can use as little properties as we can.
> 
> * I think looking at this part is proabaly best idea and not the top part for copying styles...Also for lineHeight I would just copy as it is in figma not sure why did you put 100%/16.41px?
>   Screenshot 2021-06-29 at 12 17 47

Screenshot 2021-06-29 at 14 01 05

Screenshot 2021-06-29 at 14 02 24

Right... I used the values from there (see screenshots above), I didn't know that extra css view from figma, I will edit it, thank you:)

- included some properties that are not mentioned: in the box of values I used they were... sorry.
-  100%/16.41px: lineHeight was mentioned 2 times so I figured it would be best to include both values and since one is fixed and the other relative I would add the relative first in case there is a wrapper and if not then the fixed one.
- 16.41px: it also looked weird to me but since it was the most precise value between 16px and 16.41px I guessed we should use the most precise one.

I edit everything and push back

app/src/components/market/common/history_chart/index.tsx
```diff
@@ -83,6 +82,12 @@ const AnEvenSmallerLittleBall = styled(OutcomeItemLittleBallOfJoyAndDifferentCol
   width: 8px;
 `
 
+const AxisWrapper = styled.div`
+  display: inline;
+  stroke: ${props => props.theme.colors.verticalDivider};
+  fill: ${props => props.theme.colors.textColor};
```
KadenZipfel, line 88: Looks like this still has old color values

KadenZipfel (changes requested):
Well implemented but lots of merge issues need to be fixed up 👨‍🔧

app/src/components/market/common/history_chart/index.tsx
Frankaus: Yes, although the fact that this file being here is wrong (cause of my commits issues), shall we agree that the color values update will happen on a later single issue @Mi-Lan @pimato ?

app/src/components/market/common/set_allowance/index.tsx
```diff
@@ -45,11 +45,13 @@ export const SetAllowance: React.FC<SetAllowanceProps> = (props: SetAllowancePro
...
-      <Title>Set Allowance</Title>
+      <Title>{collateral ? 'Set Allowance' : 'Enable Markets'}</Title>
       <DescriptionWrapper>
         <Description>
-          This permission allows the smart contracts to interact with your {collateral && collateral.symbol}. This has
-          to be done for each new token.
+          {collateral
+            ? `This permission allows the smart contracts to interact with your ${collateral &&
+                collateral.symbol}. This has to be done for each new token.`
+            : `This permission allows you to interact with market contracts.`}
```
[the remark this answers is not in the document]
Frankaus: Yes that PR went from 1 file changed to 26 in a single commit because I kept the habit of checking out branches by passing from master and pulling all the changes. When working here on the refactoring issues I must not do it

PR #2061, opened 2021-06-30
Feature/setting modal improvments

Mi-Lan:
This pr addresses issue where rpc node is down and displays appropriate settings modal where you can choose different rpc url, here are steps to test it out:

1) Go to src/util/networks.ts
2) If you want to test the default url breaking go to networks object and fuckup the url and first alternative url for appropriate network you are on
Screenshot 2021-07-06 at 13 53 11
3) If you wan't to test the alternative url thats already set in localStorage, inspect in chrome go to Application tab and then to localStorage there you will find rpcAddress and just fuck up the url and refresh the page 
Screenshot 2021-07-06 at 13 55 33

Check list
- [ ] Changing of rpc url in your connection modal displays urls for appropriate network user is on and changes them appropriately
- [ ] When rpc url is broken modal should popup with option to select alternative url for the network user is on and when changed to working rpc it should reload page and not show modal anymore. Also If we are on xDai tight/relayer it should detect and show xDai alternative urls


Mi-Lan:
@hexyls Hey men since you are only wizard around I have to ping you for help :D Main issue I have with this pr to have a clean solution is with window object....So basically because we don't have access to web3 instance we have to fetch current network that user is on like this with window.ethererum.chainId.
Screenshot 2021-07-02 at 20 12 08
And that creates issue because in most cases on initial render window.ethereum is null and I can't check right away what network we are on to see if url/rpc is down it takes 3.5seconds for it to update window object/dom. 
Screenshot 2021-07-02 at 20 28 42
I might be overthinking this but it may be that at the time when app is rendered initially ethereum is not yet inserted into window object...Or Im overlooking something? Also not sure if its smart to use useInterval like I did to call checkRpc every 5 seconds?

hexyls:
@Mi-Lan ayyy man so from playing around with this on my end it looks like the `ethereum` object gets injected pretty quickly but `chainId` is not available right away, you can maybe try this useEffect that fires only once and then makes a call to the provider and see if that works:

```
useEffect(() => {
    const handleRpcStatus = async () => {
      if (ethereum) {
        const chainId = await ethereum.request({ method: 'eth_chainId' })
        // TODO: rest of the code here
      }
    }
    handleRpcStatus()
  }, [])
```

let me know if that doesn't work 👍

Mi-Lan:
> @Mi-Lan ayyy man so from playing around with this on my end it looks like the `ethereum` object gets injected pretty quickly but `chainId` is not available right away (is this the same for you? or is `ethereum` literally null on initial render?), you can maybe try this useEffect that fires only once and then makes a call to the provider and see if that works:
> 
> ```
> useEffect(() => {
>     const handleRpcStatus = async () => {
>       if (ethereum) {
>         const chainId = await ethereum.request({ method: 'eth_chainId' })
>         // TODO: rest of the code here
>       }
>     }
>     handleRpcStatus()
>   }, [])
> ```
> 
> let me know if that doesn't work 👍

Wow man! Thank you that fixed it nicely!
HexylTheWizard
Heres one pic that explains it all xD Disregard my shitty photoshop skills

pimato:
Check list
- [ ] Changing of rpc url in your connection modal displays urls for appropriate network user is on and changes them appropriately
- [ ] When rpc url is broken modal should popup with option to select alternative url for the network user is on and when changed to working rpc it should reload page and not show modal anymore. Also If we are on xDai tight/relayer it should detect and show xDai alternative urls

pimato (approved):
Works great!

hexyls (approved):
@Mi-Lan lmao at that meme 😂

Made a small extra commit to restore the default Infura provider 👍

- [x] Changing of rpc url in your connection modal displays urls for appropriate network user is on and changes them appropriately
- [x] When rpc url is broken modal should popup with option to select alternative url for the network user is on and when changed to working rpc it should reload page and not show modal anymore. Also If we are on xDai tight/relayer it should detect and show xDai alternative urls

Issue #1996, opened 2021-06-16
[REFACTOR] Theme color structure and simplification


Mi-Lan:
Current situation:
 - Looking at our current implementation of colors in the theme file we have lots of colors repeated and segmented according to components that they are used and just general colors and in most cases we don't end up using it as it was supposed to and if we plan to make a different color theme in the future it wouldn't work so well
- Also there is not a great alignment and consistency with figma in terms of reusability and colors naming which makes life harder
![Screenshot 2021-06-16 at 12 22 43](https://user-images.githubusercontent.com/33226956/122202674-8f0b4580-ce9d-11eb-9f54-5ca2a47eb9e0.png)

Proposed solution:
- Looking at the way uniswap has solved this problem with just having one representation of color and reusing that in the components seems to be the way to go. Also I can't confirm this but I bet that if they have figma all the color naming is most likely consistent with the codebase so it much easier to find colors needed and reuse it. 
- Here is how their implementation looks like. Also Implementing different theme would be much clearer.
![Screenshot 2021-06-16 at 12 23 16](https://user-images.githubusercontent.com/33226956/122202759-a2b6ac00-ce9d-11eb-992e-45d722476334.png)
- This would also entail some work in figma as well....What do you guys think? @pimato @hexyls @KadenZipfel 
- Here's link to uniswap codebase where this can be found https://github.com/Uniswap/uniswap-interface/blob/main/src/theme/index.tsx

PR #2062, opened 2021-06-30
Feature/1996bis - color theme import refactor 

Frankaus:
Closes:#1996

Refactor: Fix/Refactor every theme import on the repo for the colours. We now import the new theme colour palette through props, props in a wrapper or useTheme hook.


app/src/components/common/icons/IconClose.tsx
```diff
@@ -23,10 +23,11 @@ interface Props {
 }
 
 export const IconClose = (props: Props) => {
-  const { color = '#37474F', hoverEffect = false, onClick, size = '24' } = props
+  const { color = '#37474F', hoverEffect = true, onClick, size = '24' } = props
 
   return (
     <Wrapper
+      color={'inherit'}
```
Mi-Lan, line 30: No need for color here....

app/src/components/common/icons/IconClose.tsx
```diff
@@ -38,7 +39,7 @@ export const IconClose = (props: Props) => {
       <path
         className="path"
         d="M19 6.41L17.59 5L12 10.59L6.41 5L5 6.41L10.59 12L5 17.59L6.41 19L12 13.41L17.59 19L19 17.59L13.41 12L19 6.41Z"
-        fill={hoverEffect ? '#DCDFF2' : color}
```
Mi-Lan, line 41: Why was this changed?

app/src/components/market/common/history_chart/index.tsx
```diff
@@ -29,7 +28,7 @@ const ChartTooltip = styled.div`
 `
 
 const TooltipTitle = styled.h4`
-  color: ${props => props.theme.colors.textColorDark};
+  color: ${props => props.theme.text1};
```
Mi-Lan, line 31: Man we went through this on call this morning and I left feedback on last pr. Not sure why is this still there?

app/src/components/market/common/xdai_bridge_modal/bridge_transaction_state.tsx
```diff
@@ -31,7 +30,7 @@ const SvgWrapper = styled.div`
   margin-bottom: 25px;
 `
 const BoldedText = styled.div`
-  color: ${({ theme }) => theme.colors.textColorDark};
+  color: ${props => props.theme.text1};
```
Mi-Lan, line 33: here the same

app/src/components/market/sections/market_list/market_home_container.tsx
```diff
@@ -63,17 +62,20 @@ const TextTitle = styled.div`
...
   text-decoration: underline;
 
   &:hover {
-    color: ${props => props.theme.buttonSecondary.color};
+    color: ${props => props.theme.primary3};
   }
 `
+const IconWrapper = styled(IconClose as any)`
+  color: ${props => props.theme.text1};
+`
```
Mi-Lan, line 78: So when you make it like this styled(IconClose) that is basically the same as IconClose component and you don't need wrapper for it but you can just use the wrapper you created and it shouldn't be named wrapper. Please spend some time going thought our codebase on how we do this pattern we use it a lot throughout the code and you should understand it....

app/src/components/market/sections/market_list/market_home_container.tsx
```diff
@@ -408,7 +411,9 @@ const MarketHomeContainer: React.FC = () => {
               setHasSeenBanner('true')
             }}
           >
-            <IconClose color={theme.colors.tertiary} hoverEffect size={'24'} />
+            <IconWrapper>
+              <IconClose hoverEffect size={'24'} {...restProps} />
```
Mi-Lan, line 415: you shouldn't add restProps like this but inside the IconClose please look at the rest of icons there are couple of examples how we do it...

Mi-Lan (changes requested):
In general I think you need to spend more time reading and thinking about feedback we give you so we don't repeat stuff I think we went in great detail on the call and in private chat about this. As we mentioned on the call this morning just focus on delivering good stuff and don't worry how long it will take you.

app/src/components/market/common/history_chart/index.tsx
gem1no: if you were less aggressive in criticism it might make for a more positive work environment, which may lead to less stressed workers and less mistakes. this is not how you treat someone who is new to a code base.

Mi-Lan: @gem1no I try to be patient and polite but Im not perfect I have certain thresholds. You may be right Im a little rough around the edges sometimes. But at least Im honest and you can always know where you stand with me and I speak from my heart I don't lie or try to deceive. Looking the way you left and how you are talking shit now that you are not here tells me a lot about you and that it maybe was a good decision we made since you clearly weren't honest with all of us and that is coming out now. In my life I have grown the most with people who were honest with me then with people who were superficially polite, not everyone appreciates that. https://media.giphy.com/media/qEfGLWRq9LokP6cBeX/giphy.gif

Frankaus: @gem1no no. @Mi-Lan has been very supportive, helpful, and honest. We have had quite some misunderstanding on this issue and that's 95% my fault. I am starting here and I have not taken the right attitude towards timing, quality, and research, I am fully aware and changing this. The remote and independent contribution working environment is very much like a double hedge sword and requires adaptation but mostly learning and honesty.

I have done all my commits in the way that seemed most coherent and in good faith. Though I went too fast and agitated. That will change.

app/src/components/common/icons/IconClose.tsx
Frankaus: Here I added inherit so that the SVG would get the color from [that](https://github.com/protofire/omen-exchange/pull/2062#discussion_r661726708) wrapper. I tested it and it worked well that way. Now I will search for a better way from our codebase

app/src/components/common/icons/IconClose.tsx
Frankaus: Because the hover effect is set higher up on the file, line 7-11

Screenshot 2021-07-01 at 10 07 13

app/src/components/market/common/xdai_bridge_modal/bridge_transaction_state.tsx
Frankaus: Here there is a big misunderstanding, my fault, but that seemed so logical/coherent to me. I believed I had to not only edit the theme import but also edit the colors when needed with the ones @pimato had set finals... I will revert all the colors modification and keep that for a later issue

app/src/components/market/sections/market_list/market_home_container.tsx
Frankaus: Yes, I'm gonna spend the time to look into this because it's the first time I deal with such a technique...

app/src/components/common/icons/IconClose.tsx
Mi-Lan: those two are not the same properties one is for path and other is for fill they affect 2 different parts of svg. Try changing them to different colors to see where they apply....

Mi-Lan: as we discussed on keybase it won't work because there is lot of places where we use hoverEffect and with removing of this it won't apply that color. So we either change it in all places (which is not that dry). So either we leave it as it ....Or there is maybe some other solution that will work even better....

app/src/theme/utils.ts
```diff
@@ -1,11 +1,142 @@
...
+      darker: '#b1a325',
+      medium: '#ffeb3b',
+    },
+    {
+      darker: '#a51546',
+      medium: '#e91e63',
+    },
+  ],
+}
+
```
Mi-Lan, line 137: this should be deleted from the theme file since its only use of it

app/src/components/common/icons/IconClose.tsx
```diff
@@ -23,17 +23,20 @@ interface Props {
...
 export const IconClose = (props: Props) => {
-  const { color = '#37474F', hoverEffect = false, onClick, size = '24' } = props
+  // If we keep the default value here it prevails on the styled(IconClose) and subsequent styled components I guess. I suggest to remove it
+  const { color, hoverEffect = false, onClick, size = '24', ...restProps } = props
 
   return (
     <Wrapper
-      fill="none"
+      // If we keep 'none' in the fill here then only a hardcoded default variables specified above would apply. Instead, having color here makes the styled(IconClose) color apply (without needing the harcoded value). Here the default value for color always prevails on the styled(IconClose), weird
+      fill={color}
```
Mi-Lan, line 32: Can you look maybe more into this seems that we are not passing colors in any of the cases this is probably left from some component we used to have so its maybe fine deleting it from there 
Screenshot 2021-07-05 at 20 07 56
I looked into this a bit and I believe we could move the fill property to styled componets like I did so we don't have to hardcode it...What do you think?

Mi-Lan: and please test it out and check that we are not breaking any icons throughout  the codebase

Mi-Lan (changes requested):
Rest of it good looks to me! Disregard what I mentioned on the call, I thought you deleted this part of the code which is not the case...
Screenshot 2021-07-05 at 20 04 41

app/src/theme/utils.ts
Frankaus: https://github.com/protofire/omen-exchange/pull/2049#discussion_r660508408

Mi-Lan: Yeah I meant from the theme/index maybe I should have been more specific.

app/src/components/common/icons/IconClose.tsx
```diff
@@ -34,11 +34,11 @@ export const IconClose = (props: Props) => {
       viewBox="0 0 24 24"
       width={size}
       xmlns="http://www.w3.org/2000/svg"
+      {...restProps}
     >
       <path
         className="path"
         d="M19 6.41L17.59 5L12 10.59L6.41 5L5 6.41L10.59 12L5 17.59L6.41 19L12 13.41L17.59 19L19 17.59L13.41 12L19 6.41Z"
-        fill={hoverEffect ? '#DCDFF2' : color}
```
Mi-Lan, line 41: Because you have deleted this and just left `fill: ${props => props.theme.colors.textColorDark};` in the wrapper above we get this issue...
Screenshot 2021-07-12 at 18 16 32
Where we get this Dark IconClose.. @Frankaus Did geronimo suggest that change?

Mi-Lan: Please compare master and your pr to make sure all is the same...

Frankaus: Mhhh no Geronimo didn't request any change. My bad yes, it was tertiary! I'll fix that tmrw cause I was today on the review so I don't create more issues with branching etc
Screenshot 2021-07-12 at 18 31 53

app/src/components/common/icons/IconClose.tsx
```diff
@@ -12,18 +12,18 @@ const Wrapper = styled.svg<{ hoverEffect: boolean | undefined }>`
 
   .path {
     transition: 0.2s fill;
+    fill: ${props => props.theme.colors.tertiary};
```
Mi-Lan, line 15: Ughhh man Im gonna kill you when I meet you in france this monday for real! So I tried explaining this 10 times but just change this line of code to this
`fill: ${props => (props.hoverEffect ? props.theme.colors.tertiary : props.theme.colors.textColorDark)};`
Try to understand why is that so or reread my comments and private chat....

Mi-Lan (changes requested):
Just change this line requested and we are gtg!

app/src/components/common/icons/IconClose.tsx
Mi-Lan: and If you compare create market button icon on master and your pr you woud have noticed they are not the same...

Mi-Lan (approved):

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
KadenZipfel, line 1: Looks like this was added by accident, pls delete

app/src/theme/utils.ts
```diff
@@ -1,11 +1,142 @@
 import { getLogger } from '../util/logger'
 
-import theme from './index'
-
 const logger = getLogger('Theme::Utils')
 
+const outcomes = {
```
KadenZipfel, line 5: Looks like the outcome colors aren't getting imported from the right spot now that they've moved
image

KadenZipfel (changes requested):

app/src/theme/utils.ts
```diff
@@ -1,11 +1,142 @@
...
+    },
+  ],
+}
+
 export const getOutcomeColor = (index: number): Record<string, any> => {
-  const localTheme: Record<string, any> = theme
+  const localTheme: Record<string, any> = outcomes
   const defaultReturn = { darker: '#000', medium: '#999' }
 
   if (!localTheme || !localTheme['outcomes'] || !localTheme['outcomes']['colors']) {
```
hexyls, line 142: atm this function is throwing an error because both `localTheme['outcomes']` and `localTheme['outcomes']['colors']` are undefined.

We probably don't even need to check if those have values any more because they are now hardcoded into this file.

Line 47 will throw an error:
```
const colors = localTheme.outcomes.colors
```
Can change the path to:

```
const colors = localTheme.colors
```

👍

app/src/theme/utils.ts
```diff
@@ -1,19 +1,150 @@
...
-  if (!localTheme || !localTheme['outcomes'] || !localTheme['outcomes']['colors']) {
+  if (!localTheme || !localTheme['colors']) {
     logger.error('Error when using theme or theme.outcomes')
     return defaultReturn
   }
 
-  const colors = localTheme.outcomes.colors
+  const colors = localTheme.colors
 
   return colors[index % colors.length]
```
Frankaus, line 149: Fixed. Now outcomes colours are showing properly and no more error message returning. Thanks guys for the feedback:)

Screenshot 2021-07-27 at 10 03 24

pimato (approved):

hexyls (approved):

KadenZipfel (approved):
Looks good!

Mi-Lan (changes requested):
Theme is still being imported in IconSettings
Screenshot 2021-07-29 at 12 03 22
Also the `export default theme` from theme/index file should be deleted
Screenshot 2021-07-29 at 12 05 06

app/src/components/common/icons/IconSettings.tsx
```diff
@@ -1,12 +1,11 @@
 import React from 'react'
+import styled from 'styled-components'
 
-import theme from '../../../theme'
-
+const Path = styled.path`
+  fill: ${props => props.theme.colors.textColorLighter};
+`
```
Mi-Lan, line 6: I don't think this is the best practice to target path like this. Instead it would be better to target whole svg and then target property path inside it and then fill property within it. Like so
const Svg=styled.svg`
	path{
		fill: ${props => props.theme.colors.textColorLighter};
	}
`

Mi-Lan (approved):
GJ!! Lets gooooooooo

Issue #2065, opened 2021-06-30
Can access withdraw tab without any pool tokens


KadenZipfel:
![image](https://user-images.githubusercontent.com/30579067/124007594-5b95e400-d990-11eb-9e63-9dafbc66ddea.png)

PR #2069, opened 2021-07-02
Disable withdraw tab when user has no liquidity

hexyls:
closes https://github.com/protofire/omen-exchange/issues/2065

Testing for both categorical and scalar:

- [ ] If the user has no liquidity when the Deposit tab is selected the Withdraw tab is disabled 
- [ ] If the user has liquidity, the Withdraw tab is selected, and the user withdraws all liquidity, the Withdraw tab is still enabled but switching back to the Deposit tab disables it

### DIFF
--- a/app/src/components/market/sections/market_pooling/market_pool_liquidity.tsx
+++ b/app/src/components/market/sections/market_pooling/market_pool_liquidity.tsx
@@ -33,6 +33,7 @@ import {
   getBaseTokenForCToken,
   getInitialCollateral,
   getSharesInBaseToken,
+  isDust,
 } from '../../../../util/tools'
 import {
   CompoundTokenType,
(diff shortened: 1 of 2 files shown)


KadenZipfel:
Testing:
- [ ] If the user has no liquidity when the Deposit tab is selected the Withdraw tab is disabled 
- [ ] If the user has liquidity, the Withdraw tab is selected, and the user withdraws all liquidity, the Withdraw tab is still enabled but switching back to the Deposit tab disables it

KadenZipfel (approved):
🔥

pimato:
- [ ] If the user has no liquidity when the Deposit tab is selected the Withdraw tab is disabled 
- [ ] If the user has liquidity, the Withdraw tab is selected, and the user withdraws all liquidity, the Withdraw tab is still enabled but switching back to the Deposit tab disables it

pimato (approved):
Great work!

Issue #2037, opened 2021-06-24
Create market button should be disabled until allowance is set


Mi-Lan:
- Create market should be disabled until allowance is Set 
- Below is video demonstrating that you can click on create market while allowance is not set
https://user-images.githubusercontent.com/33226956/123245205-e5ebcd00-d4e4-11eb-9bcf-a6ee955f90fa.mp4

PR #2072, opened 2021-07-02
Disable create market button if allowance is not set

hexyls:
closes https://github.com/protofire/omen-exchange/issues/2037


pimato (approved):
works!

Mi-Lan (approved):
LGTM! GJ!

PR #2079, opened 2021-07-05
Make Omen work with new TheGraph gateway

giacomolicari:
In the view of the migration to the new The Graph Mainnet network (https://thegraph.com/blog/mainnet-migration), it's now required to make HTTP requests to TheGraph go through the new gateway system.

To achieve it, it's necessary to set an API Key and the Subgraph ID as environment variables.

Omen subgraph, initially owned by Gnosis, was one of the subgraphs designated for the first migration phase (https://thegraph.com/blog/mainnet-migration-partners).

### DIFF
--- a/app/src/common/constants.ts
+++ b/app/src/common/constants.ts
@@ -53,7 +53,11 @@ export const TRADING_FEE_OPTIONS = [
   '5.00',
 ]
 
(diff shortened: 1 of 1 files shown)


KadenZipfel:
Dropped in favour of: https://github.com/protofire/omen-exchange/pull/2086

Issue #2000, opened 2021-06-16
[REFACTOR]Move all contexts we use inside contexts folder since we have most of them inside hooks folder


Mi-Lan:
-Some contexts like web3,cpk and balance are inside hook when they should be inside the contexts folder
- In general throughout the code we do a lot of prop drilling (passing data down from component to component) context api is great way around that I think we should use it more then we are...

PR #2084, opened 2021-07-07
Feature/2000 - context refactoring (2 days)

Frankaus:
closes https://github.com/protofire/omen-exchange/issues/2000

Refactor: Move the context provider that are inside the hooks folder in the context folder


app/src/hooks/index.ts
```diff
@@ -1,4 +1,4 @@
-export * from './connectedWeb3'
+export * from '../contexts/connectedWeb3'
```
Mi-Lan, line 1: Since we moved connectedWeb3 into the context it would be better if we exported it from there as well..

Mi-Lan (changes requested):
GJ! I would also add useCpk to context since we kinda bundled it with connectedWeb3 context and it is context in itself. And one more comment below

app/src/hooks/useMarkets.tsx
```diff
@@ -1,6 +1,7 @@
 import { useQuery } from '@apollo/react-hooks'
 import React from 'react'
 
+import { useConnectedWeb3Context } from '../contexts/connectedWeb3'
```
Mi-Lan, line 4: You can import it directly from the contexts folder since you created index file which exports it...Like this
Screenshot 2021-07-16 at 12 06 00

Mi-Lan (changes requested):
Added small comment... @hexyls @KadenZipfel  Wondering if it makes sense to further organize hooks folder? When I made issue I was hoping there would be more  to move 😭

hexyls (approved):
Looks good once you address @Mi-Lan comment 👍

@Mi-Lan hmmm not sure how to further organize the hooks 🤔

KadenZipfel:
> Added small comment... @hexyls @KadenZipfel Wondering if it makes sense to further organize hooks folder? When I made issue I was hoping there would be more to move 😭

Yeah, like we discussed on friday, we can probably categorize them based on uses, e.g. a directory for subgraph hooks

Frankaus:
> > Added small comment... @hexyls @KadenZipfel Wondering if it makes sense to further organize hooks folder? When I made issue I was hoping there would be more to move 😭
> 
> Yeah, like we discussed on friday, we can probably categorize them based on uses, e.g. a directory for subgraph hooks

As we spoke about it, I suggest we have subfolders: 'Graph', 'Cpk' and then the shall I leave the remaining ones it the hook folder or inside an 'Other' subfolder ? @Mi-Lan

hexyls (approved):
Looks good! 

I'm not 100% sure on the uppercase folder names e.g. `MarketData`, I don't think we use that naming convention elsewhere - @KadenZipfel what do you think?

pimato (changes requested):
@Frankaus please rename the folder structure to lowercase.

pimato:
looks like you did not test it:
```Failed to compile.

./src/contexts/connectedWeb3.tsx
Cannot find file: 'useCpk.tsx' does not match the corresponding name on disk: './src/hooks/cpk/Cpk'.```

hexyls (approved):

KadenZipfel (approved):
LGTM!

Mi-Lan (approved):
This looks much better! GJ Franko!

Issue #2088, opened 2021-07-07
Switching networks break the app


pimato:
**Describe the bug**
It looks like when a user switches networks via metamask network picker, the Omen frontend breaks.

**To Reproduce**
1. Go to http://localhost:3000/#/liquidity while being connected to Mainnet
2. Switch Network to xDai
3. BUG: app crashes, if it does not try to repeat switching networks back and forth and wait a little bit after the switch.
4. See error:
<img width="1033" alt="Bildschirmfoto 2021-07-07 um 21 59 40" src="https://user-images.githubusercontent.com/5337809/124821162-a86c5280-df6e-11eb-82e5-60bea8a773c1.png">

**Expected behavior**
Switching networks should not break the UI.

Issue #2083, opened 2021-07-07
Frontend breaks after user tries to change wallet connection


pimato:
**Describe the bug**
So it looks like after merging https://github.com/protofire/omen-exchange/pull/2061 we introduced a bug which appears once a user wants to change the connection. Here is a showcase about it:
![Bildschirmaufnahme 2021-07-07 um 10 49 16](https://user-images.githubusercontent.com/5337809/124731407-7c22e880-df12-11eb-9c04-18bce845697d.gif)

**To Reproduce**
1. Click on the Connection button at the top right corner
2. Click on the change button
3. The error is not appearing straight away, so waiting a couple of seconds will crash the app.

**Expected behavior**
The app should not crash after the user tries to change the connection.

**Desktop (please complete the following information):**
 - OS: Mac OS bigSur 11.2.1 (20D74)
 - Browser Brave
 - Version 1.26.74 Chromium: 91.0.4472.124 (Offizieller Build) (x86_64)

Mi-Lan:
I took a stab at this issue its something related to connectors I couldn't quite figure it out. I need to dig more since this is uncharted territory for me... never worked with this stuff...

PR #2089, opened 2021-07-08
Prevent crashes on connection changes when viewing market list

hexyls:
closes https://github.com/protofire/omen-exchange/issues/2088
closes https://github.com/protofire/omen-exchange/issues/2083

Main issue was stale market data during loading of a new market list, the old market list would be rendered briefly while using the old connection causing the 'contract not deployed' errors. Updated to clear the market list on reload.

Added a check to make sure the app is only rendered if the connection networkId and the CPK networkId are in sync so that we can prevent fetching data we don't want.

Something I missed in code review and I only just learned about: useEffect without a dependency array will run on every render, we had a few of these which led to some infinite loops, removed now.

### DIFF
--- a/app/src/components/market/common/list_item/index.tsx
+++ b/app/src/components/market/common/list_item/index.tsx
@@ -139,7 +139,8 @@ export const ListItem: React.FC<Props> = (props: Props) => {
       setLiquidity(await marketMaker.getTotalSupply())
     }
     marketMaker && getLiquidity()
-  })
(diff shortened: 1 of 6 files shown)


pimato (approved):
can´t replicate the issues anymore! Great work!

KadenZipfel (approved):
Code looks good and I can't reproduce the bugs. GJ!

Mi-Lan (approved):
LGTM!

Issue #2099, opened 2021-07-14
Filter markets by known oracles


hexyls:
We need to filter markets by known oracles to ensure markets can be resolved.

We can do this in a similar way to how we filter by known arbitrators. First we need to [add an `oracle` field to the fpmm object](https://github.com/protofire/omen-subgraph/issues/181) in the subgraph, then we can update the market list query to filter by known oracles.

Users should only see markets with known oracles in the market list when this is complete.

PR #2100, opened 2021-07-14
Filter markets by known oracles

hexyls:
closes https://github.com/protofire/omen-exchange/issues/2099 

Adds a $knownOracles query for the market list, should work when we update the subgraph


Frankaus (approved):
- [x]  On Rinkeby there should only be two markets listed in open markets

Screenshot 2021-08-06 at 12 02 09

KadenZipfel (approved):
- [x] On Rinkeby there should only be two markets listed in open markets

lgtm!

hexyls:
In draft until we can test with production rinkeby/mainnet/xdai subgraphs

PR #2104, opened 2021-07-15
refactor: Move abis from services into the abi folder

### LINKED ISSUE #2016: [REFACTOR]Omen Mobile
-As we discussed in the meeting we will not be doing whole mobile optimisation but just fix places where mobile UI breaks
To do:
 1) Fix mobile Header since it looks horrendous on mobile
Screenshot 2021-06-21 at 21 58 25
@pimato @KadenZipfel feel free to add more issues

### LINKED ISSUE #2063: Separate files for ABIs
In files such as `services/xdai.ts`, the ABIs can take up a fair amount of the file, making it a little bit less nice to work on. Instead we should move all our ABIs into separate files in an `/abis` folder. We should then also minify the files for slightly better load times.

Mi-Lan:
closes #2063 
- Updates abi's so they are all contained inside the abi's folder and are all json files. This pr will requeire extensive testing since it covers all the services we use.
- How to test:
- [ ] Cpk, test that transaction states work as expected and that its in sync 
- [ ] Create market works for both categorical and scalar
- [ ] Buy and sell for categorical and scalar
- [ ] Bonding works
- [ ] Redeeming works
- [ ] Deposit and withdraw for both categorical and scalar
- [ ] Token balances fetch
- [ ] Fetching and approving of allowance works
- [ ] Kleros verification works
- [ ] History section displays good data
- [ ] Relay works
- [ ] xDai tight works as expected
- [ ] First transaction with cpk creates new safe and works
- [ ] wrapped tokens work


Mi-Lan:
@KadenZipfel Hey men! I had couple of dilemmas regarding this how to best proceed. Would greatly appreciate your feedback!
1) Should we separate files in the abi folder by contract? Or by the services? Also some for some of the contracts we use only like 1 function so Im not sure what to do with those abi-s makes more sense to leave them in the file itself... .XDai service uses like 4 different contracts(multiClaim,xDaiBridge,OmniBridge, OmenGuild...)
2) Would it make more sense to make files .json ? And Im also not sure if we should keep all the abis consistent I saw that In couple of places we use these kinda string abi-s
Screenshot 2021-07-15 at 20 20 02
And in other these object like abi-s...Not sure if there is any difference between them.
Screenshot 2021-07-15 at 20 20 36

hexyls:
hmmm tricky, the object abis are more common but the array abi seems a bit easier to read 🤔 

If we decide to go with the more common object abi we can convert to it by grabbing it off the contract:

```
// @ts-expect-error ignore
console.log(JSON.stringify(this.contract.interface.abi.map(({ gas, ...rest }) => rest)))
```

^ can copy off the console

pimato:
Todo:
-Separate ABIs by service (Estimated delivery: 16th July 2021)

pimato:
- @Mi-Lan needs more time to research how to properly minify the ABI file.

Mi-Lan:
@hexyls While going through the abis I noticed in the wrapped_assets file you made separate abi for wETHabi,wxDaiabi and wsPOAabi. But they all share exactly the same functions? I went thought them one by one and they seem to be the same(types,paramters,return values etc...). Is there reason why we have separate abis for 3 of them?

hexyls:
@Mi-Lan mmmm no I don't think there is any good reason, we can just use one 👍

app/src/services/realitio.ts
```diff
@@ -14,43 +16,8 @@ import { Question, QuestionLog, TransactionStep } from '../util/types'
...
-  'function isFinalized(bytes32 question_id) view public returns (bool)',
-  'function resultFor(bytes32 question_id) external view returns (bytes32)',
-  'function submitAnswer(bytes32 question_id, bytes32 answer, uint256 max_previous)',
-  'function withdraw()',
-  'function claimWinnings(bytes32 question_id, bytes32[] history_hashes, address[] addrs, uint256[] bonds, bytes32[] answers)',
-  'function questions(bytes32 question_id) view public returns (bytes32 content_hash, address arbitrator, uint32 opening_ts, uint32 timeout, uint32 finalize_ts, bool is_pending_arbitration, uint256 bounty, bytes32 best_answer, bytes32 history_hash, uint256 bond)',
-  'event LogClaim(bytes32 indexed question_id, address indexed user, uint256 amount)',
-]
-const realitioCallAbi = [
+const askQuestionConstant = [
```
Mi-Lan, line 19: @hexyls @KadenZipfel Hey guys so this is the only hardcoded abi left standing because its the same as the one imported from relitio.json but instead of being payable its constant...And hes not on official realitio abi so in my opinion it makes sense to leave it like this?

KadenZipfel: Seems chill to leave it imo

app/src/services/unwrap_token.ts
```diff
@@ -1,7 +1,7 @@
 import { utils } from 'ethers'
 import { BigNumber } from 'ethers/utils'
 
-import { wETHabi, wsPOAabi, wxDaiabi } from '../abi/wrapped_asset'
+import wrappedAsset from '../abi/wrappedAsset.json'
```
KadenZipfel, line 4: Are they all the same ABI?

app/src/services/unwrap_token.ts
```diff
@@ -14,11 +14,9 @@ class UnwrapTokenService {
     const symbolLowerCase = symbol.toLowerCase()
     switch (symbolLowerCase) {
       case 'wxdai':
-        return wxDaiabi
       case 'wspoa':
-        return wsPOAabi
       case 'weth':
-        return wETHabi
+        return wrappedAsset
```
KadenZipfel, line 19: Doesn't this mean we're not returning anything for wxdai and wsPOAabi?

KadenZipfel (commented):
Gonna come back around to do testing but in general this looks well put together

app/src/services/unwrap_token.ts
Mi-Lan: well those that were deleted are return values not the cases for evaluation. So it will be fine...

app/src/services/unwrap_token.ts
Mi-Lan: Yeah I asked hexyl the same thing and wen't through them they looked same to me...

app/src/services/unwrap_token.ts
KadenZipfel: Why not just remove the switch and return `wrappedAsset` since they're all the same ABI?

KadenZipfel: Oh I guess this is the same thing effectively(?)

KadenZipfel (approved):
Code looks good! Haven't manually tested though

pimato:
- How to test:
- [ ] Cpk, test that transaction states work as expected and that its in sync 
- [ ] Create market works for both categorical and scalar
- [ ] Buy and sell for categorical and scalar
- [ ] Bonding works
- [ ] Redeeming works
- [ ] Deposit and withdraw for both categorical and scalar
- [ ] Token balances fetch
- [ ] Fetching and approving of allowance works
- [ ] Kleros verification works
- [ ] History section displays good data
- [ ] Relay works
- [ ] xDai tight works as expected
- [ ] First transaction with cpk creates new safe and works
- [ ] wrapped tokens work

pimato (changes requested):
```./src/components/market/common_sections/list/list_item/index.tsx
Module not found: Can't resolve '../../../../../services' in '/Users/xxx/DXdao/omen-exchange/app/src/components/market/common_sections/list/list_item'```

I am getting this error on `yarn start`

app/src/services/unwrap_token.ts
Mi-Lan: Hmm yeah I guess maybe I can do it like this? But my only concern was with the default option where it returns empty object? Do you think that can cause issues?
Screenshot 2021-08-17 at 22 29 55

Mi-Lan: This was is before?
Screenshot 2021-08-17 at 22 31 12

Mi-Lan: I just pushed it and tested it. You were right! Its way better like this got ridd of lots of code....

Mi-Lan:
Side note for testing It would be good to test withdraw with xDai,weth and wspoa! I

KadenZipfel (approved):
Changes look good! IMO in terms of testing we should bucket most tests along with other changes before the next release. A lot of merge conflicts remained in some PR's after the refactor migration so I expect there will be a decent amount of bugs. @Mi-Lan, any tests specific to this PR we should do that might otherwise get missed in release testing?

pimato:
It seems like depositing via xDai tight integration does not work:
https://user-images.githubusercontent.com/5337809/130600569-91428201-85c3-4473-8e6d-b15bccdaa9f9.mov

pimato (approved):

PR #2110, opened 2021-07-16
refactor: In progress, moved shared logic into market_buy_container

Mi-Lan:
In this pr I combine shared logic from marketBuy and scalarMakerBuy inside marketBuyContianer
How to test:
- [ ] Make sure buy works both for scalar and categorical markets


pimato (approved):
Tested:
- [x] Make sure buy works both for scalar and categorical markets

Issue #2029, opened 2021-06-23
Improve transaction state management (1-2 days)


pimato:
Requirements:
- Transaction modal should be high up in the component tree so it doesn’t always have to be imported into components (if possible) (Use contextAPI)
- `setTxHash` and `setTxState` should be included directly in the cpk constructor (as optional parameters)

pimato:
Estimate duration: 1-2 days approx.

PR #2111, opened 2021-07-19
Improve transaction state management

KadenZipfel:
Closes: https://github.com/protofire/omen-exchange/issues/2029

For basically every type of transaction we want to confirm:
- [ ] 'Transaction submitted/confirmed/etc.' message in transaction modal works as expected
- [ ] Etherscan button works as expected

Functions that go through the cpk should all work as expected, so extra testing focus should be on those that do not.


hexyls (approved):
works well for everything I tested 👍 this is a great change, will be way less effort when we add new stuff

I saw that we don't have a transaction modal for requesting verification but we can add that in a diff PR (https://github.com/protofire/omen-exchange/issues/2113)

pimato:
- [ ] 'Transaction submitted/confirmed/etc.' message in transaction modal works as expected
- [ ] Etherscan button works as expected

pimato (approved):

Issue #2010, opened 2021-06-21
use multicall for on-chain data hooks


hexyls:
With [multicall](https://github.com/makerdao/multicall.js) added in https://github.com/protofire/omen-exchange/pull/1934 we should use it across hooks that make multiple simultaneous calls for on chain data e.g. [`useBlockchainMarketMakerData`](https://github.com/protofire/omen-exchange/blob/master/app/src/hooks/useBlockchainMarketMakerData.tsx#L86). This is a strong optimization that improves loading speed.

PR #2126, opened 2021-07-26
Use multicall for market data

hexyls:
closes https://github.com/protofire/omen-exchange/issues/2010

market loading speed test with basic internet connection:
before: ~5 seconds
after: ~1 second

edit: was able to add one more optimisation based on [conditional token utils](https://github.com/gnosis/conditional-tokens-contracts/blob/master/utils/id-helpers.js#L24) but couldn't get off-chain calculation of collectionIds to work 🤷‍♂️ still got an extra ~1 second improvement doing off-chain positionId calc so that's decent anyway 

testing: market data looks normal


pimato (approved):
damned its fast!

KadenZipfel (approved):
LGTM! Awesome improvement 🙌

Mi-Lan (approved):
Gj! This is big improvment!

Issue #2026, opened 2021-06-23
Reduce bundle size to improve user experience


hexyls:
The current build of Omen is ~2.61 MB, the more we reduce this the faster the app will load for users.

@Mi-Lan has already started reducing the bundle size with PRs like https://github.com/protofire/omen-exchange/pull/2022 which is awesome.

We should probably evaluate the packages we are using and see if we can remove any, e.g. I think the `authereum` module is pretty hefty. 

We should also explore improving our [lighthouse score](https://developers.google.com/web/tools/lighthouse) which is currently ~22.

PR #2137, opened 2021-07-29
Reduce bundle size a bit

hexyls:
related https://github.com/protofire/omen-exchange/issues/2026

Was only able to reduce the bundle size a tiny bit (2.6 MB -> 2.25 MB), probably wasn't worth the effort of digging through the deps tbh but we might as well keep what we got.

It would have been nice to remove [3box-comments-react](https://github.com/3box/3box-comments-react) which is quite large, maybe we can do that later when we update the comment section 🤔

Will keep https://github.com/protofire/omen-exchange/issues/2026 open because it is more of an ongoing thing we should always be thinking about, but we can remove it from the current milestone.

Testing: App works as expected


pimato (approved):
great work!

Mi-Lan (approved):
Looks good! Yeah I would also like to remove 3box! Cant wait for that....are we planning to replace it with ceramic or something like that?

hexyls:
@Mi-Lan yeah ceramic would be super cool!

KadenZipfel (approved):
Looks good!

PR #2141, opened 2021-07-30
Add transaction modal for market verification

hexyls:
Testing:
- [ ] Transaction modal works correctly when requesting market verification


pimato:
Testing:
- [ ] Transaction modal works correctly when requesting market verification

pimato (approved):

PR #2154, opened 2021-08-06
Prevent fetching of userEarnings when the pool has no shares

hexyls:
During multicall the `feesWithdrawableBy` call throws an error if the pool has no shares. I added a check to only fetch fees if the pool has shares.

Testing:
Create a new market then withdraw all the liquidity, there should not be any multicall errors

### DIFF
--- a/app/src/hooks/market_data/useBlockchainMarketMakerData.tsx
+++ b/app/src/hooks/market_data/useBlockchainMarketMakerData.tsx
@@ -165,12 +165,6 @@ export const useBlockchainMarketMakerData = (graphMarketMakerData: Maybe<GraphMa
           call: [getCallSig(marketMaker, 'balanceOf'), cpk.address],
           returns: [['userPoolShares']],
         },
-        // get user earnings
-        {
-          target: marketMakerAddress,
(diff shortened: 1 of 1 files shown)


Frankaus:
- [x] Categorical markets
- [x]  Create market
- [x]  Buy
- [x]  Sell
- [x]  Deposit
- [x]  Withdraw
- [x]  Bond outcome
- [x]  Redeem winnings

  - [x]  Scalar markets
- [x]  Create market
- [x]  Buy
- [x]  Sell
- [x]  Deposit

- [x]  Withdraw
- [x]  Bond outcome
- [x]  Redeem winnings
- [x]  Position table works as expected
    Correct payout/profit/loss
- [x]  Market buy slider works as expected
    Correct payout/profit/loss

Account
- [x]  Correct wallet balances
- [x]  Correct omen account balances
- [ ]  xDai relay
- [ ]  Deposit
- [ ]  Withdraw
- [x]  Mobile
Should just be relatively functional as per usual

pimato (approved):
Great fix!

Mi-Lan (approved):
Works good! Both on scalar and categorical!

PR #2155, opened 2021-08-06
Don't charge relay fee when withdrawing xDai

hexyls:
We fixed this withdraw issue recently but it got lost in the refactor.


Mi-Lan (approved):
Good catch!

Issue #2148, opened 2021-08-03
Liquidity Provider fee doesn't show custom fee in tooltip.


kobellomango:
**Describe the bug**
When a market has a custom LP fee, such as 3%, the tooltip does not show this figure, but instead shows the default LP fee as 2%. 

**To Reproduce**
1. Go to https://omen.eth.link/#/0x617df4629b2b11ab775562b69ff20e148fe77fb8/buy
2. Click on 'Outcome'
3. Scroll down to 'Fund Market'
4. Enter amount
5. Hover tooltip beside 'Fee' to the right. 
6. Witness the tooltip incorrectly stating a 2% LP fee, when it's 3%.

**Expected behavior**
I expected to see "A 3% fee goes to liquidity providers"

**Screenshots**
![image](https://user-images.githubusercontent.com/85288121/128016454-ba19b320-e61f-4fca-b12b-ea909b6a5cf3.png)

Note: Screenshot taken from market with 4% fee.

**Desktop (please complete the following information):**
 - OS: macOS Big Sur
 - Browser Firefox
 - Version 90.0.2

PR #2158, opened 2021-08-06
Fix/2148 - Tooltip LP fee shows dynamic value

Frankaus:
closes https://github.com/protofire/omen-exchange/issues/2148

Tooltip LP fee shows dynamic value instead of static 3%


pimato (approved):

Mi-Lan (approved):
GJ! Nice fix dunno why we had it hardcoded like that when we had values for all tooltips done correctly!

Issue #2071, opened 2021-07-02
Move Enable Market Component below order form (1 day)


pimato:
**Describe the task that needs to be done**
The Enable Market Component is showing up below the order form in the market creation form but on the buy/sell order form it is actually above the order form:
<img width="613" alt="Bildschirmfoto 2021-07-02 um 05 45 50" src="https://user-images.githubusercontent.com/5337809/124217633-f14c8300-daf8-11eb-8dd5-959966c95f44.png">

**Describe how you have considered doing it**
We should be consistent in the UI level. Move the Enable Market component below the order form everywhere.

PR #2164, opened 2021-08-09
Fix: Move SetAllowance component below

Frankaus:
closes https://github.com/protofire/omen-exchange/issues/2071

Fix: Moved the component SetAllowance (previously Enable Markets) below to follow design spec and added margin-top on the following button with a condition: margin if the set Allowance component is present; no margin if not present


app/src/components/market/sections/market_bond/market_bond.tsx
```diff
@@ -44,6 +44,7 @@ interface Props extends RouteComponentProps<any> {
 const BottomButtonWrapper = styled(ButtonContainer)`
   margin: 0 -24px;
   padding: 20px 24px 0;
+  ${props => (props.marginTop ? '20px' : '0')}
```
Mi-Lan, line 47: @Frankaus So here you don't need this props since inside the ButtonContainer there is already marginTop which should be reused...

Mi-Lan (changes requested):
Small thing and should be gtg!

app/src/components/market/sections/market_bond/market_bond.tsx
Frankaus: right! Thanks mate

Issue #2073, opened 2021-07-02
Add 2.5% potential fee info when user wants to bond on an outcome (1/1.5 day)


pimato:
**Describe the task that needs to be done**
Add a new line on the bonding info component which gives the user the info that if his bond is not the winning bond, a 2.5% potential fee is applied.

**Describe how you have considered doing it**
<img width="1556" alt="Bildschirmfoto 2021-07-02 um 07 09 06" src="https://user-images.githubusercontent.com/5337809/124223657-68d3df80-db04-11eb-8a2d-f83e77cf0531.png">

- Add a row into the info box `Potential Fee` with the 2.5% fee in the currency used. If you hover over the question mark icon a tooltip should appear with the info: ```A 2.5% fee is applied and burned on bonds which are not the final bond.```

PR #2169, opened 2021-08-10
Feature: Add 2.5% potential fee info when user wants to bond on an ou…

Frankaus:
closes https://github.com/protofire/omen-exchange/issues/2073

Feature: Add 2.5% potential fee info when user wants to bond on an outcome


hexyls (approved):
Looks good!

app/src/components/market/market_bond/market_bond.tsx
```diff
@@ -272,6 +272,12 @@ const MarketBondWrapper: React.FC<Props> = (props: Props) => {
               value={`${bigNumberToString(bondNativeAssetAmount, nativeAsset.decimals)} ${symbol}`}
             />
             <TransactionDetailsLine />
+            <TransactionDetailsRow
+              state={ValueStates.important}
```
Mi-Lan, line 276: Screenshot 2021-08-12 at 14 05 40
I think because you put important here it shows value in different color then the rest....

Mi-Lan (approved):
Looks good now! Also 2 more things to consider potentially
1) Maybe to increase the decimal amount to 3 since for the first instance on xDai its not showing most accurate amount..Heres market to test http://localhost:3000/#/0xac4754e37eaef8bd7d72342b1c084a592dbe1ab9/set_outcome
Screenshot 2021-08-12 at 16 06 25
2) This can be made into another issue I just noticed it. But on the table on top right it says bonded and Dai next to it(the markets collateral). When in fact you can only bond with native asset in this case eth.
http://localhost:3000/#/0x6f62dbfcaec45973fa5dc25ec958663c4f9a6a00/set_outcom
Screenshot 2021-08-12 at 16 08 35
@pimato Thoughts?

pimato:
> 1. Maybe to increase the decimal amount to 3 since for the first instance on xDai its not showing most accurate amount..Heres market to test http://localhost:3000/#/0xac4754e37eaef8bd7d72342b1c084a592dbe1ab9/set_outcome
>    Screenshot 2021-08-12 at 16 06 25
can we just not round the potential fee here and just show `0.02` instead of rounded up to `0.03`?

> 2. This can be made into another issue I just noticed it. But on the table on top right it says bonded and Dai next to it(the markets collateral). When in fact you can only bond with native asset in this case eth.
>    http://localhost:3000/#/0x6f62dbfcaec45973fa5dc25ec958663c4f9a6a00/set_outcom
>    Screenshot 2021-08-12 at 16 08 35

great catch, created a separate issue: https://github.com/protofire/omen-exchange/issues

pimato (approved):

PR #2177, opened 2021-08-11
Added strict for max button formatting

Mi-Lan:
- This fixed error that was caused by numbers bigger then 1000 having comma and not being displayed properly on max button click on lock token modal 
- How to test 
https://user-images.githubusercontent.com/33226956/129030595-237fb068-a562-492d-a298-2364888d1cab.mov
- [ ] Make sure above is not happening and it displays appropriate value


pimato (changes requested):
@Mi-Lan the max button seems to work now but the `Lock OMN Button` is disabled. It looks like some rounding issue:
https://user-images.githubusercontent.com/5337809/129053503-c67dbbb0-02ae-485a-8c2b-5639b2dbb505.mov

pimato (approved):
great fix! Looks good to go!

KadenZipfel (approved):
Lgtm!

PR #2183, opened 2021-08-11
Fix: Move SetAllowance component below (updated after guild merge) - V2

Frankaus:
replaces https://github.com/protofire/omen-exchange/pull/2164
closes https://github.com/protofire/omen-exchange/issues/2071

Fix: Moved the component SetAllowance (previously Enable Markets) below to follow design spec and added margin-top on the following button with a condition: margin if the set Allowance component is present; no margin if not present


Mi-Lan (approved):
LGTM!
Screenshot 2021-08-12 at 13 21 19

pimato (changes requested):
Please remove the margin here:
Bildschirmfoto 2021-08-12 um 15 33 13

Also make sure that we do not break anything else!

pimato (approved):

Issue #2118, opened 2021-07-21
Use new text styles and color variables for Check Address Modal (2 hours)


pimato:
Now that we have clearly defined text & color style guide we should use the the new styles with the Check-Address Modal:
![Check-Address](https://user-images.githubusercontent.com/5337809/126504182-664f0bc6-ec94-4085-8659-5513915b46c6.png)

[Figma Design Specification](https://www.figma.com/file/R056tSYMjvfwaFidEBItYy/Widgets-V2?node-id=3832%3A241)

Issue #2117, opened 2021-07-21
Use new text styles and color variables for Omen Guild Membership Modal (2 hours)


pimato:
Now that we have clearly defined text & color style guide we should use the the new styles with the Omen Guild Membership Modal:
![guild-membership](https://user-images.githubusercontent.com/5337809/126503732-01a9b414-ba35-4b3b-8470-99a4f8e33e4e.png)

[Figma Design Specification](https://www.figma.com/file/R056tSYMjvfwaFidEBItYy/Widgets-V2?node-id=3832%3A241)

Issue #2116, opened 2021-07-21
Use new text styles and color variables for Omen Guild Airdrop Modal (1/2 Day)


pimato:
Now that we have clearly defined text & color style guide we should adopt the Airdrop Modal with the new styles:
![claim](https://user-images.githubusercontent.com/5337809/126503013-89986f6c-94df-4809-91d4-1ad5d882c9a0.png)

[Figma Design Specification](https://www.figma.com/file/R056tSYMjvfwaFidEBItYy/Widgets-V2?node-id=3832%3A241)

PR #2185, opened 2021-08-11
Ref: Update colours and text with new theme in Airdrop Modal

### LINKED ISSUE #2119: Use new text styles and color variables for Lock Omen Token Modal (2 hours)
Now that we have clearly defined text & color style guide we should use the the new styles with the Lock Omen Token Modal:
lock-omen

[Figma Design Specification](https://www.figma.com/file/R056tSYMjvfwaFidEBItYy/Widgets-V2?node-id=3832%3A241)

Frankaus:
closes https://github.com/protofire/omen-exchange/issues/2116
closes https://github.com/protofire/omen-exchange/issues/2118
closes https://github.com/protofire/omen-exchange/issues/2117
closes #2119 

We have redefined the Text styled component use in the new theme and applied it throughout the entire Airdrop modal together with the Guild Membership section

Note for review:
Please let me know if these 2 colours on the numbers show correctly cause I couldn't test that.
Screenshot 2021-08-13 at 15 48 09


app/src/theme/index.tsx
```diff
@@ -387,17 +387,26 @@ function themeAggregator(darkMode: boolean) {
...
-  letter-spacing: 0.2px;
+  font-size: ${props => props.fontSize};
+  font-weight: ${props => props.fontWeight};
+  letter-spacing: ${props => (props.letterSpacing ? props.letterSpacing : '0.2px')};
+  line-height: ${props => props.lineHeight};
+  ${props => props.margin && `margin: ${props.margin}`};
+  ${props => props.marginBottom && `margin-bottom: ${props.marginBottom}`};
+  ${props => props.marginLeft && `margin-left: ${props.marginLeft}`};
+  ${props => props.marginTop && `margin-top: ${props.marginTop}`};
+  ${props => props.display && `display: ${props.display}`};
```
Mi-Lan, line 409: Hey men so this is not a way to go about it....The last parameter for each Type {...props} is supposed to do that what you wrote here. It basically passes all the properties down to the text type. If it doesn't work like that then we need to adjust it to work...

hexyls:
@Frankaus just added a small example commit for how we might do this:

It looks like we mostly only need a custom prop for the color because we need to look up the color in our theme. But for the rest of the styling we can just go with the standard `style` prop:

```
bodyMedium(props: any) {
  const { style, ...rest } = props
  return <TextWrapper {...rest} style={{ fontSize: 14, fontWeight: 500, lineHeight: '18px', ...style }} />
},
```

we also expose the `style` prop and apply it last, just in case we want to override it elsewhere.

Frankaus:
> @Frankaus just added a small example commit for how we might do this:
> 
> It looks like we mostly only need a custom prop for the color because we need to look up the color in our theme. But for the rest of the styling we can just go with the standard `style` prop:
> 
> ```
> bodyMedium(props: any) {
>   const { style, ...rest } = props
>   return <TextWrapper {...rest} style={{ fontSize: 14, fontWeight: 500, lineHeight: '18px', ...style }} />
> },
> ```
> 
> we also expose the `style` prop and apply it last, just in case we want to override it elsewhere.

Thank you so much for your help! Will test it now and move on:)

Mi-Lan:
> @Frankaus just added a small example commit for how we might do this:
> 
> It looks like we mostly only need a custom prop for the color because we need to look up the color in our theme. But for the rest of the styling we can just go with the standard `style` prop:
> 
> ```
> bodyMedium(props: any) {
>   const { style, ...rest } = props
>   return <TextWrapper {...rest} style={{ fontSize: 14, fontWeight: 500, lineHeight: '18px', ...style }} />
> },
> ```
> 
> we also expose the `style` prop and apply it last, just in case we want to override it elsewhere.

Yeah I just looked into it! Thats how I would also go about it but I would maybe add addition this this so that we can have cleaner code. 
`bodyMedium(props: any) {
    const { style, ...rest } = props
    return <TextWrapper {...rest} style={{ fontSize: 14, fontWeight: 500, lineHeight: '18px', ...style, ...rest }} />
  },`
Because with this below we can just write and also all non style would just be invalid and won't hurt the style 
`<TYPE.bodyMedium
            color={'blue'}
            onClick={() => {
              console.log('here')
            }}
            padding="22px 0 0 0"
            text-align="right"
            width={'100'}
          >
            Hersssse
          </TYPE.bodyMedium>`

app/src/theme/index.tsx
```diff
@@ -387,33 +387,71 @@ function themeAggregator(darkMode: boolean) {
...
   return <StyledThemeProvider theme={themeAggregator(darkMode)}>{children}</StyledThemeProvider>
 }
 
-const TextWrapper = styled.div<{ color: keyof Colors }>`
-  color: ${props => props.theme.text3};
+const TextWrapper = styled.div<{ color: keyof Colors; letterSpacing: string }>`
+  color: ${({ color, theme }) => (theme as any)[color]};
   font-family: Roboto;
-  letter-spacing: 0.2px;
+  letter-spacing: ${props => (props.letterSpacing ? props.letterSpacing : '0.2px')};
```
Mi-Lan, line 401: I think there is no need for letterSpacing attribute here....With new additions there seems no reason...

Frankaus: Rightly yes

Mi-Lan (commented):
Lock modal colors don't seem to match the designs...https://www.figma.com/file/R056tSYMjvfwaFidEBItYy/Widgets-V2?node-id=3832%3A241
Screenshot 2021-08-16 at 20 02 49

Mi-Lan:
which components to check out:

- [ ] airdrop card
- [ ] airdrop card check address 
- [ ] modal for token locking

pimato:
wrong green on the airdrop alert modal:
Bildschirmfoto 2021-08-17 um 13 58 58

should be `green` from our color palette

pimato (changes requested):
Bildschirmfoto 2021-08-17 um 14 01 23

red color is wrong, should be `error` color

Frankaus:
@pimato We need to change the way we call green red in our palette. Because with the new way we infer color into the textWrapper it seems that the machine takes a generic green red etc. Can you suggest a name for them pls? A simple green1 or greenOmen might solve it:)

pimato:
> @pimato We need to change the way we call `green` `red` in our palette. Because with the new way we infer color into the textWrapper it seems that the machine takes a generic green red etc. Can you suggest a name for them pls? A simple `green1` or `greenOmen` might solve it:)

for which specific colors do we need this exactly?

Frankaus:
> > @pimato We need to change the way we call `green` `red` in our palette. Because with the new way we infer color into the textWrapper it seems that the machine takes a generic green red etc. Can you suggest a name for them pls? A simple `green1` or `greenOmen` might solve it:)
> 
> for which specific colors do we need this exactly?

for `white` `green` and `red`

pimato:
> > > @pimato We need to change the way we call `green` `red` in our palette. Because with the new way we infer color into the textWrapper it seems that the machine takes a generic green red etc. Can you suggest a name for them pls? A simple `green1` or `greenOmen` might solve it:)
> > 
> > 
> > for which specific colors do we need this exactly?
> 
> for `white` `green` and `red`

just changed it from 

`white` to `snow`
`red` to `alert`
and `green` to `profit`

hope this helps

Frankaus:
Colours work well that way, nice catch. There was no need to edit the `white` because we use a standard #fff white hex code. In the event we wish to change that white I will edit inside the theme and make the changes else where as well. We are using the white value inside the `button`, `global_style` and `react_tooltip_styles_override` files so I that event I suggest to make the change in a separate PR. It's easy and quick to do

pimato (changes requested):
Looks like the date here is has the color #000 but should be `text1`:
Bildschirmfoto 2021-08-19 um 16 13 59

Bildschirmfoto 2021-08-19 um 16 14 30

pimato (approved):

pimato:
Date and Icon should be in one line:
Bildschirmfoto 2021-08-24 um 14 49 58

pimato (approved):

app/src/components/modal/modal_lock_yo_tokens/index.tsx
```diff
@@ -61,15 +52,6 @@ const ModalMain = styled.div`
...
 `
-const LightDataItem = styled.div`
-  color: ${props => props.theme.textfield.placeholderColor};
-`
-const DarkDataItem = styled.div`
-  display: flex;
-  align-items: center;
-  color: ${props => props.theme.colors.textColorDark};
-  font-weight: ${props => props.theme.textfield.fontWeight};
-`
```
Mi-Lan, line 72: @Frankaus So we should only use pure inline TYPE.something when there is one instance of it. Where we have repeating styles like this we should still use styled classes but instead of div you can put type like this.
const LightDataItem = styled(TYPE.bodyMedium) and try to incorporate color into styled. 
Because you are now repeating styles inline which is not good at all...

Mi-Lan (changes requested):
Requested changes...

app/src/components/modal/modal_lock_yo_tokens/index.tsx
```diff
@@ -54,22 +46,23 @@ const DataRow = styled.div`
   width: 100%;
   line-height: ${props => props.theme.fonts.defaultLineHeight};
 `
+
+const LightDataItem = styled(TYPE.bodyRegular)<{ color: keyof Colors }>`
+  color: ${props => props.theme.text2};
+`
+const DarkDataItem = styled(TYPE.bodyMedium)<{ color: keyof Colors }>`
```
Mi-Lan, line 53: Wondering whats the use of this <{ color: keyof Colors }> since you are not passing it inline but directly on the styled. It can be deleted....

Mi-Lan (approved):
LGTM!

Issue #2178, opened 2021-08-11
Blockscout link not working for claiming OMN tokens via xDai tight integration


pimato:
**Describe the bug**
So it seems like the link to blockscout on the transaction modal for claiming OMN tokens via xDai tight integration is broken. It gives you a blockscout link with page not found.

**Expected behavior**
Link should open up a blockscout page from the correct relayed transaction.

**Screenshots**
https://user-images.githubusercontent.com/5337809/129031348-fe125079-6aba-47cc-90ff-ea502c421caf.mov

KadenZipfel:
I've actually reproduced this on a different app and had the same issue. Even the link etherscan was giving me was not working, so I think the issue is simply that blockscout is failing to index transactions

pimato:
> I've actually reproduced this on a different app and had the same issue. Even the link etherscan was giving me was not working, so I think the issue is simply that blockscout is failing to index transactions

Will revisit in the coming days in close if fixed by blockscout.

hexyls:
Yeah just tested out a normal xdai buy and had the same issue, hopefully resolves itself 👍

PR #2187, opened 2021-08-12
Update blockscout url

hexyls:
closes https://github.com/protofire/omen-exchange/issues/2178

testing: Make any tx on xdai, clicking blockscout link should show you tx information


pimato (approved):
I think both urls work but I think it is good that we replace the legacy one with the new one.

Mi-Lan (approved):
Looks good!

PR #2188, opened 2021-08-12
Only fetch airdrop info when connected to relay

hexyls:
Now that we are only doing the airdrop through the relay there is no point fetching airdrop info on other networks.

PR also hides the airdrop card in the guild modal if not connected to the relay.


pimato (approved):
works great!

Mi-Lan (approved):
Works!

PR #2078, opened 2021-07-05
Feature/1997bis - Text Style Handling

Frankaus:
closes https://github.com/protofire/omen-exchange/issues/1997

Refactor: Create new TextWrapper and refactor 5 text types

Figma design specification:
https://www.figma.com/file/zVniQrYgkj8AfQs5zNeBrS/Components-FINAL?node-id=39%3A22

PR #1851, opened 2021-04-21
Add liquidity mining data to market details view

KadenZipfel:
Closes: https://github.com/protofire/omen-exchange/issues/1833

PR #2205, opened 2021-08-13
Liq rewards proposals

### LINKED ISSUE #2044: Create Design Specification for manual rewards request for the Guild

hexyls:
closes https://github.com/protofire/omen-exchange/issues/2044


hexyls:
@pimato - we can just do an initial styling review for this one, then I can fix up any issues in prep for https://github.com/protofire/omen-exchange/pull/1851 being merged

pimato (changes requested):
Currently it is possible to pick two markets and request liquidity rewards which we don´t want:

https://user-images.githubusercontent.com/5337809/130075581-331c6413-3557-405e-b4f3-c7bc579c5d22.mov

The user should only be able to pick one market.

pimato:
In order to have not stretched components like this we need to give the text box a fixed height of 54px and align the text vertically centered.

currently:
Bildschirmfoto 2021-08-19 um 15 29 35

should:
Bildschirmfoto 2021-08-19 um 15 31 23

pimato:
Now that we have our clearly defined text styles, we should start using them: 
https://github.com/protofire/omen-exchange/pull/2078/files

app/src/components/common/icons/IconBack.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react'
+
+export const IconBack = () => (
+  <svg fill="none" height="12" viewBox="0 0 18 12" width="18" xmlns="http://www.w3.org/2000/svg">
+    <path d="M18 5H3.83L7.41 1.41L6 0L0 6L6 12L7.41 10.59L3.83 7H18V5Z" />
+  </svg>
+)
```
Mi-Lan, line 7: I don't think its necessary to import another Icon here I used on my proposal details pr IconArrowBack with slight modification to support this color https://github.com/protofire/omen-exchange/pull/2208/files#

app/src/components/guild/index.tsx
```diff
@@ -0,0 +1,276 @@
+import { BigNumber } from 'ethers/utils'
+import React, { useEffect, useState } from 'react'
+import styled from 'styled-components'
+
```
Mi-Lan, line 4: Great that you created guild in the components folder I did the same thing on mine! Also In order to keep separation of concerns I think it might make sense to create guild folder inside pages folder and store all the logic there and separate it from view which will be stored in components? Curios what you and Kaden think about this and how we should go about it...You can also checkout maybe how I did it and if it makes sense or if we should do It differently.... @KadenZipfel @hexyls

app/src/components/guild/index.tsx
```diff
@@ -0,0 +1,276 @@
...
+  @media (max-width: ${props => props.theme.themeBreakPoints.sm}) {
+    flex-wrap: wrap;
+  }
+`
+
+const ProposalTitle = styled.div`
+  font-size: 22px;
+  font-weight: 500;
+  line-height: 26px;
+  color: ${props => props.theme.text3};
```
Mi-Lan, line 53: Also for all the text files we should be using new TYPE.something from theme/index.js file It is now on master with recent pr from francesco that we pushed...

Mi-Lan (changes requested):
@hexyls Hey men! Requested some changes so we keep the codebase aligned with stuff from refactoring. Curios what you think!

app/src/components/common/icons/IconBack.tsx
hexyls: Cheers dude, copied!

app/src/components/guild/index.tsx
hexyls: done 🏄‍♂️ was cool putting `TYPE` into action, pretty powerful and easy to use

app/src/components/guild/index.tsx
hexyls: I really like this idea, made a `proposed_rewards_page.tsx` and a `proposed_rewards_view.tsx`, much better now

PR #2211, opened 2021-08-16
Ref: Refactored Deposit/Withdrawal modal

### LINKED ISSUE #2196: Use new text styles and color variables for Transaction Modal (2 hours)
Now that we have clearly defined text & color style guide we should adopt the Transaction Modal with the new styles

[Figma Spec](https://www.figma.com/file/R056tSYMjvfwaFidEBItYy/Widgets-V2?node-id=3832%3A241)

Frankaus:
closes #2196

### DIFF
--- a/app/src/components/modal/modal_deposit_withdraw/index.tsx
+++ b/app/src/components/modal/modal_deposit_withdraw/index.tsx
@@ -9,6 +9,7 @@ import styled, { withTheme } from 'styled-components'
 import { DAI_TO_XDAI_TOKEN_BRIDGE_ADDRESS, OMNI_BRIDGE_MAINNET_ADDRESS } from '../../../common/constants'
 import { useConnectedWeb3Context } from '../../../contexts'
 import { ERC20Service, XdaiService } from '../../../services'
+import { TYPE } from '../../../theme'
 import { bridgeTokensList, getNativeAsset, getToken, networkIds } from '../../../util/networks'
 import { getImageUrl } from '../../../util/token'
 import { bigNumberToString, waitForConfirmations } from '../../../util/tools'
@@ -24,16 +25,13 @@ import { IconInfo } from '../../common/icons/IconInfo'
 import { Image } from '../../market/common_sections/message_text/token_item'
 import {
   BalanceItem,
-  BalanceItemBalance,
   BalanceItemSide,
-  BalanceItemTitle,
   BalanceItems,
   BalanceSection,
   ContentWrapper,
   ModalCard,
   ModalNavigation,
   ModalNavigationLeft,
-  ModalTitle,
 } from '../common_styled'
 import { ModalTransactionWrapper } from '../modal_transaction'
 
@@ -50,12 +48,6 @@ const InputInfo = styled.div`
   border-radius: 4px;
   padding: ${props => props.theme.textfield.paddingVertical + ' ' + props.theme.textfield.paddingHorizontal};
   line-height: 20px;
-  letter-spacing: 0.2px;
-`
-const WalletText = styled.div`
-  margin-bottom: 14px;
-  color: ${props => props.theme.text2};
-  line-height: 16.41px;
   letter-spacing: 0.2px;
 `
 
@@ -265,17 +257,17 @@ export const ModalDepositWithdraw = (props: Props) => {
         <BalanceItemSide>
           <RadioInput checked={currencySelected === item} name={item} outcomeIndex={-1} readOnly />
           <Image size={'24'} src={getImageUrl(address)} style={{ marginLeft: '12px', marginRight: '12px' }} />
-          <BalanceItemTitle notSelected={currencySelected !== item}>
+          <TYPE.bodyRegular color={currencySelected !== item ? 'text2' : 'text1'} margin={'0px'}>
             {name ? name : symbol.toLowerCase()}
-          </BalanceItemTitle>
+          </TYPE.bodyRegular>
         </BalanceItemSide>
         <BalanceItemSide>
-          <BalanceItemBalance>
+          <TYPE.bodyRegular color={'text2'} margin={'0px'}>
             {token?.balance
               ? bigNumberToString(new BigNumber(token?.balance), decimals, symbol === 'DAI' ? 2 : 3)
               : '0.00'}{' '}
             {symbol}
-          </BalanceItemBalance>
+          </TYPE.bodyRegular>
         </BalanceItemSide>
       </BalanceItem>
     )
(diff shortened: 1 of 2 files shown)


pimato (approved):

Mi-Lan (changes requested):
Seems that hover effect for wallet currencies is lost with your changes....

https://user-images.githubusercontent.com/33226956/130101585-d3e88c1f-665a-4f2e-b0a0-bcf1da4b1bf4.mov

app/src/components/modal/common_styled/index.tsx
```diff
@@ -91,6 +91,9 @@ export const BalanceItem = styled.div<{ hover?: boolean }>`
   }
 
   &:hover {
+    .hover {
```
Frankaus, line 94: here we can implement a similar but different solution. We can import `TextWrapper` from the theme and target it `${TextWrapper` inside the :hover call. Though also the token balances will get the hover effect, which in this case looks also good. But the first solution, implemented here, is more flexible to use now andin the future cause we simply have to add the className `hover` to activate the hover when we wish within a `BalanceItem`

Mi-Lan: So the reason that It doesn't work currently is because its selecting balanceItemSide and BalanceItemTitle for hover and you have replaced them with TYPE.sometihing....I would suggest to leave those classes and handle them like airdrop issue if possible. Or try to target those classes also you have to be careful when changing and deleting styled components like this because In some cases we select them based on styled class like here.
Screenshot 2021-08-25 at 18 23 37
Also why are you using theme.colors.textColorDark when that is theme.text1 like the item that is below in this picture....

app/src/App.tsx
```diff
@@ -59,9 +59,7 @@ const App: React.FC = (props: any) => {
           >
             <ContentWrapper>
               <ConnectionModalNavigation>
-                <ModalTitle>
-                  <span style={{ fontFamily: 'Roboto' }}>Settings</span>
-                </ModalTitle>
+                <ModalTitle marginTop={'2px'}>Settings</ModalTitle>
```
Mi-Lan, line 62: What is the use of this top margin? Seems unrelated to the issue.... was it wrong?

app/src/components/modal/modal_deposit_withdraw/index.tsx
```diff
@@ -304,7 +297,9 @@ export const ModalDepositWithdraw = (props: Props) => {
                   setAmountToDisplay('')
                 }}
               />
-              <ModalTitle style={{ marginLeft: '16px' }}>{exchangeType} Asset</ModalTitle>
+              <TYPE.heading3 color={'text1'} margin={'2px 0px 0px 16px'}>
```
Mi-Lan, line 300: Also this 2px top margin is making items not be 100% vertically aligned I think its fine having just 16px left margin @pimato wdyt?
Screenshot 2021-08-30 at 13 25 24

app/src/components/modal/modal_deposit_withdraw/index.tsx
```diff
@@ -344,17 +339,19 @@ export const ModalDepositWithdraw = (props: Props) => {
           {exchangeType === ExchangeType.withdraw && currencySelected !== 'dai' && xDaiBalance?.isZero() ? (
             <InputInfo>
               <IconAlertInverted />
-              <div style={{ marginLeft: '12px' }}>Fund your Omen Account with Dai to proceed with the withdrawal.</div>
+              <BalanceItemBalance marginLeft={'12px'}>
+                Fund your Omen Account with Dai to proceed with the withdrawal.
+              </BalanceItemBalance>
```
Mi-Lan, line 344: Not sure but this text doesn't seem to be the same color as in figma...And also we need to pay attention when reusing classes that it makes sense... For example here it would maybe make sense to name it something more general if we are going to use it for multiple stuff...

app/src/components/modal/common_styled/index.tsx
```diff
@@ -68,16 +64,8 @@ export const BalanceItemSide = styled.div`
...
-  font-size: ${props => props.theme.fonts.defaultSize};
+export const BalanceItemTitle = styled(TYPE.bodyRegular)<{ notSelected: boolean }>`
   color: ${props => (props.notSelected ? props.theme.text2 : props.theme.text1)};
-  margin: 0;
-`
-
-export const BalanceItemBalance = styled.p`
-  font-size: ${props => props.theme.fonts.defaultSize};
-  color: ${props => props.theme.text2};
-  margin: 0;
```
Mi-Lan, line 80: Also since we now only have one use for balanceItemTitle and its seems redundatnt we should consolidate these 2 classes into one. And name it something more general like that would make more sense with new addtions.  Like maybe DataItemLight or something like that, im shit at naming....

Mi-Lan (changes requested):
@Frankaus Left couple of things to be changed...

app/src/App.tsx
Frankaus: @Mi-Lan the marginTop of 2px was there before as in the case of the comment below. 

It aligns it vertically, if you delete it, you see it's not 100% vertically aligned. It seems to me it's aligned with the marginTop, that's why I kept it

Mi-Lan: @Frankaus Thats not proper way to align items vertically....Look at other examples in the codebase how we do it...

app/src/components/modal/modal_deposit_withdraw/index.tsx
pimato: Yeah that makes sense as long as `Deposit Asset` is vertical centered to the arrow and close icon!

Frankaus: Will use other technique than the marginTop for that

app/src/App.tsx
Frankaus: Agreed

app/src/components/modal/modal_deposit_withdraw/index.tsx
Frankaus: In this case that naming didn't make sense yes.. will revert that one to a its TYPE here it's ok.

Frankaus: the color on figma was wrong. It hsould have been text2 like in the code. All good

app/src/components/modal/common_styled/index.tsx
Frankaus: I looked into it and it's quite impossible because we would loose the hover effect, or at least, the hover effect would also apply to the balances of the item. Let's discuss that on a quick call with the files in front of us...

Mi-Lan (approved):
I pushed some small changes and think its ready to be merged...

pimato (approved):

PR #2215, opened 2021-08-17
Fix: bigNumberToString fn also return <0.001 

### LINKED ISSUE #2214: bigNumberToString fn should also format `< x.xx1`
Currently the new formatting function bigNumberToString formats a small number with 2 decimals with `<0.01` when its argument `strict` is set to false.

It should do the same job with 3 decimals: `<0.001`

Frankaus:
closes #2214 

**Testing**
After merging #2169 you can test this feature in the Set Outcome stage of the market bond component in the Potential Fee entry row having `<0.001` while other entries in market buy still show `<0.01`


app/src/util/tools/formatting.ts
```diff
@@ -33,6 +33,9 @@ export const formatNumber = (number: string, decimals = 2): string => {
   if (number.length < 1) {
     return `0${decimals > 0 ? '.' + '0'.repeat(decimals) : ''}`
   }
+  if (Number(number) < 0.001 && Number(number) > 0 && decimals > 2) {
+    return '<0.001'
+  }
```
Mi-Lan, line 38: @Frankaus This works but I don't think its the best approach. So we wanna make it so that according to number of decimal places (precision) we define it checks for that amount of decimal places.
For example :
If we have one decimal place it will evaluate to <0.1
if we have 2 decimals it will evaluate to <0.01
if we have 3 decimals it will evaluate to <0.001
if we have 4 decimal it will evaluate to <0.0001
etc....
And we should make it in one place and to do this programatically not hardcoding it...

Mi-Lan (changes requested):
Left feedback on how I think this should be done

app/src/util/tools/formatting.ts
Frankaus: Definitely better approach yes. I thought of a for loop and it works on my machine. That should do the job, hope you like it

app/src/util/tools/formatting.ts
```diff
@@ -33,8 +33,13 @@ export const formatNumber = (number: string, decimals = 2): string => {
...
   }
-  if (Number(number) < 0.01 && Number(number) > 0) {
-    return '<0.01'
+
+  if (Number(number) > 0) {
+    for (let i = 1; i < 10; i++) {
+      if (Number(number) < Number('0.' + '0'.repeat(i) + '1') && decimals === i + 1) {
+        return '<0.' + '0'.repeat(i) + '1'
+      }
+    }
```
Mi-Lan, line 42: @Frankaus Why are we making this loop? I think because we have decimals we can just determine which evaluation we need to do and use the value for evaluation to be based on number of decimals....

app/src/util/tools/formatting.ts
```diff
@@ -34,12 +34,16 @@ export const formatNumber = (number: string, decimals = 2): string => {
...
-    }
+  // if (Number(number) > 0) {
+  //   for (let i = 1; i < 10; i++) {
+  //     if (Number(number) < Number('0.' + '0'.repeat(i) + '1') && decimals === i + 1) {
+  //       return '<0.' + '0'.repeat(i) + '1'
+  //     }
+  //   }
+  // }
+
+  if (Number(number) > 0 && Number(number) < Number('0.' + '0'.repeat(decimals - 1) + '1')) {
```
Frankaus, line 45: I think the for loop solution is more dynamic and precise but I leave you decide what's best

Mi-Lan:
@hexyls Hey men would be good if you have time to take a look I added tests also and it should act as expected...But haven't dwelled that much how it looks on components themselves or if it has some issues. But should work in theory...

hexyls (approved):
Looks good!

PR #2216, opened 2021-08-18
Fix: Fixed collateral description in outcome table when on Set Outcome

### LINKED ISSUE #2192: Fix static collateral description in outcome table in `Set Outcome - View`
**Describe the bug**
It looks like in the `Set Outcome - View` we have the static description of collateral `DAI` for the `Bonded - Column. Omen is still using both Mainnet and xDai so instead of always showing `DAI` we should show the native asset of the chain as description.

**To Reproduce**
1. Go to http://localhost:3000/#/0x028c7df3ab5c97770795212b8d99c0390dd39c34/set_outcome
4. See error:
Bildschirmfoto 2021-08-12 um 16 16 50

**Expected behavior**
Should have shown ETH as Bonded collateral.

Frankaus:
closes #2192

The `Bonded` header of the Outcome Table was set on the Symbol of the collateral question while we want it to be set to the native asset of the chain

### DIFF
--- a/app/src/components/market/common_sections/tables/outcome_table/index.tsx
+++ b/app/src/components/market/common_sections/tables/outcome_table/index.tsx
@@ -124,6 +124,7 @@ export const OutcomeTable = (props: Props) => {
 
   const TableCellsAlign = ['left', 'left', 'right', 'right', 'right', 'right', 'right']
   const symbol = useSymbol(collateral)
+  const bondCollateral = relay ? 'DAI' : 'ETH'
   const renderTableHeader = () => {
     return (
       <THead>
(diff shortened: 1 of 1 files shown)


pimato (approved):

app/src/components/market/common_sections/tables/outcome_table/index.tsx
```diff
@@ -124,6 +124,7 @@ export const OutcomeTable = (props: Props) => {
 
   const TableCellsAlign = ['left', 'left', 'right', 'right', 'right', 'right', 'right']
   const symbol = useSymbol(collateral)
+  const bondCollateral = relay ? 'DAI' : 'ETH'
```
Mi-Lan, line 127: @Frankaus Because of the way you made it..This test market on xDai is showing eth
http://localhost:3001/#/0xebb1b67bafbb74a22e293b636ee66d2cbc865f9f/finalize
Screenshot 2021-08-18 at 17 31 07
So the simplest solution is to use nativeAsset variable which is already present in this file...

Mi-Lan (changes requested):
Explained below how to do this

